Issue #176, opened 2026-06-26
Detect the GP relay when the PO page loads


Xilous:
UC Nexus should find and connect to the relay as the PO module opens, not wait for the Create PO dialog. surface a relay status indicator on the PO page and pre-warm the connection so the user knows up front whether GP actions are available.

- on the PO module / page load, background-check relay presence (/health) and show a status indicator (connected vN, or not detected on this machine); optionally pre-fetch the relay credential. drive the Create PO and Register actions off this state.

this is done once opening /app/po shows the relay status without opening any dialog, and the indicator flips correctly when the relay is stopped and restarted.

it touches po/index.tsx, which the lifecycle-rename issue also touches, so sequence it after that one or merge carefully. the relayClient already exists.

Issue #175, opened 2026-06-26
Register a Draft PO into GP (PO-request acceptance) with GP-vendor mapping


Xilous:
imported PO requests are created as Draft with vendor names from the hardware schedule, which will not match GP. a PO user accepts a Draft by mapping it to the real GP vendor and pushing it to GP, turning Draft into GP-Registered. this is the import path (path 1); the end result must be identical to a manually created PO (path 2), so they share the GP push.

- the PO list gets a "Register in GP" action on a Draft PO that opens a dialog with the GP section (relay status, company, buyer, the GP-sourced cost code) plus a GP-vendor selector. the selector defaults to a best-guess name match against the synced GP vendors but the user must confirm it before registering. on submit it runs GP-first - relay /po, then on success record the GP po_number, set the mapped vendor and cost code on the PO, and advance Draft to GP-Registered. on a GP failure nothing changes and it stays Draft.
- the manual Create PO submit is refactored to share that same GP-push path.
- the backend adds a way to register a draft (set vendor + cost_code, record the GP result, advance the status), via a new mutation or by extending recordPoGpSync + updatePo.

this is done once an imported Draft PO can be registered into GP from the PO list, becomes GP-Registered with the GP number, and a GP failure leaves it Draft.

depends on the lifecycle-rename issue (GP_REGISTERED) and the cost-codes-from-GP issue (the job-sourced dropdown). it touches CreatePODialog and the PO module.

Issue #174, opened 2026-06-26
PO lifecycle: rename ORDERED to GP_REGISTERED and remove gp_sync_status


Xilous:
a PO only exists once it is in GP, so the "Ordered" state really means "registered in GP" - rename it. and because a PO that fails to create in GP is never created in UC Nexus either, there is no persisted FAILED, so the separate gp_sync_status field is redundant with the status and should go.

- the backend renames the po_status enum value ORDERED to GP_REGISTERED via an alembic migration (ALTER TYPE ... RENAME VALUE) and updates every reference - the vendor-confirmed auto-transition, the receiving transitions, mark_po_as_ordered, and the POStatistics ordered count. it also drops the gp_sync_status column + enum and the part of recordPoGpSync that set it, keeping recordPoGpSync for the po_number plus the DRAFT to GP_REGISTERED advance.
- the frontend uses GP-Registered in the status labels, chips, filters and sort, and removes the gpSyncStatus chip and query field added earlier.
- the resulting lifecycle is Draft to GP-Registered to Vendor-Confirmed to Partially-Received / Closed; manual POs start at GP-Registered, imported PO requests start at Draft.

this is done once a PO created in GP shows status GP-Registered, gpSyncStatus is gone everywhere, the existing transitions still work, and the migration up / down plus alembic check pass.

foundational - the draft-registration issue depends on it. it touches po/index.tsx and CreatePODialog and the backend status code, so sequence it before the draft-registration and presence issues that also touch po/index.tsx, or merge carefully.

Issue #172, opened 2026-06-26
Source PO cost codes from GP per job, and fix the hardcoded cost element


Xilous:
cost codes are defined per job in GP (JC00701), and the third segment (Cost_Element) varies by cost code - on job 80003 it is 210-200 element 2, 310-000 element 3, 510-000 element 5, 610-000 element 6. the Create PO dialog currently hardcodes element 2 on every cost code, which is right only for the 2xx codes and would be rejected or mis-booked by GP for the rest. cost codes have to come from GP per job, not the static issue #121 list.

- a new relay endpoint GET /cost-codes?company=&job= reads JC00701 for the job and returns each cost code (the two-segment number, the description, and the real Cost_Element), bearer auth and read-only, the same shape as GET /vendors and GET /buyers.
- relayClient gets a getRelayCostCodes(company, job) wrapper.
- the Create PO dialog replaces the static cost-code list with a dropdown populated from the relay for the selected project's job (project_id is the job number), and builds the relay /po cost_code from GP's element for that code rather than a hardcoded 2.
- a stock PO with no project carries no cost code and stays non-inventoried.

this is done once picking a project loads that job's cost codes from GP and a PO created with e.g. 310-000 books to element 3 in GP (check the WS10101 / JC line), while a job with no cost codes shows a clear empty state.

independent and runnable alongside the others. it touches the relay and CreatePODialog.tsx, which the lifecycle-rename and draft-registration issues also touch, so coordinate merges there.

Xilous:
done in PR #181 (branch relay-validate-job-cost-code), commit ac4af4b. cost codes now come from GP per job, and the /po cost element is the code's real one instead of a hardcoded 2.

the relay reads them live off JC00701:
- new GET /cost-codes?company=&job=, bearer auth and read-only, the same shape as GET /vendors and GET /buyers. reads the job's active rows (WS_Inactive = 0) and returns one entry per code: cost_code (the two-segment number e.g. 310-000), description (Cost_Code_Description), and the integer cost_element. a job with no cost codes returns an empty list.
- JC00701 turned out to carry a real Cost_Code_Description and a per-row Cost_Element (confirmed against the live TUBC schema), so nothing about the description is hardcoded.
- list_cost_codes in econnect.py is the read; CostCodeOut / CostCodesResponse are the response models.

on the UI side relayClient gains getRelayCostCodes(company, job), and CreatePODialog drops the static issue #121 list and the hardcoded element 2:
[...]

PR #181, opened 2026-06-29
relay: validate job number and cost code against GP before creating a PO

### LINKED ISSUE #173: Relay: validate job number and cost code against GP before creating a PO
GP rejects an unregistered buyer (POP00101), which the relay now pre-checks. GP also has a job master and per-job cost codes, so a /po with a job that is not a real GP job, or a cost code not set up on that job, should fail with a clear relay error instead of a raw eConnect one - mirroring the buyer check.

- in the relay /po path, before the eConnect orchestration: validate job_number against the GP job master (JC00102) for job-cost lines, and validate each cost code (segments + element) exists on that job in JC00701. return clean errors such as job_not_registered and cost_code_not_on_job, like buyer_not_registered.

this is done once a /po with a made-up job or a cost code not on the job returns a clear 400 naming the problem, and a valid one still succeeds.

relay-only and parallel-safe. it touches the relay main.py / econnect.py additively, the same files as the cost-codes issue.

Xilous:
closes #173.

GP rejects an unregistered buyer (POP00101), which the relay already pre-checks and turns into a clean buyer_not_registered 400. GP also has a job master and per-job cost codes, so a /po carrying a made-up job or a cost code not set up on that job was failing with a raw eConnect error mid-transaction. this adds the matching pre-checks so those cases fail fast with a clear relay error instead.

relay-only, additive, read-only, and parallel-safe. touches the same two files as the cost-codes work (main.py / econnect.py).

in econnect.py:
- split_cost_code(cost_code) - pulled the '210-200-2' -> (cc1, cc2, '', '', cost_element) parse out of apply_wennsoft_integration into a shared function, and refactored the wsi call to use it. the pre-check and the wsi call now split the cost code with the SAME code, so the validated split and the applied split can't drift.
- job_exists(conn, job) - read-only COUNT(*) against the job master JC00102 on WS_Job_Number.
- cost_code_on_job(conn, job, cost_code) - read-only COUNT(*) against the cost-code detail master JC00701, matching the exact six-column key the WennSoft proc joins on (WS_Job_Number + Cost_Code_Number_1..4 + Cost_Element). the key shape comes straight from JCSubContractorDetail.sql's JC00701 join, so the pre-check accepts exactly what the proc would.

in main.py /po, right after the buyer check and before the orchestration:
- loops the PI=2 lines and raises clean 400s - job_not_registered when the job isn't in JC00102, cost_code_not_on_job when the code isn't on that job in JC00701.
- job existence is cached per-request so a PO repeating a job across lines doesn't re-query.
- PI=1 lines are skipped (the model validator already guarantees PI=2 carries both job + cost code and PI=1 carries neither).

behaviour is unconditional, like the buyer check (no config flag). job_exists accepts any job present in JC00102 with no active/open-only filter, because that mirrors what the wsi proc itself accepts.

done criteria from the issue are met: a /po with a made-up job or a cost code not on the job returns a clear 400 naming the problem, and a valid one still flows into the existing 5-step orchestration unchanged.

new tests in tests/test_job_validation.py cover the split_cost_code parse (trailing digit is Cost_Element, bad element defaults to 0) plus fake-cursor tests asserting job_exists / cost_code_on_job hit the right tables, bind the right params, and return the right boolean off the count. full relay suite is 28 passed, ruff clean.

also folds in #177 (receive against a PO through the relay), wired the same two-call way as Create PO: post the GP receipt via the relay first, then record the UC Nexus receive. a receive has to reach GP to count, just like a PO. this part is backend + frontend rather than relay-only, but it lives in entirely different files from the validation work above, so the two don't overlap.

a /receipt needs two things about the PO that UC Nexus wasn't storing yet:
- gp_company on the PO - which GP company the PO lives in (TUBC/TUCSH). recorded at registration; recordPoGpSync now carries it alongside the returned po_number.
- gp_line_ord per line - GP's POP10110.ORD (16384, 32768, ...). there is no relay PO-read endpoint, so it is recorded at create time: create_po stamps ORD = line index * 16384 in send order, and Create PO sends the GP lines and the UC Nexus lines from the same array in the same order, so the mapping is authoritative.
- migration 038 adds both columns (nullable), and gpCompany / gpLineOrd are exposed on the GraphQL types + mappers.

relayClient.postRelayReceipt posts the receipt and surfaces the relay's qty_exceeds_remaining / line_not_receivable / po_not_found / po_line_not_found.

ReceiveModal now reflects the three calls made on the issue:
- the relay must be up to receive at all - a banner blocks the receive when the relay isn't detected, since the receive posts the GP receipt.
- only GP-registered POs can be received - a PO with no GP number + company is blocked with a warning naming it.
- put-away is mandatory now: every received unit lands in a bin, which also gives GP a non-empty rack location per line.
- one relay line per PO line carrying the total received qty (not one per bin), so the relay's remaining-quantity over-receipt check stays authoritative; the per-bin placement + deficient counts stay UC-Nexus-side via createReceive.

one consequence worth knowing: POs created before this change have no gp_company / gp_line_ord, so they read as not-registered and can't be received until recreated through Create PO. fine for the dev/test data.

frontend tsc + vite build are clean, eslint clean on the touched files; the backend schema builds and imports clean (the db-backed tests skip locally with no postgres).

also folds in #178 (package the relay for deployment): a logon auto-start exe plus a DPAPI-protected shared_secret at rest. that part lives in its own set of files (dpapi.py, cli.py / __main__.py / relay_entry.py, the install scripts, ucnexus-relay.spec, the relay-release workflow, docs/relay-deployment.md) and doesn't touch the validation or receive work above. i wrote the full rundown of that work as a comment on the #178 issue rather than duplicating it here, so check that issue for the details.


Xilous:
folded the #179 relay hardening into this branch as commit f222fef, rebased onto the relay-receipt commit (be412d1) with no conflicts since it's relay-only and in different files from the receive work. the same writeup is on the #179 issue; recording it here too.

the PO-number guard now checks history, not just the active table. assert_po_number_available (which only hit POP10100) is replaced by a read-only po_number_in_use(conn, po_number) that checks all three places a PO number can live and short-circuits on the first hit, active-first so the common path does no extra history reads:
- POP10100.PONUMBER - an active (open/work) PO
- POP30100.PONUMBER - a historical (closed/posted/voided) PO
- POP30300.VNDDOCNM - a posted PO receipt's vendor-doc number (= the PO it received against; POP30300 has no PONUMBER column of its own)

/po rejects a reused client-supplied number with a clean 400 po_number_taken that names where the collision is, instead of the old 502 econnect_error that only surfaced mid-orchestration.

every error the relay raises now returns the same envelope. errors.error_body(error, message, **context) is the single shape and all raised errors go through it, so a client always sees {error, message, context}.
- main.py routes all of company_not_allowed, buyer_unresolved, buyer_not_registered, job_not_registered, cost_code_not_on_job, po_number_taken, po_not_found, po_line_not_found, line_not_receivable, qty_exceeds_remaining, sql_error and econnect_error through it.
- auth.py's 401 unauthorized used to be a bare {error} with no message; it carries one now too.
- econnect_error keeps proc / error_state / error_description under context, so the structured eConnect detail is still there for logs; the frontend relayClient reads message first, so the change is non-breaking.
- FastAPI's own 422 request-validation errors keep their framework array shape, which relayClient already handles on a separate branch.

USD POs are left untouched, as the issue said - the relay still assumes CAD and multicurrency is a separate later effort.

two new test files cover this. test_po_number_validation.py asserts all three tables are queried on the right columns, each match is labelled, and an active match short-circuits before the history reads. test_errors.py asserts the envelope shape and the 401 envelope on the wire. full relay suite is 36 passed, ruff check clean.

Xilous:
bundled the work for issue 174 into this PR on top of the relay changes, since it touches the same PO files (CreatePODialog, the PO mutations/queries, ReceiveModal).

a PO only exists once it is registered in GP, so the old "Ordered" state really meant "registered in GP". renamed the po_status value ORDERED -> GP_REGISTERED. and because a PO that fails to create in GP is never created in UC Nexus either, there is no persisted FAILED, so gp_sync_status was redundant with the status and is gone.

on the backend, migration 039 renames the enum value via ALTER TYPE ... RENAME VALUE and drops the gp_sync_status column + enum (both directions reversible). POStatus.GP_REGISTERED now stands in everywhere ORDERED was:
- the vendor-confirmed auto-transition + revert, the receiving status gate, mark_po_as_ordered, open POs, and the dashboard/admin/import rollups
- record_gp_sync_result keeps po_number + gp_company (the field this PR added) and the DRAFT -> GP_REGISTERED advance, but no longer takes or sets gp_sync_status. it's only ever called after GP succeeds, so a failed push never reaches it, and recordPoGpSync drops its gpSyncStatus arg.
- POStatistics.ordered became gp_registered

the frontend shows GP-Registered in the status label, chip, filter and the status sort, plus the stat card. the gpSyncStatus chip and every gpSyncStatus query field are removed. ReceiveModal's "is this PO registered in GP" gate now keys on gpCompany + poNumber instead of gpSyncStatus === SYNCED - those two are recorded in the same step that advances the PO to GP-Registered, so it's the same condition.

the resulting lifecycle is Draft -> GP-Registered -> Vendor-Confirmed -> Partially-Received / Closed. manual POs created through the dialog push to GP first and land at GP-Registered; imported PO requests still start at Draft.

ran ruff (clean), the frontend tsc + vite build (clean), eslint (clean, pre-existing warnings only), confirmed the strawberry schema builds and pytest collects. the db-backed tests skip locally with no postgres, and i couldn't run the alembic up/down locally for the same reason, so the CI migration-integrity job is the check on the migration round-trip.

closes #174

Xilous:
folded the issue #172 work into this PR as well (commits ac4af4b + c797422). closes #172.

cost codes now come from GP per job and the /po cost element is the code's real one instead of a hardcoded 2:
- the relay gets a GET /cost-codes?company=&job= read endpoint, bearer auth and read-only, the same shape as GET /vendors and GET /buyers. it reads the job's active JC00701 rows (WS_Inactive = 0) and returns cost_code (the two-segment number e.g. 310-000), description (Cost_Code_Description), and the integer cost_element. a job with no codes returns an empty list. list_cost_codes lives in econnect.py, CostCodeOut / CostCodesResponse in models.py.
- the c797422 follow-up switches /cost-codes onto the error_body envelope this branch already uses (from the #179 hardening), so it matches vendors/buyers instead of the inline error dicts the cherry-pick first brought over.
- relayClient gains getRelayCostCodes(company, job), and CreatePODialog drops the static issue #121 list and the hardcoded element 2. when a project is picked and the relay is up it loads that job's codes (project_id IS the job number), reloads on project/company change, and each option's value is GP's full phase-step-element (e.g. 310-000-3) so the /po cost_code carries the code's own element. a stock PO with no project carries no cost code.

this overlaps the same CreatePODialog.tsx / main.py / econnect.py that the lifecycle-rename (#174) and the #173 validation touch, but the cherry-pick reconciled clean and the suite is green on the branch, relay ruff plus 47 tests pass and frontend build plus lint clean. verified read-only against live GP. GET /cost-codes?company=TUBC&job=80003 returns the 7 real codes with the right elements (210-200 -> 2, 310-000 -> 3, 510-000 -> 5, 610-000 -> 6).

Xilous:
#176 is in on this branch (commit d491a02). the PO page now finds the relay as it loads instead of waiting for the Create PO dialog to open, so a user knows up front whether GP actions will work.

- a useRelayStatus hook probes /health on mount, polls every 10s, and re-checks when the tab regains focus/visibility. the header shows a status chip (checking relay… / relay connected vN / GP relay not detected on this machine) with no dialog open, and it flips on its own when the relay is stopped and restarted, no page reload.
- the first successful health check warms the relay credential (prefetchRelaySecret) so the first GP action doesn't pay for the relayCredential round-trip. no credential call happens while the relay is down.
- Create PO is driven off that state - the button is disabled with a tooltip when the relay isn't detected, since every PO is GP-first and can't be created without the relay.

the relay state is the page's single source of truth and is passed into CreatePODialog as a relayHealth prop, so the dialog drops its own /health probe and reflects the same status. the in-dialog chip and the existing GP-required validation gate now key off the page state, and the dialog still self-probes when used without the prop. the #175 Register action isn't on this branch yet but can read the same hook when it lands.

new files are frontend/src/relay/useRelayStatus.ts and RelayStatusChip.tsx; relayClient.ts gained prefetchRelaySecret; po/index.tsx and CreatePODialog.tsx are wired to the hook. frontend lint and build are clean.

one thing remains before this is verified end to end - simulated testing on /app/po against Railway with the local relay started and stopped, to watch the chip and the Create PO button flip. that needs this branch deployed.

closes #176

Xilous:
closes #175. an imported Draft PO can now be accepted into GP from the PO list (the import path), landing in the exact same end state as a manually created PO since both share the GP-first push.

clicking "Register in GP" on a Draft PO row, or the same button now in the PO detail modal, opens the renamed GpPurchaseOrderDialog in register mode:
- the project is locked to the draft's project
- a GP-vendor selector defaults to a best-guess name match against the synced GP vendors, with an "Imported as: …" hint so the user confirms the mapping before registering
- the GP section carries relay status, company, buyer, and the GP-sourced cost code
- the draft's line items are prefilled and fully editable (add / remove / edit)

on submit it runs GP-first: relay /po, then on success it maps the chosen GP vendor + cost code onto the PO, records GP's returned po_number + company, replaces the line items with what was pushed, and advances Draft -> GP-Registered. a GP failure changes nothing and the PO stays Draft.

the new registerPoInGp mutation + register_po_in_gp repo do the acceptance atomically in one commit: validate the PO is DRAFT, set vendor + cost_code, reconcile line items by id (update kept, create added, delete removed), assign gp_line_ord positionally in the pushed order so a later relay /receipt can target each GP line, stamp po_number + gp_company, and advance the status with ordered_at. removed lines null out HardwareItem.po_line_item_id before delete (the FK is RESTRICT), so the imported hardware rows are orphaned rather than blocking the delete - the accepted cost of editing the imported line set. backed by RegisterPOInput / RegisterPOLineItemInput.

on the frontend CreatePODialog is renamed to GpPurchaseOrderDialog with a single shared GP-first push path branching to create-or-register, so the manual Create PO submit and the draft registration go through the same relay /po call. the Register in GP action appears on Draft rows in the PO list and in the PO detail modal, both gated on the relay being up. the detail modal's old "Mark as Ordered" button is replaced by Register in GP - marking a draft ordered without an actual GP push would produce a GP_REGISTERED PO that does not exist in GP, contradicting the GP_REGISTERED invariant. the markPoAsOrdered mutation/backend is left in place but no longer surfaced.

test_register_po_in_gp.py covers keep-all, add/edit/remove, gp_line_ord assignment, status advance, hardware-item orphaning, and non-draft rejection. ruff check/format are clean, the strawberry schema builds with the new types, the full backend suite collects (db tests skip locally and run in CI), and frontend lint + build + vitest all pass.

this is stacked on the #172 cost-codes commit already in this PR, so it inherits the job-sourced cost-code dropdown the register dialog needs.

Issue #200, opened 2026-07-07
drop the vendor mirror: remove syncGpVendors + Vendor.gp_vendor_id, PO stores GP vendor id + name snapshot


Xilous:
Drop vendor mirror.
remove syncGpVendors + Vendor GP-mirror role (gp_vendor_id)
POs store GP vendor id + name snapshot for display
vendors read live via gpVendors everywhere else
model + migration change; dev data disposable

depends on server-side-PO slice.

acceptance criteria
- remove syncGpVendors mutation, GpVendorInput, SyncGpVendorsResult
- drop Vendor.gp_vendor_id column
- purchase_orders store GP vendor id + vendor-name snapshot for display
- alembic migration adds the PO columns, drops vendors.gp_vendor_id; alembic check clean; up / down / up passes

relay machinery final teardown, moved from #197. #200 last consumer of browser relay client.
- remove getRelayVendors + checkRelayHealth from vendor-sync path (GpVendorSyncPage)
- delete frontend/src/relay/relayClient.ts
- remove backend relay_credential field + GET_RELAY_CREDENTIAL query
- delete useRelayStatus.ts + remaining handling:
  Chrome LNA
  loopback
  mixed-content
end: zero references to browser relay client anywhere. app builds.

files likely touched
backend/app/schemas/mutations.py -> remove sync_gp_vendors
backend/app/schemas/inputs.py -> remove GpVendorInput
backend/app/schemas/types.py -> remove SyncGpVendorsResult; PO vendor-snapshot fields
backend/app/models/vendor.py -> drop gp_vendor_id
backend/app/models/purchase_order.py -> add gp_vendor_id + vendor_name_snapshot
backend/app/repositories/vendor_repository.py -> remove sync_gp_vendors
backend/app/repositories/po_repository.py -> store GP vendor id + name snapshot
backend/alembic/versions/ -> new migration
frontend/src/modules/po/GpVendorSyncPage.tsx -> remove relay vendor-sync
frontend/src/relay/relayClient.ts -> delete
frontend/src/relay/useRelayStatus.ts -> delete
backend/app/schemas/queries.py -> remove relay_credential
frontend/src/graphql/queries.ts -> remove GET_RELAY_CREDENTIAL

carved from #186.

Xilous:
implemented and squash-merged into the draft pr https://github.com/Upper-Canada-Specialty-Hardware/uc-nexus/pull/201 as commit `f848cff`.
[...]

Issue #199, opened 2026-07-07
PO create/register + receive brokered server-side via relay_call


Xilous:
move GP-first orchestration out of browser into resolvers:
createPo
registerPoInGp
receive

each resolver calls relay_call for the GP write first -> records the UC Nexus row. GP-first invariant, server-side.
first real GP write over the socket end to end.

acceptance
- createPo / registerPoInGp / receive do the GP write via relay_call first -> persist the UC Nexus row.
- buyer from gpBuyers; relay per-workstation buyer fallback no longer decides it.
- received_by = acting UC Nexus user from Clerk token, not the relay Windows account.
- frontend Create-PO dialog + Receive modal stop calling relay directly -> call these resolvers.
- one job-cost PO against a TUBC job lands in GP; UC Nexus PO shows GP-Registered with GP's returned number; receipt posts to GP.

relayClient teardown, from #197.
precondition: createPo + receive brokered server-side.
delete from frontend/src/relay/relayClient.ts:
postRelayPo
postRelayReceipt
drop their browser-direct calls in:
GpPurchaseOrderDialog.tsx
ReceiveModal.tsx
keep relayClient.ts. getRelayVendors + checkRelayHealth back vendor-sync until #200.
end state:
- app builds
- PO-create + receive run through backend

files likely touched
backend/app/schemas/mutations.py (create_po, register_po_in_gp, create_receive)
backend/app/schemas/inputs.py (buyer on PO input; received_by from context)
backend/app/repositories/po_repository.py
backend/app/repositories/receiving_repository.py
frontend/src/modules/po/GpPurchaseOrderDialog.tsx (call backend, drop relay client)
frontend/src/modules/warehouse/ReceiveModal.tsx (call backend receive, drop relay client)
frontend/src/relay/relayClient.ts (remove postRelayPo + postRelayReceipt)

depends on live-GP-reads slice (relay_call, gpBuyers).
carved from #186.

Xilous:
implemented and squash-merged into the draft pr https://github.com/Upper-Canada-Specialty-Hardware/uc-nexus/pull/201 as commit `d9144f9`.
[...]

Issue #198, opened 2026-07-07
projects sourced from GP: adopt a GP job, live gpJobs picker, drop free-form create_project


Xilous:
project picker + import target -> live gpJobs query.

create a project = adopt a GP job:
pick from gpJobs
store GP job number as project identity
snapshot display name at adopt time

free-form create_project removed (arbitrary project_id / description).

openings, hardware, rack data still anchor to the project row unchanged.

depends on live-GP-reads slice (needs gpJobs).

files likely touched
backend/app/schemas/mutations.py (create_project -> adopt-a-GP-job)
backend/app/repositories/project_repository.py
backend/app/schemas/inputs.py (CreateProjectInput -> adopt input)
frontend/src/modules/import/ (project picker + import-target selection)

carved from #186.

Xilous:
implemented and squash-merged into the draft pr https://github.com/Upper-Canada-Specialty-Hardware/uc-nexus/pull/201 as commit `ddb7f3f`.
Implemented issue #198: project creation is now "adopt a GP job" instead of free-form entry. Backend - CreateProjectInput/create_project/createProject replaced end-to-end by AdoptGpJobInput (job_number, optional job_name) / project_repository.adopt_gp_job / the adoptGpJob mutation, which sets project_id to the GP job number (immutable identity) and description to a one-time snapshot of the job name; client and other editable fields stay null until set via the existing update flow. ConflictError still guards against adopting the same job twice. Added backend/tests/test_project_repository_adopt.py and fixed the one other repository-level test helper that used the old signature. Frontend - added the GET_GP_JOBS query and ADOPT_GP_JOB mutation, replaced CreateProjectDialog.tsx with AdoptGpJobDialog.tsx (GP company selector + relay status chip + Autocomplete job picker sourced live from gpJobs, wired through the existing relay_gateway path used by the PO/vendor GP screens), and updated the Import module's landing page button/empty-state text accordingly. Updated testing/CLAUDE.md's project-seeding note to document the new mutation and that the UI dialog now requires a live relay connection. Verified: backend schema builds cleanly with no createProject/CreateProjectInput remnants, ruff check and the full non-DB pytest suite pass; frontend tsc -b and eslint pass clean on all touched files. Changes are committed on issue-198 (not pushed/PR'd, per contract).

Xilous:
delivered by #201 (merged 2026-07-08), commit ddb7f3f "issue 198: squash-merge issue-198 into mux/gp-relay-central".

[...]

Issue #197, opened 2026-07-07
live GP reads through the backend: gpJobs/gpVendors/gpBuyers/gpCostCodes + relayStatus, delete browser relay client


Xilous:
expose GP catalog reads as backend GraphQL fields via relay_call.
cut browser direct line to relay.

depends on relay-to-backend channel slice (relay_call + live connection).

new GraphQL query fields go through relay_call:
gpJobs(company) -> JC00102
gpVendors(company) -> PM00200
gpBuyers(company) -> POP00101
gpCostCodes(company, job) -> JC00701

relayStatus reports channel connected, from backend gateway.

Create-PO dialog dropdowns populate from these fields, not browser relay client:
vendor
buyer
cost-code

delete relayClient.ts
delete useRelayStatus.ts per-tab health poll
delete relayCredential fetch
delete Chrome LNA / loopback / mixed-content handling
remove backend relay_credential field
remove GET_RELAY_CREDENTIAL query

RelayStatusChip reads backend relayStatus (connected / offline), no local health probe.

files likely touched
backend/app/schemas/queries.py -> add gpJobs/gpVendors/gpBuyers/gpCostCodes + relayStatus; remove relay_credential
backend/app/schemas/types.py -> GP read result types + RelayStatus
frontend/src/relay/relayClient.ts -> delete
frontend/src/relay/useRelayStatus.ts -> delete
frontend/src/relay/RelayStatusChip.tsx -> read backend relayStatus
frontend/src/graphql/queries.ts -> remove GET_RELAY_CREDENTIAL; add GP reads + relayStatus
frontend/src/modules/po/GpPurchaseOrderDialog.tsx -> dropdowns off backend fields

carved from #186.

Xilous:
implemented and squash-merged into the draft pr https://github.com/Upper-Canada-Specialty-Hardware/uc-nexus/pull/201 as commit `e47a456`.
[...]

PR #201, opened 2026-07-07
mux effort: mux/gp-relay-central

Xilous:
mux tracking PR for gp-relay-central. squash-merges one commit per issue (#196, #197, #198, #199, #200) onto mux/gp-relay-central. +2390 -1522 across 57 files.

re-architects how UC Nexus reaches Microsoft GP through the on-prem relay.

previously browser -> relay direct at http://localhost:7321 (Chrome LNA loopback), Bearer secret fetched from backend via relayCredential.
now relay dials OUT -> backend, persistent wss:// at /relay-link, auth via enrolled secret on connect handshake.
RelayGateway singleton tracks the one live socket.
relay_call(company, op, payload) sends {id, op, company, payload} -> awaits correlated {id, ok, result|error}.
GraphQL resolvers broker GP server-side.

transport / foundation (#196, #197)
- backend/app/services/relay_gateway.py -> RelayGateway + relay_call (correlation, 30s timeout, supersede old connection, fail pending on disconnect).
- errors.py adds RelayUnavailableError, RelayTimeoutError, RelayCallError.
- backend/main.py -> /relay-link WS route. connect handshake authed via relay_repository.authenticate_secret (replaces get_credential). frontend no longer fetches the raw secret.
- localhost relay/src/ucnexus_relay/channel.py -> outbound wss client, reconnect exponential backoff. dispatches ops on a worker thread:
  list_vendors
  list_buyers
  list_cost_codes
  list_jobs
  create_po
  create_receipt
- localhost relay/src/ucnexus_relay/ops.py -> transport-agnostic create_po_op / create_receipt_op, shared by inbound HTTP routes (main.py) + WS channel so they can't drift. RelayOpError for validation failures.
- relay cli.py runs channel alongside existing HTTP server. blank [channel].backend_url keeps it HTTP-only.

adopt GP job (#198)
- createProject / CreateProjectInput removed.
- project adopted from a live GP job (JC00102) via adoptGpJob(input: AdoptGpJobInput). project_id = jobNumber (immutable identity). description = jobName snapshotted at adopt time.
- new gpJobs(company) query + AdoptGpJobDialog (replaces CreateProjectDialog). AdoptGpJobDialog needs a connected relay. adoptGpJob does not call the relay.

GP-first, server-side (#199)
- createPo / registerPoInGp / createReceive push to GP via relay_call BEFORE persisting, so a GP rejection changes nothing in UC Nexus. no numberless DRAFT window.
- received_by resolved from the Clerk token server-side (auth.resolve_display_name), not a client string and not the relay's Windows account. CreateReceiveInput.received_by removed.
- new gpBuyers(company) / gpCostCodes(company, job) queries. frontend posts line data; backend builds the relay payload (services/gp_po.py).

drop the vendor mirror (#200)
- migration 040 adds purchase_orders.gp_vendor_id + vendor_name_snapshot, drops vendors.gp_vendor_id.
- syncGpVendors mutation + GpVendorSyncPage removed. GP vendor picked live from gpVendors(company) at push time, snapshotted onto the PO for display. vendor_id stays an optional link to a UC Nexus vendor record (contact info), no GP meaning.
- new gpVendors(company) query.

relay status
- browser /health probe removed (relayClient.ts, useRelayStatus.ts).
- relayStatus query reads gateway.connected (the relay-to-backend WS channel). RelayStatusChip + PO/receive dialogs poll every 10s.

known limitations / follow-ups
- pyproject pins websockets ^13.0 (lock 13.1) but channel.py calls websockets.connect(additional_headers=...), the 14.0+ top-level param; on 13.x connect is the legacy client (extra_headers). confirm a live relay-to-backend connection.
- RelayGateway is per-process in-memory. backend must run one process / one replica. relay_call + relayStatus break on any request that lands on a replica without the socket.
- createReceive / createPo not atomic across GP and Postgres. DB failure after the GP write orphans the GP receipt/PO. old client-side receive retry-idempotency guard (gpReceiptPostedRef) removed. failure toasts currently say "nothing was changed" even for a post-GP-write failure.
- a relay enrolls for one company. PO/receive/adopt dialogs still offer TUBC + TUCSH, so the other company fails at submit.

tests
- backend covers test_relay_gateway, test_gp_po, test_gp_relay_reads, test_project_repository_adopt, test_relay_installs (authenticate_secret). plus updated test_register_po_in_gp / test_relay_gp_fields.
- relay test_channel (op dispatch). channel connect path (_run_once) not exercised by CI.


Xilous:
issue #196 addressed by commit `1b1ace4`.

Xilous:
issue #197 addressed by commit `e47a456`.

Xilous:
issue #198 addressed by commit `ddb7f3f`.

Xilous:
issue #199 addressed by commit `d9144f9`.

Xilous:
issue #200 addressed by commit `f848cff`.

localhost relay/src/ucnexus_relay/channel.py
```diff
@@ -0,0 +1,180 @@
...
+        logger.exception("unhandled error dispatching op", extra={"op": op, "id": job_id})
+        reply = {"ok": False, "error": errors.error_body("internal_error", str(e))}
+    reply["id"] = job_id
+    return reply
+
+
+async def _run_once(url: str, secret: str, cfg) -> None:
+    async with websockets.connect(
+        url,
+        additional_headers={"Authorization": f"Bearer {secret}"},
```
Xilous: confirmed feature-breaker on the pinned version.
websockets pinned `^13.0` in pyproject, locked at 13.1.
this calls `websockets.connect(additional_headers=...)`.
on 13.x the top-level `websockets.connect` is the legacy client, keyword `extra_headers`; `additional_headers` is the 14.0+ name.
on 13.1 this raises TypeError -> `run_forever`'s `except Exception` swallows it -> retries forever -> the outbound channel never connects -> every relay_call fails with RelayUnavailableError:
gp_jobs
gp_vendors
gp_buyers
gp_cost_codes
create_po
register_po_in_gp
create_receive
poetry.lock resolves websockets 13.1.
use `extra_headers` while the pin stays on 13.x, or bump the pin to websockets >=14 and keep additional_headers.
applied in the review working tree, switched to extra_headers.

backend/main.py
```diff
@@ -2,16 +2,19 @@
...
 from app.auth import get_context
+from app.database import SessionLocal
 from app.errors import AppError
+from app.repositories import relay_repository
 from app.schemas.mutations import Mutation
 from app.schemas.queries import Query
+from app.services.relay_gateway import gateway as relay_gateway
 
 
 class ErrorHandlerExtension(SchemaExtension):
```
Xilous, line 21: ErrorHandlerExtension.resolve is synchronous.
#199/#200 made seven resolvers async:
create_po
register_po_in_gp
create_receive
gp_jobs
gp_vendors
gp_buyers
gp_cost_codes
for an async resolver `_next(...)` returns a coroutine strawberry awaits AFTER this method returns (via await_maybe) -> the try/except never sees the exception.
an AppError from any of those resolvers reaches the client with its message but WITHOUT extensions.code / field, and RelayCallError.detail is dropped -> breaks the AppError-to-code contract.
a ValidationError from a sync resolver returns extensions {code, field}; from an async resolver returns {}.
fix applied in the review working tree, keep the sync path; when `_next` returns an awaitable, await it inside an async wrapper running the same AppError-to-GraphQLError mapping.

backend/app/schemas/mutations.py
```diff
@@ -594,6 +672,38 @@ def create_receive(self, input: CreateReceiveInput) -> ReceiveRecord:
...
+                        "locations": li["locations"],
+                    }
+                )
+
+            payload = gp_po.build_create_receipt_payload(
+                po_number=po.po_number,
+                received_by=received_by,
+                line_items=receipt_line_items,
+            )
+            await relay_gateway.relay_call(po.gp_company, "create_receipt", payload)
```
Xilous: GP-first, no idempotency, no compensation.
relay_call posts the GP receipt here -> warehouse_repository.create_receive + session.commit run after.
commit or a repo-side check fails -> GP receipt already posted, UC Nexus records nothing.
ReceiveModal shows 'Receiving <PO> failed, so nothing was recorded' - false.
old client-side gpReceiptPostedRef dedup (retry finished the UC record WITHOUT re-posting GP) removed, nothing replaces it -> retry posts a SECOND GP receipt (double-count) or trips GP's over-receipt rejection.
eligibility validated only inside warehouse_repository.create_receive, which runs AFTER the relay post:
PO status
quantity <= pending
location sums
received_by length
an over-receive or a receive against an already-complete PO posts a GP receipt on the first attempt before UC rejects it.
not fixed here; needs an idempotency key on the receipt op plus validate-before-push. highest-risk item.
at minimum, run the eligibility checks before relay_call.

backend/app/schemas/mutations.py
```diff
@@ -436,13 +480,42 @@ def register_po_in_gp(self, input: RegisterPOInput) -> PurchaseOrder:
...
+            payload = gp_po.build_create_po_payload(
+                vendor_gp_id=input.gp_vendor_id,
+                vendor_contact_name=vendor_contact_name,
+                buyer_id=input.buyer_id,
+                job_number=job_number,
+                cost_code=input.cost_code,
+                po_number=None,
+                line_items=line_items_data,
+            )
+            gp_result = await relay_gateway.relay_call(input.gp_company, "create_po", payload)
```
Xilous: same GP-first non-atomicity on the PO push; create_po at line 419 has the identical shape.
relay_call creates the GP PO before the repo write.
po_repository.register_po_in_gp / create_po then raises:
InvalidStateTransitionError on a double-submit
ValidationError from _assert_po_number_available
any IntegrityError
-> session rolls back, GP PO already exists with no UC linkage, an orphan.
retry sends po_number=None -> GP reserves a fresh PONUMBER -> retry duplicates the GP PO.
frontend toast 'GP did not accept the PO, so nothing was changed' is wrong here.
not fixed here; needs the same push-then-persist compensation / idempotency as create_receive.

backend/app/schemas/mutations.py
```diff
@@ -436,13 +480,42 @@ def register_po_in_gp(self, input: RegisterPOInput) -> PurchaseOrder:
...
+            if vendor_id is not None:
+                vendor = session.get(VendorModel, vendor_id)
+                if vendor is None:
+                    raise NotFoundError(f"Vendor {vendor_id} not found")
+                vendor_contact_name = vendor.contact_name
+
+            job_number = None
+            if po.project_id is not None:
+                project = session.get(ProjectModel, po.project_id)
+                job_number = project.project_id if project else None
```
Xilous: asymmetric with create_po. a missing project row here silently sets job_number=None; create_po (line 407) raises NotFoundError for the same condition.
a project-linked PO whose project row is absent gets pushed to GP as a non-job PO -> silently mis-books cost accounting, no error.
build_create_po_payload keys product_indicator on job_number -> every line becomes non-inventoried, cost_code dropped.
applied in the review working tree, raise NotFoundError when the project is missing, matching create_po.

backend/app/schemas/queries.py
```diff
@@ -1011,22 +1035,46 @@ def vendor(self, id: strawberry.ID) -> Vendor | None:
...
     def relay_installs(self, info: strawberry.Info) -> list[RelayInstallInfo]:
         require_admin(info)
         with SessionLocal() as session:
             return [_relay_install_to_type(ri) for ri in relay_repository.list_installs(session)]
 
+    @strawberry.field
+    def relay_status(self, info: strawberry.Info) -> RelayStatus:
+        """Whether the outbound relay WS channel is currently connected, for the relay status chip."""
+        require_user(info)
+        return RelayStatus(connected=relay_gateway.connected)
```
Xilous: relay_status returns only `connected` (any relay, any company).
relay_call rejects an op whose company != the enrolled install's company (relay_gateway.py:66).
AdoptGpJobDialog and GpPurchaseOrderDialog both offer TUBC + TUCSH.
connected relay enrolled for one company, user picks the other -> RelayStatusChip shows green 'connected', every gp_* read returns empty, final submit fails with RelayUnavailableError('enrolled for X, not Y').
the gateway already knows the connected company (self._company), never surfaces it.
not applied; needs a schema + frontend change. add the connected company to RelayStatus and drive the dialog's company selection from it, or disable the mismatched company.

localhost relay/src/ucnexus_relay/cli.py
```diff
@@ -13,17 +13,31 @@
...
+    # unused; it's unrelated to `channel`, which is an outbound `websockets` CLIENT run alongside this
+    # server below (a `websockets` build IS bundled for that).
+    config = uvicorn.Config(app, host=s.server.host, port=s.server.port, loop="asyncio", http="h11", ws="none")
+    server = uvicorn.Server(config)
+
+    async def _run() -> None:
+        # the outbound channel is additive to the existing inbound HTTP server - a blank
+        # [channel].backend_url makes channel.run_forever() a no-op, so this is safe on a relay that
+        # hasn't been reconfigured for it yet.
+        await asyncio.gather(server.serve(), channel.run_forever())
```
Xilous: relay hangs on shutdown.
channel.run_forever loops indefinitely (reconnect-with-backoff, stop_event is None here) -> after uvicorn's server.serve() returns on SIGINT / service stop, gather keeps awaiting run_forever, asyncio.run never returns -> process must be force-killed, service-manager stop times out.
only bites once [channel].backend_url is set, i.e. when the new channel is in use.
applied in the review working tree, run the channel as a task and cancel it in a finally once server.serve() returns (run_forever already re-raises CancelledError).

backend/app/schemas/mutations.py
```diff
@@ -382,15 +393,42 @@ def create_po(self, input: CreatePOInput) -> PurchaseOrder:
...
+            payload = gp_po.build_create_po_payload(
+                vendor_gp_id=input.gp_vendor_id,
+                vendor_contact_name=vendor_contact_name,
+                buyer_id=input.buyer_id,
+                job_number=job_number,
+                cost_code=input.cost_code,
+                po_number=input.po_number,
+                line_items=line_items_data,
+            )
+            gp_result = await relay_gateway.relay_call(input.gp_company, "create_po", payload)
```
Xilous: this resolver holds an open SessionLocal (a checked-out Postgres connection) across `await relay_call`, which can take up to the 30s relay timeout.
under a slow relay or a few concurrent PO creates/receives, every in-flight mutation pins one pool connection waiting on the relay -> pool exhausts -> all other GraphQL requests block on checkout.
the pool-starvation signature the repo's perf rules warn about, driven by relay latency.
these resolvers are async but call synchronous SQLAlchemy plus a synchronous Clerk HTTP call (resolve_display_name), which block the single event loop that also runs the /relay-link read loop -> a slow DB/Clerk call delays reply delivery for other in-flight relay_calls.
not fixed here; needs restructuring - gather payload data -> release the session -> call the relay -> persist; and offload the sync DB/Clerk work.

backend/app/auth.py
```diff
@@ -122,6 +122,15 @@ def require_user(info) -> dict:
     return {"user_id": user_id}
 
 
+def resolve_display_name(user_id: str) -> str:
+    """Full name (falling back to email, then the Clerk user id) for the acting user - issue #199's
+    server-side received_by, so a receive records the Clerk-authenticated caller, not the relay's
+    Windows account. Mirrors the frontend's useIdentity() fullName-or-email convention."""
+    profile = user_repository.get_user(user_id)
```
Xilous, line 129: resolve_display_name runs on every receive -> calls user_repository.get_user, a synchronous Clerk Backend API round-trip with resp.raise_for_status().
1. any Clerk 5xx/timeout raises httpx.HTTPStatusError, not an AppError -> receive fails with an opaque 500. receiving coupled to Clerk availability.
2. a Clerk profile with no first/last name and no matching primary email returns the raw Clerk user id (e.g. 'user_2abc...') -> written as received_by on both the GP receipt and the UC record.
consider catching the Clerk failure into an AppError and falling back more gracefully than the raw id.

localhost relay/src/ucnexus_relay/channel.py
```diff
@@ -0,0 +1,180 @@
...
+        ping_interval=cfg.ping_interval,
+        ping_timeout=cfg.ping_timeout,
+    ) as ws:
+        logger.info("channel connected", extra={"url": url})
+        async for raw in ws:
```
Xilous: read loop awaits `_handle_job(job)` inline inside `async for raw in ws` -> the next frame isn't read until the current job's reply is sent.
asyncio.to_thread keeps pings flowing but does NOT make jobs concurrent. the docstring's 'a slow GP call doesn't block the read loop' is true only for keepalive, not throughput.
backend fans out several relay_calls on one socket (Create-PO page fires gp_vendors + gp_buyers + gp_jobs + gp_cost_codes at once) -> they run strictly one-at-a-time; a slow create_po ahead of them can push later jobs past the backend's 30s relay_call timeout even though GP is healthy.
not applied; dispatch each job as its own task (asyncio.create_task) so the read loop keeps pulling frames.

backend/app/services/relay_gateway.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def _fail_all(self, message: str) -> None:
+        pending, self._pending = self._pending, {}
+        for future in pending.values():
+            if not future.done():
+                future.set_exception(RelayUnavailableError(message))
+
+    def resolve(self, reply: dict) -> None:
+        """Called by the /relay-link route's read loop with each {id, ok, result|error} reply."""
+        future = self._pending.pop(reply.get("id"), None)
```
Xilous, line 64: two supersede hazards in the single-socket model.
1. dropped reply. a new relay registers -> register() calls _fail_all -> empties _pending, fails in-flight calls with 'superseded'. if a valid reply for one of those calls is already on the wire, resolve() pops nothing and the real result (a completed GP create_po/receipt) is discarded -> caller sees a relay failure for a write that actually committed in GP.
2. two-relay thrash. model doc allows one relay per workstation. two enroll -> each new connection force-closes the other (4409), whose backoff loop immediately reconnects and supersedes back -> infinite flap, relay_call lands non-deterministically.
both inherent to the POC one-socket design. flagging so the single-relay constraint is enforced (or the gateway keyed by install) before multi-workstation use.

backend/app/services/relay_gateway.py
```diff
@@ -0,0 +1,90 @@
...
+        """Send {id, op, company, payload} to the connected relay and await its correlated reply.
+        Raises RelayUnavailableError (no/wrong-company connection or send failure), RelayTimeoutError,
+        or RelayCallError (the relay itself answered ok=false)."""
+        if self._socket is None:
+            raise RelayUnavailableError()
+        if self._company and company != self._company:
+            raise RelayUnavailableError(f"connected relay is enrolled for {self._company}, not {company}")
+
+        job_id = uuid.uuid4().hex
+        future: asyncio.Future = asyncio.get_event_loop().create_future()
```
Xilous: minor. use asyncio.get_running_loop() here.
get_event_loop() is deprecated when there's no running loop (3.12+); if relay_call were ever called outside a running loop it would create the future on a different loop than register()/resolve() run on -> set_result never wakes the awaiter.
inside this async method a running loop always exists, works today.
applied in the review working tree, get_running_loop().

backend/app/repositories/project_repository.py
```diff
@@ -9,20 +9,21 @@
...
 
 
-def create_project(session: Session, project_id: str, description: str, client: str) -> ProjectModel:
-    """Create a new project. Raises ConflictError if project_id already exists."""
-    existing = session.scalars(select(ProjectModel).where(ProjectModel.project_id == project_id)).first()
+def adopt_gp_job(session: Session, job_number: str, job_name: str | None) -> ProjectModel:
+    """Adopt a live GP job (JC00102) as a project. job_number becomes the project's identity
+    (project_id, immutable); job_name is a snapshot of GP's job description at adopt time, not
+    synced afterward. Raises ConflictError if this job has already been adopted."""
+    existing = session.scalars(select(ProjectModel).where(ProjectModel.project_id == job_number)).first()
```
Xilous, line 22: two gaps in the job-number identity.
1. no normalization/guard. adopt_gp_job trusts the caller (the mutation is ungated), uses job_number verbatim as project_id -> adoptGpJob(jobNumber:'') creates an identity-less project; ' 1001 ' dodges the already-adopted check against '1001'. old CreateProjectDialog trimmed client-side; direct callers don't.
applied in the review working tree, strip job_number and reject empty.
2. no company scope. project_id = jobNumber, company discarded -> a job number that exists in two GP companies collides (adopting '1001' from TUCSH after TUBC raises 'already adopted' for a different job); project records nothing about its source company.
not fixed; needs a schema/identity change. flagging as a design limitation.

PR #234, opened 2026-07-13
Capture manufacturer, suggest ordering vendor, write it to the GP PO

### LINKED ISSUE #228: exploration + implementation: vendor and manufacturer gp
gp via nexus relay

apparently vendor and manufacturers are 2 separate entities

Titan hardware schedule contains the manufacturer name - but this is probably not an exact match to the registered profiles in GP

PO users supposedly are able to order the hardware from either the manufacturer or the vendor (actual separate entities within GP?)

UC Nexus needs to allow PO users to do this, but depending on what is true with how GP works

exploration required but only within the confines of TUBC

### LINKED ISSUE #231: manufacturer capture: TITAN manufacturer end to end (parser -> GraphQL -> DB)
capture TITAN manufacturer on every hardware item, carry end to end: parser -> GraphQL schema -> DB. no GP calls. no vendor logic. root slice for #228, no deps; other two slices depend on it.

- parser extracts manufacturer from each product's Material_List_Fields block, next to vendorNo. candidate list Manufacturer then MFG. null when absent.
- manufacturer flows through importSchedule / finalize input -> persisted on HardwareItem.
- round-trips: import schedule with manufacturers -> reload project -> each item shows manufacturer via hydration path.
- item with no manufacturer shows blank, not error.
- migration adds hardware_items.manufacturer nullable string, no backfill. alembic check passes. CI upgrade/downgrade/check green.

files likely touched:
- frontend/src/workers/parserLogic.ts, extractHardwareItems - add manufacturer from Material_List_Fields block next to vendorNo (textOrNull(mlf?.Manufacturer) ?? textOrNull(mlf?.MFG)), push onto each pushed hardware item.
- frontend/src/types/hardwareSchedule.ts, frontend/src/modules/import/types.ts - add manufacturer: string | null to ParsedHardwareItem.
- frontend/src/graphql/queries.ts, frontend/src/graphql/mutations.ts - add manufacturer to hardware-item selection sets and import input.
- frontend/src/modules/import/hydrateSchedule.ts, mapHardwareItem - map manufacturer back from server response.
- frontend/src/modules/import/ReconciliationStep.tsx, frontend/src/modules/import/ClassificationGrid.tsx - optional manufacturer column where vendor_no shows, blank when null.
- backend/app/models/hardware.py, HardwareItem - add manufacturer: Mapped[str | None] = mapped_column(String, nullable=True) next to vendor_no.
- backend/alembic/versions/ - new autogenerated migration adding hardware_items.manufacturer.
- backend/app/schemas/inputs.py, HardwareItemInput - add manufacturer: str | None = None.
- backend/app/graphql/import.graphql, input HardwareItemInput - add manufacturer: String.
- backend/app/schemas/mutations.py, import mutation - add manufacturer to hi_data dict.
- backend/app/repositories/import_repository.py, finalize_import_session - add manufacturer to BOTH HardwareItem constructions (IN_PO path and AVAILABLE path).
- backend/app/schemas/types.py - add manufacturer to HardwareItem type and ProjectScheduleHardwareItem type.
- backend/app/schemas/queries.py, _project_schedule_hardware_item_to_type - add manufacturer; include in repository dict feeding resolver.
- frontend/src/workers/parserLogic.test.ts - fixture with manufacturer element asserted on parsed item, one without asserted null.
- backend/tests/test_import_repository_full_schedule.py - payload with manufacturer persists HardwareItem row, round-trips through schedule query.

TITAN XML element name for manufacturer unconfirmed. confirm against live export, trim candidate list (Manufacturer, MFG) once known.

carved from #228.

### LINKED ISSUE #232: manufacturer -> vendor suggestion + learn-on-confirm
suggest ordering vendor from hardware line's TITAN manufacturer, learn mapping on PO confirm.
no relay change.
depends on #231 (manufacturer captured on HardwareItem).
parallel with #233 - disjoint files.
carved from #228.

- resolve manufacturer -> GP vendor. saved mapping wins, else ranked fuzzy match vs live vendor list.
- persistent mapping table, learns from confirmed POs.
- PO dialog pre-selects suggested vendor, override allowed.

acceptance
suggestVendorForManufacturer(gpCompany, manufacturer)
- saved mapping -> that vendor, score 100, savedMapping true
- no mapping -> top-N live vendors ranked by fuzzy score, savedMapping false

single-manufacturer PO -> pre-select suggested vendor in gpVendors picker, label why, override to any vendor.
after PO persists, each distinct manufacturer on PO lines upserted to chosen vendor, source=confirmed. second PO same manufacturer auto-picks learned vendor.
mapping-write failure never fails PO (savepoint + best-effort).
2+ manufacturers mapping to different vendors -> warn + explicit pick, no pre-select.
manufacturer with no GP vendor -> soft note + manual pick.

files touched
backend/app/models/manufacturer_vendor_map.py (new) - ManufacturerVendorMap:
id (uuid pk)
gp_company
manufacturer_key (normalized)
manufacturer_label
gp_vendor_id (15)
gp_vendor_name
source (confirmed|imported)
created_at
updated_at
unique (gp_company, manufacturer_key)

backend/alembic/versions/ (new migration) - table + unique index.

backend/app/services/manufacturer_match.py (new) - normalize(name):
uppercase
collapse whitespace
strip punctuation
strip trailing legal suffixes (INC, LTD, LTEE, CORP, CO, OF CANADA, CANADA)
same normalized key feeds mapping key + fuzzy compare.
score with rapidfuzz token_set_ratio, fallback difflib.SequenceMatcher, decide in review.

backend/pyproject.toml - add rapidfuzz.

backend/app/repositories/manufacturer_vendor_map_repository.py (new):
lookup(session, gp_company, manufacturer_key)
upsert(session, gp_company, manufacturer_key, label, gp_vendor_id, gp_vendor_name, source)

backend/app/repositories/po_repository.py - create_po + register_po_in_gp. after local persist, for each distinct manufacturer resolved from PO hardware lines (match HardwareItem on project_id + hardware_category + product_code), upsert (gp_company, normalize(manufacturer)) -> chosen gp_vendor_id / vendor_name_snapshot, source=confirmed. wrap in session.begin_nested() + try/except so mapping failure rolls back only mapping, never PO. keep this off mutations.py - that is the file #233 edits, staying off it keeps the two slices disjoint and parallel.

backend/app/schemas/queries.py - suggestVendorForManufacturer resolver. normalize manufacturer, mapping hit -> saved vendor, else fetch live vendor list (gpVendors relay path), rank by token_set_ratio, return top N with scores.

backend/app/schemas/types.py - VendorSuggestion + VendorCandidate (gp_vendor_id, gp_vendor_name, score).

backend/app/graphql/import.graphql - new query + types.

frontend/src/modules/po/GpPurchaseOrderDialog.tsx - on open, compute distinct manufacturer(s) of selected hardware lines.
exactly one -> call suggestVendorForManufacturer, pre-select top candidate, label (suggested: manufacturer = X), list other candidates.
2+ mapping to different vendors -> warning + explicit pick, no pre-select.
empty suggestion -> soft note + manual pick.

frontend/src/graphql/queries.ts - suggestVendorForManufacturer query document.

tests
backend
- normalize() (suffix/case/punct)
- suggestion resolver returns saved vendor on mapping hit, ranked fuzzy candidates otherwise
- learn-on-confirm in po_repository upserts; second suggestVendorForManufacturer returns saved mapping
- mapping-write failure (savepoint) doesn't fail PO
frontend
- dialog pre-selects on single manufacturer, warns on mixed, soft-notes on empty

### LINKED ISSUE #233: write manufacturer onto GP PO line (taPoLine USRDEFND1)
write captured manufacturer onto GP PO line via relay, taPoLine UDF 1 (@I_vUSRDEFND1). option A, default from #228.
depends on #231 (manufacturer captured on HardwareItem).
independent of #232 (Phase 2 vendor-suggestion) - either order.
carved from #228.

thread each line's manufacturer from PO payload -> create_po_line -> @I_vUSRDEFND1 on taPoLine EXEC (RTRIM + truncate 50).

the change must satisfy:
- create_po_line with manufacturer sets USRDEFND1; without -> field blank; over 50 chars truncates.
- existing taPoLine read-back assertion still holds.
- PO with null-manufacturer line succeeds, field blank.
- backend resolves each payload line's manufacturer from matching hardware items (project_id + hardware_category + product_code); disagreeing items -> first non-null + log.

expect these files to change:
- backend/app/schemas/mutations.py - _prepare_create_po + _prepare_register_po resolve each payload line's manufacturer from matching hardware items. line_items_data already carries hardware_category + product_code (the join key), no new frontend input. add to each relay payload line dict, disagreeing items -> first non-null + log.
- localhost relay/src/ucnexus_relay/models.py - POLine add manufacturer: str | None = Field(default=None, max_length=50).
- localhost relay/src/ucnexus_relay/econnect.py - create_po_line add param manufacturer: str | None = None; add @I_vUSRDEFND1 = ? to taPoLine EXEC (RTRIM + truncate 50); bind it.
- localhost relay/src/ucnexus_relay/ops.py - create_po orchestration passes line.manufacturer -> create_po_line.
- localhost relay/tests/ - create_po_line with manufacturer sets USRDEFND1; without -> blank; over 50 truncates; keep taPoLine read-back assertion.
- backend/tests/test_register_po_in_gp.py + backend/tests/test_po_repository_create_po.py - _prepare_create_po / _prepare_register_po attach right manufacturer per line; mixed-item line -> first non-null.

before merge, confirm @I_vUSRDEFND1 unused at this customer - not read or printed on PO. manual GP check user runs.
if not free -> switch to free UDF or option B (@I_vCMMTTEXT line comment).

B (@I_vCMMTTEXT line comment) and C (native POP10140 via internal zDP proc) considered, not chosen. C breaks eConnect-only rule.

Xilous:
Capture the TITAN manufacturer on every hardware item and carry it end to end into Nexus (parser, GraphQL, database). Use that manufacturer to suggest the ordering vendor when creating a GP purchase order, learning the manufacturer-to-vendor mapping each time a PO is confirmed and letting the user override. Write the manufacturer onto the GP PO line through the relay so it lands in Great Plains, using the taPoLine user-defined field (option A), gated on a pre-flight that the field is unused at the customer.

carved from #228.

issues in this effort, in dependency order:
- #231 manufacturer capture: TITAN manufacturer end to end (parser -> GraphQL -> DB)
- #232 manufacturer -> vendor suggestion + learn-on-confirm (needs #231)
- #233 write manufacturer onto GP PO line (taPoLine USRDEFND1) (needs #231)

each issue squash-merges into this pr as one commit; this pr is the single review surface for the whole effort.

Closes #228
Closes #231
Closes #232
Closes #233


Xilous:
issue #231 addressed by commit `5fd2085`.

Xilous:
issue #233 addressed by commit `f0a9eae`.

Xilous:
issue #232 addressed by commit `633bfec`.

backend/pyproject.toml
```diff
@@ -17,6 +17,7 @@ uvicorn = {version = "0.34.0", extras = ["standard"]}
 boto3 = ">=1.35.0"
 httpx = ">=0.27.0"
 pyjwt = {extras = ["crypto"], version = "^2.13.0"}
+rapidfuzz = ">=3.0.0"
```
Xilous, line 20: rapidfuzz was added here but poetry.lock was not regenerated: it has no rapidfuzz entry and its content-hash no longer matches pyproject.toml. CI runs `pip install poetry` (resolves to latest, 2.x) then `poetry install`, which hard-fails on an out-of-date lock ("pyproject.toml changed significantly since poetry.lock was last generated. Run `poetry lock` to fix the lock file."). Result: the whole backend job (lint, format, tests) fails, and the Railway deploy install breaks the same way. Run `poetry lock` in backend/ and commit the refreshed lock.

localhost relay/src/ucnexus_relay/models.py
```diff
@@ -21,6 +21,10 @@ class POLine(BaseModel):
     product_indicator: int = 1  # 1 = Non-Inventoried, 2 = Job Cost
     job_number: str | None = None
     cost_code: str | None = None  # 'phase-step-type' e.g. '210-200-2'
+    # captured manufacturer -> taPoLine USRDEFND1 (POP10110). char(50) in GP; the relay RTRIMs +
+    # truncates to 50 when binding, and this caps it at the boundary so an over-length value is
+    # rejected here rather than silently cut.
+    manufacturer: str | None = Field(default=None, max_length=50)
```
Xilous: max_length=50 makes Pydantic reject the whole create_po request when a manufacturer name exceeds 50 chars, before create_po_line ever runs. So an over-length manufacturer - a best-effort optional annotation - fails the entire GP PO push with a validation error, not just a truncated field. create_po_line already caps with (manufacturer or "").rstrip()[:50], which is the authoritative char(50) cap for the GP write. Drop the max_length constraint here so an over-length value is truncated (cap, not reject), and update the adjacent comment which currently states rejection is intended.

backend/app/schemas/queries.py
```diff
@@ -110,6 +112,20 @@
...
 
+def _po_line_item_manufacturer(li) -> str | None:
+    """Issue #232: derive a PO line's manufacturer from its linked HardwareItem rows (distinct, first
+    non-null if they disagree). Read only when the hardware_items relationship was eagerly loaded, so a
+    PO path that didn't selectinload it returns None rather than triggering an N+1 lazy load."""
+    from sqlalchemy import inspect as sa_inspect
+
+    if "hardware_items" in sa_inspect(li).unloaded:
+        return None
+    for hi in li.hardware_items:
```
Xilous, line 122: _po_line_item_manufacturer takes the first non-null across li.hardware_items in relationship order, and the hardware_items relationship (purchase_order.py) has no order_by, so the choice is DB-dependent. _resolve_line_manufacturers in mutations.py picks the first non-null ordered by created_at, id. When two HardwareItems for the same (hardware_category, product_code) carry different manufacturers, the vendor suggestion shown from this value can derive from a different manufacturer than the one written to GP USRDEFND1. Add order_by(created_at, id) to the hardware_items relationship so both paths resolve the same manufacturer.

backend/app/schemas/queries.py
```diff
@@ -1076,6 +1094,50 @@ async def gp_cost_codes(self, info: strawberry.Info, company: str, job: str) ->
...
+                    candidates=[
+                        VendorCandidate(
+                            gp_vendor_id=mapping.gp_vendor_id,
+                            gp_vendor_name=mapping.gp_vendor_name,
+                            score=100.0,
+                        )
+                    ],
+                )
+
+        result = await relay_gateway.relay_call(gp_company, "list_vendors")
```
Xilous, line 1120: When manufacturer normalizes to empty (blank input), the saved-mapping lookup returns None and rank_vendors returns [] regardless, yet this still issues a list_vendors relay round-trip only to build an empty result. Return an empty VendorSuggestion before the relay call when key is falsy, to skip the wasted round-trip.

frontend/src/modules/po/GpPurchaseOrderDialog.tsx
```diff
@@ -179,6 +189,60 @@ export default function GpPurchaseOrderDialog({
...
+  useEffect(() => {
+    if (!open) setMfrSuggestions({});
+  }, [open]);
+
+  useEffect(() => {
+    if (!open || !relayConnected || !company || distinctManufacturers.length === 0) return;
+    let cancelled = false;
+    (async () => {
+      const results: Record<string, ManufacturerSuggestion> = {};
+      for (const manufacturer of distinctManufacturers) {
```
Xilous: The per-manufacturer suggestion queries run sequentially - await is inside the for loop - so N distinct manufacturers means N serial relay round-trips before suggestionsLoaded flips. Run them concurrently with Promise.all(distinctManufacturers.map(...)) so they overlap.

backend/app/repositories/po_repository.py
```diff
@@ -13,9 +14,80 @@
...
+            product_codes = {code for _cat, code in pairs}
+            rows = session.execute(
+                select(
+                    HardwareItem.hardware_category,
+                    HardwareItem.product_code,
+                    HardwareItem.manufacturer,
+                ).where(
+                    HardwareItem.project_id == project_id,
+                    HardwareItem.product_code.in_(product_codes),
+                    HardwareItem.manufacturer.isnot(None),
```
Xilous, line 61: This select has no order_by, so labels_by_key.setdefault(key, label) keeps a DB-order-dependent label per normalized key. Two hardware rows with the same normalized manufacturer but different raw labels ("Acme Inc" vs "Acme, Inc.") can persist either one as manufacturer_label across runs. Add order_by(HardwareItem.created_at, HardwareItem.id) for a deterministic label.

frontend/src/modules/po/GpPurchaseOrderDialog.tsx
```diff
@@ -255,9 +329,31 @@ export default function GpPurchaseOrderDialog({
...
+  useEffect(() => {
+    if (!open || !suggestionsLoaded || userPickedVendorRef.current) return;
+    if (appliedMfrSignatureRef.current === manufacturerKey) return;
+    if (manufacturerHint.kind === 'single') {
+      const vendors = gpVendorsData?.gpVendors ?? [];
+      if (!vendors.some((v) => v.vendorId === manufacturerHint.top.gpVendorId)) return;
+      appliedMfrSignatureRef.current = manufacturerKey;
+      setGpVendorId(manufacturerHint.top.gpVendorId);
+      setGpVendorName(manufacturerHint.top.gpVendorName);
+      setVendorConfirmed(true);
```
Xilous, line 345: pre-select auto-confirms weak fuzzy match.
effect sets vendorConfirmed=true whenever manufacturerHint.kind === 'single'.
hint = top rank_vendors candidate, no score floor.
rank_vendors returns candidates whenever company has any vendors.
weak best match (30%) still pre-selected AND confirmed.
user can push PO to wrong GP vendor without explicit pick.

saved mapping (score 100) auto-confirm is correct. fuzzy match should either:
- pre-select but leave vendorConfirmed false
- gate auto-confirm on minimum score

backend/app/schemas/queries.py
```diff
@@ -1066,6 +1084,54 @@ async def gp_cost_codes(self, info: strawberry.Info, company: str, job: str) ->
...
```
Xilous, line 1120: one full PM00200 vendor-list fetch per manufacturer.
suggest_vendor_for_manufacturer -> relay_call(list_vendors) every invocation.
dialog fires it once per distinct manufacturer.
Apollo cache-first keys per (company, manufacturer); does not dedupe the GP fetch.
PO spanning K manufacturers pulls the full vendor list K times over Railway -> relay -> GP.

fetch the list once, rank every manufacturer against it:
- batch resolver taking a list of manufacturers, or
- rank client-side from a single list_vendors call.

backend/app/schemas/mutations.py
```diff
@@ -134,6 +138,49 @@ def _load_po_type(po_id: uuid.UUID) -> PurchaseOrder:
         return _po_to_type(po)
 
 
+def _resolve_line_manufacturers(session, project_id, line_items_data) -> list[str | None]:
```
Xilous, line 141: `"first non-null manufacturer, ordered by created_at then id"` reimplemented in 4 places:
_resolve_line_manufacturers (here)
po_repository._learn_manufacturer_vendor_mappings
queries._po_line_item_manufacturer
POLineItem.hardware_items relationship order_by

they agree today only because each repeats the ordering + first-non-null independently.
change to the rule in one place diverges the manufacturer written to GP from displayed and learned. extract single shared resolver.

backend/app/repositories/po_repository.py
```diff
@@ -297,7 +395,9 @@ def get_purchase_orders(
     stmt = (
         select(PurchaseOrder)
         .options(
-            selectinload(PurchaseOrder.line_items),
+            # Nested load of each line's HardwareItem rows so the derived `manufacturer` field (issue
+            # #232) is one selectin query, not an N+1 lazy load per line.
+            selectinload(PurchaseOrder.line_items).selectinload(POLineItem.hardware_items),
```
Xilous, line 400: get_purchase_orders (list) -> selectinload line_items.hardware_items -> _po_line_item_manufacturer picks first manufacturer.
loads every HardwareItem row per line for one derived string.
per this repo's list-resolver rules, a grouped scalar query is lighter than materializing the child collection per line.
use grouped select of POLineItem id + first manufacturer instead. same for get_purchase_order.

PR #303, opened 2026-07-21
relay-link heartbeat: reap a dead relay within ~a minute

### LINKED ISSUE #277: relay-link heartbeat: dead relay reads as connected for an hour
on 2026-07-16 the relay process on the tagging workstation died abruptly (no clean WS close) somewhere in a ~1h idle window. /relay-link has no heartbeat - connected is just "socket registered" and unregister only runs when a read fails - so relayStatus kept reporting connected until the platform reaped the dead TCP connection over an hour later. nothing was calling GP in that window, so the UI chip said "relay connected" the whole time the relay was gone.

- add ping/pong on /relay-link with a short timeout (ping every 30s, unregister after 2 misses) so a dead relay flips relayStatus within a minute
- a relay_call against a half-dead socket currently burns the full 30s DEFAULT_TIMEOUT before failing; the heartbeat reap lets it fail fast with the existing RelayUnavailableError
- the relay already auto-reconnects (it survived ~8 backend redeploys today), so no relay-side change is strictly needed; websocket clients answer pings automatically

backend/main.py relay_link route + app/services/relay_gateway.py

Xilous:
closes #277

/relay-link had no heartbeat: `connected` meant only "socket registered", and unregister ran only when a read failed. a relay that died without a clean close read as connected until the platform reaped the dead TCP socket ~an hour later - nothing was calling GP in that window, so the chip stayed green the whole time the relay was gone.

ASGI has no ping/pong control frame, so the route can't emit WS protocol pings (those live below the app in uvicorn's `--ws-ping-*`, which wasn't reaping here). this rides an application-level `{"type":"ping"}` / `{"type":"pong"}` data exchange instead, which needs the small relay-side answer.

backend
- route runs the read loop and a heartbeat as two concurrent tasks; first to finish cancels the other, `finally` still unregisters
- heartbeat pings every HEARTBEAT_INTERVAL_SECONDS (20s), reaps after 2 unanswered pongs -> close -> unregister; relayStatus flips in ~40s, under a minute with the frontend's 10s poll
- reap arms only after the first pong on a connection, so a relay that doesn't answer pings yet is never falsely killed (no deploy-order thrash)
- reap runs the existing `_fail_all`, so an in-flight relay_call on a dead socket fails fast with RelayUnavailableError instead of hanging the full 30s

relay
- channel answers `{"type":"ping"}` with `{"type":"pong"}` through its existing writer queue, not dispatched as a job
- the client's own protocol ping (corporate-proxy keepalive) is unchanged

no migration, no frontend change. deploy relay first (answering pings is harmless to the current backend), then backend.

### DIFF
--- a/backend/app/services/relay_gateway.py
+++ b/backend/app/services/relay_gateway.py
@@ -7,8 +7,10 @@ This is deliberate: the old supersede-the-incumbent behaviour (issue #202 #6) bo
 relays thrash - each new connection force-closing the other, reconnecting, and superseding back - and
 could drop a valid in-flight reply, failing a relay_call whose GP write had actually committed. Failing
 pending calls now happens only on a genuine disconnect (unregister), where the reply truly can't arrive.
-A briefly half-dead incumbent is detected by the websockets ping timeout, which fires unregister and
-frees the slot for the next reconnect."""
+A briefly half-dead incumbent is detected by the app-level heartbeat (issue #277): the /relay-link route
+pings the relay on a data message every HEARTBEAT_INTERVAL_SECONDS and, after HEARTBEAT_MAX_MISSED
+unanswered pings, closes the socket - which fires unregister, frees the slot, and flips relayStatus
+within ~a minute instead of the ~hour it took the platform to reap the dead TCP connection."""
 
 import asyncio
 import uuid
@@ -19,12 +21,25 @@ from app.errors import RelayCallError, RelayTimeoutError, RelayUnavailableError
 
 DEFAULT_TIMEOUT_SECONDS = 30.0
 
+# App-level heartbeat (issue #277). ASGI exposes no WS ping/pong control frame, so the route rides a
+# {"type": "ping"} / {"type": "pong"} data exchange rather than protocol pings. The relay's websockets
+# client answers protocol pings automatically but NOT these data pings, so it carries a matching handler
+# (relay channel.py). Detection is up to HEARTBEAT_INTERVAL_SECONDS * HEARTBEAT_MAX_MISSED (~40s here).
+HEARTBEAT_INTERVAL_SECONDS = 20.0
+HEARTBEAT_MAX_MISSED = 2
+
 
 class RelayGateway:
     def __init__(self) -> None:
         self._socket: WebSocket | None = None
         self._company: str | None = None
         self._pending: dict[str, asyncio.Future] = {}
+        # Heartbeat bookkeeping for the current connection (issue #277). `_unanswered_pings` counts pings
+        # sent since the last pong; `_heartbeat_armed` only turns True after the relay answers its first
+        # ping, so a relay build that doesn't speak the data heartbeat yet is never falsely reaped - it
+        # just falls back to the old disconnect-on-read behaviour.
+        self._unanswered_pings = 0
+        self._heartbeat_armed = False
 
     @property
     def connected(self) -> bool:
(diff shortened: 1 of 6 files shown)


backend/main.py
```diff
@@ -82,6 +84,65 @@ def health():
...
 
+async def _relay_read_loop(websocket: WebSocket) -> None:
+    """Feed each frame the relay sends to relay_gateway: a {"type": "pong"} answers the heartbeat
+    (issue #277), anything else is a {id, ok, result|error} job reply to correlate with relay_call()."""
+    while True:
+        message = await websocket.receive_json()
+        if isinstance(message, dict) and message.get("type") == "pong":
+            relay_gateway.note_pong()
+        else:
+            relay_gateway.resolve(message)
```
Xilous, line 95: non-dict json frame from the relay -> `resolve(message)` -> `reply.get("id")` raises AttributeError (a list/str/number has no `.get`), crashing the read loop. the route`s except only catches WebSocketDisconnect + CancelledError, so it propagates uncaught (finally still unregisters, but it logs an error). latent today since the relay only sends dicts, but this is a trust boundary - worth an `isinstance(message, dict)` guard before resolve. low.

backend/main.py
```diff
@@ -116,11 +177,14 @@ async def relay_link(websocket: WebSocket):
         await websocket.close(code=4409)
         return
     try:
-        while True:
-            reply = await websocket.receive_json()
-            relay_gateway.resolve(reply)
+        await _serve_relay_link(websocket)
     except WebSocketDisconnect:
         pass
+    except asyncio.CancelledError:
```
Xilous, line 183: `except asyncio.CancelledError: pass` swallows the cancel instead of re-raising after cleanup - on graceful shutdown the route ends "normally" rather than propagating. reads intentional (documented, paired with the client-side-close test change), so likely fine; flagging only because swallowing CancelledError is usually a smell. low, probably wontfix.

backend/app/services/relay_gateway.py
```diff
@@ -44,15 +59,34 @@ def try_register(self, company: str, websocket: WebSocket) -> bool:
...
+        if self._socket is not None:
+            self._unanswered_pings = 0
+            self._heartbeat_armed = True
+
+    def register_ping_miss(self) -> bool:
+        """Called by the route's heartbeat once per interval, just before it sends the next ping. Counts
+        the pings the relay hasn't answered yet and returns True when it has gone quiet long enough to
+        reap (HEARTBEAT_MAX_MISSED unanswered pings) - but only once armed, so a relay that never answers
+        a ping is left to the disconnect-on-read path rather than being killed on a false positive."""
+        self._unanswered_pings += 1
```
Xilous, line 87: minor asymmetry: note_pong guards `if self._socket is not None` before touching the counters, but register_ping_miss increments unconditionally. safe today (the heartbeat task is cancelled before unregister, so it never runs post-unregister), but a later decoupling of the heartbeat lifecycle from registration would let it bump a phantom counter. low, defensive.

relay/src/ucnexus_relay/channel.py
```diff
@@ -245,6 +259,12 @@ async def _dispatch_job(job: dict) -> None:
                 except (json.JSONDecodeError, TypeError):
                     logger.warning("channel received a non-JSON message", extra={"raw": str(raw)[:200]})
                     continue
+                pong = _heartbeat_reply(job)
+                if pong is not None:
+                    # Backend heartbeat (issue #277): answer through the same writer queue so the pong
+                    # never interleaves mid-frame with a job reply, and don't dispatch it as a job.
+                    await send_queue.put(pong)
```
Xilous, line 266: the pong wiring here (put on send_queue + continue) is untested - test_channel_heartbeat only exercises _heartbeat_reply in isolation. a regression that dropped the continue (dispatching the ping as a job) or used the wrong queue would not be caught. low; consider an integration test over the _run_once dispatch loop.

PR #306, opened 2026-07-21
fix(relay): stop the self-update loop, migrate onefile -> onedir so relaunch survives Defender

Xilous:
relay update looped a popping console window (`find /i "ucnexus-relay"`), returned on every close. two layered causes.

1. self-update helper was a detached `update-helper.bat`:
- both success + failure branches ran `start "" exe app`, so closing the relay did not end the sequence
- `taskkill /f /im ucnexus-relay.exe` killed every relay process, not the target pid
- no circuit breaker, timeout, or graceful abort
- `DETACHED_PROCESS | CREATE_NO_WINDOW` is contradictory -> cmd children each opened a window

replaced with an update-apply subcommand:
- windowless, logged to `update.log`
- bounded by ~90s deadline + per-target attempt ledger (`update-state.json`, max 3 attempts)
- kills by pid (reuses `single_instance.force_kill_pid`)
- relaunches once

Updates tab surfaces the last outcome.

2. relay was PyInstaller ONEFILE, re-extracts the C-extension `.pyd` to `%TEMP%` every launch. on a Windows Defender box a freshly-written exe's extraction is scanned as it loads -> the relaunched relay crashes (`_multiprocessing` / `PIL._imaging`). immediate post-update relaunch = the loop.

migrated onefile -> ONEDIR. `.pyd` permanent, scanned once at install/update. a running relay holds them open, so its folder cannot be renamed in place; each version lives in `app-<build>/`, and a stable `current` directory junction (shortcuts + autostart target it) points at the active version.

update flow: extract the new version alongside -> a helper from the OLD settled version force-stops by pid -> repoints the junction -> health-gated relaunch -> rolls the junction back to the previous version if the new one does not come up.

also polished:
- log rotation
- single-sourced VERSION
- quieted the de-enrolled reconnect flood
- bounded UI log tailing
- documented the TrustServerCertificate trade-off

validated on a Defender box: installed the CI onedir build, ran a real update, relaunched HEALTHY (serve + channel, no C-ext crash), junction repointed, clean update.log, success ledger, single relaunch, zero console windows.

onefile -> onedir cannot self-update (the old updater downloads a zip it cannot apply), so the first onedir build is installed manually. onedir -> onedir updates work via the junction. release ships `ucnexus-relay.zip`.

### DIFF
--- a/relay/src/ucnexus_relay/single_instance.py
+++ b/relay/src/ucnexus_relay/single_instance.py
@@ -395,13 +395,15 @@ def installed_build(config_dir, timeout: float = 8.0) -> str | None:
     return (out.stdout or "").strip() or None
 
 
-def launch_installed(config_dir, minimized: bool = False) -> None:
+def launch_installed(config_dir, minimized: bool = False) -> int | None:
     """Spawn the installed exe (via the `current` junction) as the desktop app, detached so it outlives
     this process. The relaunched exe runs the single-instance gate itself and becomes the owner.
-    Autostart + shortcuts target the stable `current` path, so a version change needs no refresh here."""
+    Autostart + shortcuts target the stable `current` path, so a version change needs no refresh here.
+    Returns the launched app's pid so a caller (the updater's health-gated retry) can force it down BY PID
+    if it comes up unhealthy - onedir, so the exe process IS the app process (no bootloader child)."""
     exe = installed_exe_path(config_dir)
     args = [str(exe), "app"] + (["--minimized"] if minimized else [])
-    subprocess.Popen(  # noqa: S603 (our own installed exe path)
+    proc = subprocess.Popen(  # noqa: S603 (our own installed exe path)
         args,
         cwd=str(config_dir),
         creationflags=_DETACHED,  # DETACHED alone (GUI exe: no console either way; not paired with NO_WINDOW)
@@ -409,3 +411,4 @@ def launch_installed(config_dir, minimized: bool = False) -> None:
         stdout=subprocess.DEVNULL,
         stderr=subprocess.DEVNULL,
     )
+    return proc.pid
--- a/relay/src/ucnexus_relay/single_instance.py
+++ b/relay/src/ucnexus_relay/single_instance.py
@@ -10,9 +10,9 @@ same serve. This module makes `app` single-instance and version-aware:
     build), or defers to it (older / dev), per decide_action();
   - a session-local named mutex serialises two simultaneous cold starts so exactly one becomes owner.
 
-Promote-to-installed lives here too: a newer exe launched from anywhere (Downloads, a USB stick) swaps
-itself over the installed exe and relaunches from there, so the single-file distributable doubles as its
-own upgrader (see promote_decision / promote_swap).
+The installed exe is resolved through the onedir `current` junction (see layout.installed_exe); updates
+swap versions by repointing that junction (see updater), not by copying an exe, so there is no
+promote-from-Downloads path here anymore.
 
 Everything is import-safe without a GUI backend - app.py wires the show / quit callbacks to the window -
 so the decision logic and the lockfile plumbing unit-test without opening a window.
@@ -21,7 +21,6 @@ so the decision logic and the lockfile plumbing unit-test without opening a wind
 import json
 import os
 import secrets
-import shutil
 import socket
 import subprocess
 import sys
@@ -29,7 +28,7 @@ import threading
 import time
 from pathlib import Path
 
-from . import updater
+from . import layout, updater
 from .logging_setup import get_logger
 
 logger = get_logger()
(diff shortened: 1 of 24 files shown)


relay/src/ucnexus_relay/updater.py
```diff
@@ -89,138 +118,368 @@ def _unlink_quietly(p: Path) -> None:
...
+    url = _health_url(install_dir)
+    for i in range(1, attempts + 1):
+        if _monotonic() >= deadline:
+            break
+        single_instance.launch_installed(install_dir)
+        log.info("relaunch attempt %d of %d", i, attempts)
+        if _wait_for_health(url, min(deadline, _monotonic() + _HEALTH_WAIT_PER_ATTEMPT)):
+            log.info("relay healthy after relaunch")
+            return True
+        _kill_relay_pids(None, install_dir, log)  # clear a crashed/half attempt before retrying
```
Xilous: Retry and rollback here can leave the relay down. Between attempts this calls `_kill_relay_pids(None, ...)`, which force-stops only the serve child (from relay.pid), not the `app` process `launch_installed` just spawned. That app is single-instance and long-lived, so if attempt 1 comes up alive-but-unhealthy - or merely times out the 25s health wait on a slow Defender box, the exact case this PR targets - then:
- attempt 2's `launch_installed` finds the still-alive same-build owner and focus-deflects (returns without starting serve)
- the rollback relaunch (older build) also focus-deflects against that owner
`ensure_serve` runs once with no supervisor, so serve is never restarted. Result: relay down, contradicting the "never left down" guarantee, precisely when the health gate is supposed to save it.

Fixed in this review: `launch_installed` now returns the app pid, and the between-attempts kill force-stops that pid + its serve, so the next attempt and the rollback launch are no longer deflected.

relay/src/ucnexus_relay/updater.py
```diff
@@ -89,138 +118,368 @@ def _unlink_quietly(p: Path) -> None:
...
+    _wait_for_pid_exit(app_pid, deadline, log)
+    _kill_relay_pids(app_pid, install_dir, log)
+
+    if _cancel_requested(install_dir):
+        log.info("cancel requested during teardown; aborting without relaunch")
+        _finish_ledger(install_dir, "cancelled", None)
+        _clear_cancel(install_dir)
+        return {"ok": False, "cancelled": True}
+
+    if not layout.repoint_current(install_dir, new_dir):
```
Xilous, line 449: A failed repoint is swallowed (warning only) and the code proceeds to relaunch as if it succeeded. `repoint_current` does `remove_junction` then `mklink /J`; if mklink fails, `current` is left removed. The relaunch then comes up via the newest-version fallback in `installed_exe`, health passes, and the ledger records "success" - but `current\ucnexus-relay.exe` no longer exists, so the next logon's HKCU Run autostart and the Desktop/Start-menu shortcuts (all targeting `current`) are broken, and the success-path cleanup also deletes the old version.

Fixed in this review: on a failed repoint, roll the junction back to the previous version, relaunch it, and record "failed" instead of continuing.

relay/src/ucnexus_relay/updater.py
```diff
@@ -89,138 +118,368 @@ def _unlink_quietly(p: Path) -> None:
...
+    log = _update_logger(install_dir)
+
+    ledger = read_ledger(install_dir)
+    attempts = ledger.get("attempts", 0) + 1
+    target = ledger.get("target_build") or "?"
+    target_dir = ledger.get("target_dir")
+    ledger["status"] = "applying"
+    ledger["attempts"] = attempts
+    ledger["updated_at"] = _now_iso()
+    ledger.setdefault("first_attempt_at", _now_iso())
```
Xilous: `first_attempt_at` is never stamped for a fresh target. `_stage_ledger` seeds the key as `None` for a new target, so `setdefault` sees the key already present and is a no-op - the field stays `None` for the life of that target.

Fixed in this review: stamp it explicitly when falsy.

relay/src/ucnexus_relay/updater.py
```diff
@@ -89,138 +118,368 @@ def _unlink_quietly(p: Path) -> None:
...
+    install_dir = Path(install_dir)
+    ver, error = _download_and_extract(url, install_dir, None)
+    if ver is None:
+        return {"ok": False, "error": error}
+
+    old_target = _current_target(install_dir)
+    setup.stop_serve(install_dir)
+    ok = layout.repoint_current(install_dir, ver)
+    try:
+        setup.start_serve(single_instance.installed_exe_path(install_dir), install_dir)
```
Xilous, line 491: Minor state mismatch on the standalone path: serve is started from `installed_exe_path` before the repoint result (`ok`) is checked. If the repoint failed, `current` is broken/absent so serve starts from the new-version fallback, then the junction is rolled back to old - leaving serve running the NEW exe while `current` points at OLD, with no restart to reconcile.

Low impact: this path only runs when `app.current()` is None, which the real desktop UI never hits (the Api is only live inside the app window). Not changed.

relay/src/ucnexus_relay/single_instance.py
```diff
@@ -381,7 +368,9 @@ def bring_to_front(title: str = WINDOW_TITLE) -> bool:
 
 
 def installed_exe_path(config_dir) -> Path:
-    return Path(config_dir) / ASSET_NAME
+    """The exe to run/inspect for this install - resolved through the onedir `current` junction to the
+    active app-<build>/ version (see layout.installed_exe)."""
+    return layout.installed_exe(config_dir)
```
Xilous, line 373: Dead code after removing promote: `installed_build`, `same_path`, and the `print-build` CLI subcommand are now referenced only by their own tests - the `_maybe_promote` path that consumed them was deleted in this PR. Candidates for removal (with their tests).

Not changed: `print-build` is a documented CLI subcommand and may be worth keeping as a diagnostic, so removal is a judgment call for the author.

PR #319, opened 2026-07-23
fix(relay): bundle idna codec's stringprep dep so self-update health probe can't crash the helper (#318)

### LINKED ISSUE #318: IT admin tried updating the relay and ran into an error
Traceback (most recent call last):
  File "relay_entry.py", line 13, in <module>
  File "ucnexus_relay\cli.py", line 232, in main
  File "ucnexus_relay\updater.py", line 447, in apply_staged_update
  File "ucnexus_relay\updater.py", line 325, in _relaunch_and_wait_healthy
  File "ucnexus_relay\updater.py", line 304, in _wait_for_health
  File "urllib\request.py", line 216, in urlopen
  File "urllib\request.py", line 519, in open
  File "urllib\request.py", line 536, in _open
  File "urllib\request.py", line 496, in _call_chain
  File "urllib\request.py", line 1377, in http_open
  File "urllib\request.py", line 1348, in do_open
  File "http\client.py", line 1303, in request
  File "http\client.py", line 1349, in _send_request
  File "http\client.py", line 1298, in endheaders
  File "http\client.py", line 1058, in _send_output
  File "http\client.py", line 996, in send
  File "http\client.py", line 962, in connect
  File "socket.py", line 827, in create_connection
  File "socket.py", line 962, in getaddrinfo
LookupError: unknown encoding: idna

Xilous:
Closes #318.

IT admin updated relay -> unhandled traceback `LookupError: unknown encoding: idna`.

self-update spawns detached helper (`ucnexus-relay update-apply`) -> repoints `current` junction to new build -> health-gates relaunch polling `http://127.0.0.1:7321/health`. poll goes urllib -> http.client -> `socket.getaddrinfo` -> loads `idna` codec to encode host.

onedir bundle carries `encodings/idna.pyc`, not idna.py's `import stringprep`.
PyInstaller default `encodings` collection adds idna.pyc without analyzing imports -> `stringprep` never bundled.
runtime `import encodings.idna` dies on missing `stringprep` -> codec registry swallows ImportError -> caller gets `LookupError: unknown encoding: idna`.

pre-fix bundle -> idna.pyc + unicodedata present, stringprep MISSING -> codecs.lookup('idna') fails (#318).
post-fix bundle -> all three present -> codecs.lookup('idna') succeeds.

- add `encodings.idna` + `stringprep` to spec `hiddenimports`. stringprep is the load-bearing line; naming `encodings.idna` makes PyInstaller analyze it and follow its stringprep + unicodedata deps.
- broaden `_wait_for_health` from `(OSError, json.JSONDecodeError)` to swallow any probe error, retry within deadline.
- regression test added.

pre-fix installs run the old crashing probe on the transition update (helper runs from OLD build exe); junction is already repointed + new build already launched before the crash, so relay still comes up on the fixed build. one traceback on that one update; later updates clean.

### DIFF
--- a/relay/src/ucnexus_relay/updater.py
+++ b/relay/src/ucnexus_relay/updater.py
@@ -39,13 +39,13 @@ _MIN_EXE_BYTES = 5_000_000  # the extracted exe
 
 # --- self-update helper bounds ------------------------------------------------------------------------
 _STATE_FILE = "update-state.json"  # the attempt ledger the helper reads/writes and the UI surfaces
-_CANCEL_FLAG = "update-cancel"     # touched by a user-initiated shutdown to abort an in-flight update
-_UPDATE_LOG = "update.log"         # the helper's own log, next to config.toml
+_CANCEL_FLAG = "update-cancel"  # touched by a user-initiated shutdown to abort an in-flight update
+_UPDATE_LOG = "update.log"  # the helper's own log, next to config.toml
 _DOWNLOAD_NAME = "ucnexus-relay-download.zip"
-MAX_ATTEMPTS = 3                   # circuit breaker: give up on a target build after this many helper runs
-_GLOBAL_DEADLINE_SECONDS = 90.0    # hard wall-clock cap on the whole sequence
-_APP_EXIT_WAIT_SECONDS = 20.0      # how long to wait for the app to exit on its own before force-killing
-_HEALTH_WAIT_PER_ATTEMPT = 25.0    # per relaunch attempt, how long to wait for /health before retrying
+MAX_ATTEMPTS = 3  # circuit breaker: give up on a target build after this many helper runs
+_GLOBAL_DEADLINE_SECONDS = 90.0  # hard wall-clock cap on the whole sequence
+_APP_EXIT_WAIT_SECONDS = 20.0  # how long to wait for the app to exit on its own before force-killing
+_HEALTH_WAIT_PER_ATTEMPT = 25.0  # per relaunch attempt, how long to wait for /health before retrying
 _POLL_SECONDS = 0.5
 
 _DETACHED_PROCESS = 0x00000008  # DETACHED alone (never paired with CREATE_NO_WINDOW - see stage/_spawn)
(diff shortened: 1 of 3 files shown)


relay/src/ucnexus_relay/updater.py
```diff
@@ -304,7 +304,10 @@ def _wait_for_health(url: str, deadline: float) -> bool:
             with urllib.request.urlopen(url, timeout=2) as r:  # noqa: S310 (localhost health URL)
                 if json.loads(r.read().decode()).get("status") == "ok":
                     return True
-        except (OSError, json.JSONDecodeError):
+        except Exception:  # noqa: BLE001 - the probe's ONLY job is "is it up yet?"; ANY failure (connection
```
Xilous: Broadening to a bare `except Exception: pass` with no logging removes all diagnostic signal from a genuinely broken build. The original #318 failure was at least a loud traceback naming `LookupError: unknown encoding: idna`. After this change any probe failure (missing codec, TLS error, wrong host, a real logic bug in the probe) is swallowed silently, the loop just burns the whole deadline, and the only downstream log is the generic `did not become healthy; rolling back`. The next #318-class regression will be much harder to root-cause than the crash it replaced. Consider threading the caller's `log` into `_wait_for_health` and logging the swallowed exception (debug per-probe, or once on the final timeout) so the cause survives.

relay/tests/test_updater.py
```diff
@@ -253,6 +269,31 @@ def test_kill_relay_pids_uses_pid_not_image_name(tmp_path, monkeypatch):
     assert killed == [4242, 9001]  # app pid + serve pid, both by pid (never taskkill /im)
 
 
+# --- _wait_for_health ---------------------------------------------------------------------------------
+
+
+def test_wait_for_health_returns_true_on_ok(monkeypatch):
+    monkeypatch.setattr(updater.urllib.request, "urlopen", lambda *a, **k: _Resp({"status": "ok"}))
+    monkeypatch.setattr(updater, "_monotonic", lambda: 0.0)  # always before the deadline
```
Xilous: This happy-path test pins `_monotonic` to a constant `0.0` (always before the 5.0 deadline) and does not patch `_sleep`. If `_wait_for_health` ever regresses to not return True, `while 0.0 < 5.0` never exits and each iteration calls the real `time.sleep(2.0)` - the test hangs forever (CI timeout) instead of failing cleanly. Drive `_monotonic` off a tick list that steps past the deadline, the way `test_wait_for_health_swallows_a_probe_error_instead_of_crashing` does, so a regression terminates as a clean failure.

relay/src/ucnexus_relay/updater.py
Xilous: done in ed9f206.
`_wait_for_health` takes an optional logger. on a never-healthy timeout it logs the last swallowed error `health probe <url> never returned ok before the deadline; last error: <repr>`.
`_relaunch_and_wait_healthy` threads its `log` in.
broad catch retained, but no longer erases signal: a never-up rollback now names the last failure.
added `test_wait_for_health_logs_last_error_on_timeout`.

relay/tests/test_updater.py
Xilous: done in ed9f206.
happy-path test drives `_monotonic` off a tick list stepping past the deadline, no-ops `_sleep`.
regression that stops returning True now times out to False instead of hanging on the real clock.

PR #322, opened 2026-07-23
fix(import): drop BY_OTHERS from shop-assembly finalize classifications (#321)

### LINKED ISSUE #321: Shop-assembly re-import finalize sends BY_OTHERS in the SITE_HARDWARE/SHOP_HARDWARE classification field, blocks SAR creation
finalizeImportSession rejects the input:
Variable '$input' got invalid value 'BY_OTHERS' at 'input.classifications[0].classification'; Value 'BY_OTHERS' does not exist in 'Classification' enum.

Classification enum = [SITE_HARDWARE, SHOP_HARDWARE] only
ClassificationInput.classification = NON_NULL Classification (SITE_HARDWARE or SHOP_HARDWARE)
frontend sends 'BY_OTHERS' (By UCSH / By Others ownership value) into that field; classifications[0] triggered it

repro (project 80003, opening 0019-EX):
- Import for Create Purchase Orders. classify 2 SARGENT items By UCSH + Shop, other 20 By Others.
- Register PO in GP, receive it. 2 SARGENT items -> In Inventory.
- Re-import same opening, purpose Pull Request for Shop Assembly.
- Classification step shows only the 2 in-inventory items; classify both Shop.
- Finalize -> fails with error above.

root cause: shop-assembly finalize `classifications` array includes By-Others items carrying 'BY_OTHERS'.

fix - one of:
- map/convert them
- exclude ownership-By-Others items from classifications array (they are not shop-assembled)

impact: blocks SAR creation for any schedule with any By-Others items (essentially all real hardware schedules). shop assembly can't start through the import wizard.

Xilous:
Fixes #321.

Shop-assembly re-import Finalize rejected at GraphQL: "Value 'BY_OTHERS' does not exist in 'Classification' enum". hits any schedule with By-Others items (i.e. essentially every real schedule). blocks SAR creation.

import wizard keeps one shared classifications Map keyed hardware_category|product_code|unit_cost.
on re-import a useEffect pre-populates Map with BY_OTHERS for every parsed item matching project exclusion table.
effect gated on isReimport only, not import purpose.
assembly Classification step shows only RECEIVED items, sets SITE_HARDWARE/SHOP_HARDWARE.
BY_OTHERS entries stay in Map invisibly.
finalize assembly branch maps every Map entry -> input.classifications -> BY_OTHERS into Site/Shop-only Classification field -> Strawberry rejects mutation before resolver.

fix must be frontend-side: GraphQL enum rejects BY_OTHERS before the resolver runs, backend can't guard it.

exclude ownership-By-Others from classifications array (issue option 2).

- add toClassificationInputs() in frontend/src/modules/import/types.ts. projects classifications Map -> finalize ClassificationInput entries, keeps only SITE_HARDWARE/SHOP_HARDWARE. allow-list also drops empty/unexpected values (more robust than != BY_OTHERS deny-list).
- use in assembly branch of buildFinalizeInput in ImportWizard.tsx.
- PO branch unchanged (already excludes By-Others via byOthersKeys; its values are always Site/Shop).
- backend unchanged: By-Others AVAILABLE rows already persist with classification = None, same as the existing PO path.

classificationInputs.test.ts covers:
- BY_OTHERS dropped, Site/Shop kept
- decimal unit-cost parsing
- empty/unexpected dropped
- empty Map

### DIFF
--- a/frontend/src/modules/import/ImportWizard.tsx
+++ b/frontend/src/modules/import/ImportWizard.tsx
@@ -36,7 +36,7 @@ import { GET_PROJECT_EXCLUDED_ITEMS, GET_PROJECT_HARDWARE_SCHEDULE, RECONCILE_SC
 import { GET_PROJECTS } from '../../graphql/shared';
 import type { ClassificationRow } from './ClassificationGrid';
 import type { AggregatedHardwareItem, ImportPurpose, ReconciliationRow, ShippingPRDraft } from './types';
-import { aggregationKey, classificationKey } from './types';
+import { aggregationKey, classificationKey, toClassificationInputs } from './types';
 import type { Project } from '../../types/project';
 import type { ProjectHardwareScheduleResponse } from './hydrateSchedule';
 import { mapScheduleResponseToParseResult } from './hydrateSchedule';
@@ -694,10 +694,10 @@ export default function ImportWizard({ open, project, onClose }: ImportWizardPro
         : null,
       excludedItems,
       classifications: purpose === 'assembly'
-        ? Array.from(classifications.entries()).map(([key, cls]) => {
-            const [hardwareCategory, productCode, unitCost] = key.split('|');
-            return { hardwareCategory, productCode, unitCost: parseFloat(unitCost), classification: cls };
-          })
+        // #321: only Site/Shop belong here. Re-imports pre-populate the classifications Map with
+        // BY_OTHERS (ownership) from the exclusion table; those items are out of scope for shop
+        // assembly and BY_OTHERS is not in the Classification enum, so toClassificationInputs drops them.
+        ? toClassificationInputs(classifications)
         : purpose === 'po'
           // Issue #216: the PM's Site/Shop picks from the Classification step's second axis.
           // By-Others items are out of scope and carry none.
--- a/frontend/src/modules/import/types.ts
+++ b/frontend/src/modules/import/types.ts
@@ -59,3 +59,24 @@ export const ASSEMBLY_OPTIONS: ClassificationOption[] = [
   { value: 'SITE_HARDWARE', label: 'Site', color: 'success' },
   { value: 'SHOP_HARDWARE', label: 'Shop', color: 'info' },
 ];
+
+export interface ClassificationInputEntry {
+  hardwareCategory: string;
+  productCode: string;
+  unitCost: number;
+  classification: string;
+}
+
+// #321: project the shared classifications Map into finalize ClassificationInput entries, keeping
+// only real Site/Shop values. Re-imports seed the Map with BY_OTHERS (an ownership value, not a
+// Site/Shop one) from the exclusion table; those items are out of scope for shop assembly and the
+// Classification GraphQL enum only accepts SITE_HARDWARE/SHOP_HARDWARE, so they are dropped here.
+// Allow-listing the two valid values also drops any empty/unexpected value.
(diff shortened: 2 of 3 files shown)


frontend/src/modules/import/types.ts
```diff
@@ -59,3 +59,24 @@ export const ASSEMBLY_OPTIONS: ClassificationOption[] = [
...
+// #321: project the shared classifications Map into finalize ClassificationInput entries, keeping
+// only real Site/Shop values. Re-imports seed the Map with BY_OTHERS (an ownership value, not a
+// Site/Shop one) from the exclusion table; those items are out of scope for shop assembly and the
+// Classification GraphQL enum only accepts SITE_HARDWARE/SHOP_HARDWARE, so they are dropped here.
+// Allow-listing the two valid values also drops any empty/unexpected value.
+export function toClassificationInputs(classifications: Map<string, string>): ClassificationInputEntry[] {
+  return Array.from(classifications.entries())
+    .filter(([, cls]) => cls === 'SITE_HARDWARE' || cls === 'SHOP_HARDWARE')
+    .map(([key, cls]) => {
+      const [hardwareCategory, productCode, unitCost] = key.split('|');
```
Xilous, line 79: toClassificationInputs destructures key.split('|') into exactly three vars, but classificationKey builds the key as hardware_category|product_code|unit_cost. If any hardware_category or product_code contains a '|', the split yields >3 segments: productCode picks up the wrong segment and unitCost becomes a non-numeric string, so parseFloat returns NaN. The backend keys classification_map on (hardware_category, product_code, unit_cost) and matches with hi_data unit_cost, so a mismatched/NaN key never matches and the item silently persists with classification=None instead of the PM's Site/Shop pick. This is a pre-existing keying assumption (classificationKey uses the same encoding) but it is re-implemented here. Low likelihood for hardware part numbers, though a guarded split (split with a limit and rejoin the tail, or a non-textual key) would make it safe.

frontend/src/modules/import/ImportWizard.tsx
```diff
@@ -694,10 +694,10 @@ export default function ImportWizard({ open, project, onClose }: ImportWizardPro
...
       excludedItems,
       classifications: purpose === 'assembly'
-        ? Array.from(classifications.entries()).map(([key, cls]) => {
-            const [hardwareCategory, productCode, unitCost] = key.split('|');
-            return { hardwareCategory, productCode, unitCost: parseFloat(unitCost), classification: cls };
-          })
+        // #321: only Site/Shop belong here. Re-imports pre-populate the classifications Map with
+        // BY_OTHERS (ownership) from the exclusion table; those items are out of scope for shop
+        // assembly and BY_OTHERS is not in the Classification enum, so toClassificationInputs drops them.
+        ? toClassificationInputs(classifications)
```
Xilous, line 700: The two sibling finalize branches now filter the same classifications concept two different ways: assembly uses an allow-list (cls === SITE_HARDWARE || cls === SHOP_HARDWARE) inside toClassificationInputs, while the PO branch just below uses a deny-list (cls !== '' && !byOthersKeys.has(key)). They also duplicate the identical key.split('|') + parseFloat projection (PO branch still inlines it; toClassificationInputs re-implements it). Two costs: any future third Classification value would be silently dropped by assembly but kept by PO, and the projection is maintained in two places. Consider having the PO branch pass its own predicate into a shared projector so both paths share one code path.

PR #331, opened 2026-07-23
feat(shop-assembly): manager UI to assign pulled openings to team members (#330)

### LINKED ISSUE #330: No UI workflow for a shop-assembly manager to assign accepted requests to team members: only 'Assign to me' exists; Assignments page is an empty stub
no UI for shop-assembly manager to assign accepted requests to team members. only "Assign to me" exists.

manager accepts shop-assembly requests -> assigns resulting openings to team members (other shop-assembly users). almost always someone else, not self.

full-lifecycle e2e, project 80003, opening 0019-EX, after SAR pull Completed + leaf openings PULLED:
- Shop Assembly -> Assignments tab (/app/shop-assembly/assignments) renders empty. main innerHTML length 0, no elements, no console errors, no loading state. stub. by its name this is where the manager assign-to-team-member interface belongs.
- only assign action in shop-assembly module is "Assign to me" on Assemble List (expand a Pulled work unit). hardcoded to current user, no user picker.
- My Work lists only openings assigned to current user. worker view, not assignment tool.

backend supports it. assignOpenings(openingIds, assignedTo) accepts arbitrary user id, assigned openings to other users via API in earlier testing. only UI missing.

Assignments page or request-accept flow should let manager pick a shop-assembly team member, assign one or more pulled openings.

Xilous:
shop-assembly manager can now assign unassigned pulled openings to other team members.

Closes #330.

/app/shop-assembly/assignments (issue's URL) has trailing "s" -> falls through wildcard redirect -> empty stub.
real route /app/shop-assembly/assign, already renders self-claim drag-drop board. this PR adds on top.

backend
- shopAssemblyMembers query -> shop-assembly members holding "Shop Assembly User" or "Shop Assembly Manager", from Clerk list_users().
- require_role(info, role) auth helper gates it; only "Shop Assembly Manager" enumerates members.
- user_repository.list_shop_assembly_members() filters list_users() by role.
- assignOpenings unchanged; accepts arbitrary assignedToUserId + display name.

frontend
- ManagerAssignPanel on Assignments board. multi-select unassigned pulled openings -> pick team member -> assign one or more. per-member current-load summary.
- mounted above self-claim board, gated to "Shop Assembly Manager". plain "Shop Assembly User" sees board only.
- assigned work -> member's My Work (myWork filters by assignedToUserId).

decisions
- accept-flow assignment rejected; openings not PULLED until warehouse pulls. Assignments page is home.
- reassign/unassign of assigned openings out of scope; panel lists only unassigned. removeOpeningFromUser exists for fast-follow.

### DIFF
--- a/backend/app/auth.py
+++ b/backend/app/auth.py
@@ -20,6 +20,7 @@ from app.errors import AppError
 from app.repositories import user_repository
 
 ADMIN_ROLE = "Admin/Manager"
+SHOP_ASSEMBLY_MANAGER_ROLE = "Shop Assembly Manager"
 
 _CLERK_JWKS_URL = "https://api.clerk.com/v1/jwks"
 _JWKS_TTL_SECONDS = 3600.0
@@ -146,4 +147,24 @@ def require_admin(info) -> dict:
     roles = user_repository.get_user_roles(user_id)
     if ADMIN_ROLE not in roles:
         raise ForbiddenError("Admin/Manager role required")
+    return {"user_id": user_id, "roles": roles}
+
+
+def require_role(info, role: str) -> dict:
+    """Enforce that the caller holds a specific role (looked up via the Clerk Backend API, like
+    require_admin). Returns {user_id, roles}. Use for role-gated resolvers other than Admin/Manager,
+    e.g. the shop-assembly manager assignment tools (#330)."""
+    request = info.context["request"]
+    token = _bearer_token(request)
+    if not token:
+        raise AuthError("Authentication required")
+
+    claims = verify_clerk_token(token)
+    user_id = claims.get("sub")
+    if not user_id:
+        raise AuthError("Authentication token has no subject")
+
+    roles = user_repository.get_user_roles(user_id)
+    if role not in roles:
+        raise ForbiddenError(f"{role} role required")
     return {"user_id": user_id, "roles": roles}
--- a/backend/app/repositories/user_repository.py
+++ b/backend/app/repositories/user_repository.py
@@ -7,6 +7,10 @@ from app.errors import AppError
 
 CLERK_API_BASE = "https://api.clerk.com/v1"
 
+# Roles that make a user a shop-assembly team member (#330): assignable work in the shop-assembly
+# module. A manager can assign to a plain user or to another manager.
+SHOP_ASSEMBLY_ROLES = ("Shop Assembly User", "Shop Assembly Manager")
+
 _client = httpx.Client(base_url=CLERK_API_BASE, timeout=30.0)
 
 
@@ -78,6 +82,14 @@ def list_users() -> list[dict]:
         offset += limit
 
     return users
+
+
+def list_shop_assembly_members() -> list[dict]:
+    """Shop-assembly team members (#330): the subset of Clerk users holding a shop-assembly role,
+    for the manager assignment picker. Reuses list_users() and filters client-side to keep one
+    source of the user summary shape."""
+    members = set(SHOP_ASSEMBLY_ROLES)
+    return [u for u in list_users() if members.intersection(u["roles"])]
 
 
 def get_user(user_id: str) -> dict:
--- a/backend/app/schemas/shop_assembly.py
+++ b/backend/app/schemas/shop_assembly.py
@@ -122,7 +122,13 @@ class ShopAssemblyMutations:
             return shop_assembly_request_to_type(refreshed)
 
     @strawberry.mutation
-    def assign_openings(self, input: AssignOpeningsInput) -> list[ShopAssemblyOpening]:
+    def assign_openings(self, info: strawberry.Info, input: AssignOpeningsInput) -> list[ShopAssemblyOpening]:
+        """Assign pulled openings to a user (#330). Any signed-in user may self-assign (the "Assign to
+        me" board); assigning to *another* user is Shop Assembly Manager-gated, so the manager-only
+        guarantee holds at the data layer, not just the UI."""
+        auth = require_user(info)
+        if input.assigned_to_user_id != auth["user_id"]:
+            require_role(info, SHOP_ASSEMBLY_MANAGER_ROLE)
         opening_ids = [uuid.UUID(str(oid)) for oid in input.opening_ids]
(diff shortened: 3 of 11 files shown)


backend/app/schemas/shop_assembly.py
```diff
@@ -53,6 +54,13 @@ def shop_assembly_requests(
             )
             return [shop_assembly_request_to_type(r) for r in reqs]
 
+    @strawberry.field
+    def shop_assembly_members(self, info: strawberry.Info) -> list[ClerkUser]:
+        """Shop-assembly team members for the manager assignment picker (#330). Manager-gated: only
+        a Shop Assembly Manager may enumerate members and assign pulled openings to them."""
+        require_role(info, SHOP_ASSEMBLY_MANAGER_ROLE)
```
Xilous, line 61: This manager gate protects only the member list. The actual privileged action - assign_openings (line 125) - has no gate at all: no require_role and no require_user (it takes no info param). Since there is no global auth (every other mutation gates itself), any caller can POST assignOpenings with an arbitrary assignedToUserId + display name and assign work to any user, unauthenticated, bypassing the Shop Assembly Manager restriction this feature is built around.

The enforcement is at the wrong layer. Self-claim also uses assign_openings (assign to own id), so a blanket manager-only gate would break it - enforce require_user, plus require_role(info, SHOP_ASSEMBLY_MANAGER_ROLE) when assignedToUserId is not the caller's own id.

backend/app/auth.py
```diff
@@ -147,3 +148,23 @@ def require_admin(info) -> dict:
     if ADMIN_ROLE not in roles:
         raise ForbiddenError("Admin/Manager role required")
     return {"user_id": user_id, "roles": roles}
+
+
+def require_role(info, role: str) -> dict:
```
Xilous, line 153: require_role has no Admin/Manager override: it forbids anyone who does not literally hold the requested role. An Admin/Manager who lacks the "Shop Assembly Manager" role cannot enumerate shopAssemblyMembers, and the AssignmentBoard mount gate (hasRole('Shop Assembly Manager')) hides the panel from them too. If Admin/Manager is meant to be a superuser here, allow it through (e.g. `if role not in roles and ADMIN_ROLE not in roles`). If this exclusion is intended, ignore.

frontend/src/modules/shop-assembly/ManagerAssignPanel.tsx
```diff
@@ -0,0 +1,263 @@
...
+        </Button>
+      </Stack>
+
+      {loadByMember.length > 0 && (
+        <Stack direction="row" spacing={1} flexWrap="wrap" useFlexGap sx={{ mb: 2 }}>
+          <Typography variant="caption" color="text.secondary" sx={{ alignSelf: 'center' }}>
+            Current load:
+          </Typography>
+          {loadByMember.map((l) => (
+            <Chip key={l.name} size="small" variant="outlined" label={`${l.name}: ${l.count}`} />
```
Xilous: key={l.name} uses the display name as the React key. loadByMember groups counts by assignedToUserId (so counts are correct), but two distinct members who share a display name collide on this key, causing a React duplicate-key warning and one chip dropping. Carry the user id into the loadByMember value and key on it.

frontend/src/modules/shop-assembly/ManagerAssignPanel.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+  // Shares the GET_ASSEMBLE_LIST cache with the board (same query + vars), so a refetch here keeps
+  // both views in sync.
+  const { data: listData, refetch } = useQuery<{ assembleList: AssembleOpening[] }>(GET_ASSEMBLE_LIST);
+  const { data: memberData } = useQuery<{ shopAssemblyMembers: Member[] }>(GET_SHOP_ASSEMBLY_MEMBERS);
+
+  const openings = listData?.assembleList ?? [];
+  const members = memberData?.shopAssemblyMembers ?? [];
+
+  const available = useMemo(
```
Xilous: This available filter (pullStatus === 'PULLED' && assignedToUserId === null && assemblyStatus === 'PENDING') is a verbatim copy of the unclaimed-pool filter in AssignmentBoard.tsx (lines 203-205). The two now have to be kept in sync by hand. Extract a shared predicate (e.g. isUnclaimedPulled(o)) and use it in both.

backend/app/repositories/user_repository.py
```diff
@@ -80,6 +84,14 @@ def list_users() -> list[dict]:
     return users
 
 
+def list_shop_assembly_members() -> list[dict]:
```
Xilous, line 87: list_shop_assembly_members calls list_users(), which paginates every Clerk user (100/page) and filters client-side. Combined with the require_role gate on the resolver (a get_user_roles Clerk call), each shopAssemblyMembers query does 1 + N Clerk API round-trips. It is consistent with the existing users query and Apollo caches it after first fetch, so this is minor - just noting the cost since the panel mounts on every manager visit to the Assignments board.

Xilous (commented):
1. frontend/src/components/RequestsReviewPage.tsx line 108: const [reopenRequest] = useMutation(reopenMutation ?? acceptMutation, ...). reopenMutation is an optional prop and falls back to acceptMutation. In mode="approved", a caller that omits reopenMutation makes the Reopen button silently fire the ACCEPT mutation instead (re-mints the pull request, destructive) with no error surfaced. Both current callers pass reopenMutation so it works today, but the fallback hides a missing required prop. Guard handleReopen to no-op when reopenMutation is undefined, or type the props so reopenMutation is required when mode is "approved".

2. backend/app/repositories/shop_assembly_repository.py line 473: the reopen_shop_assembly_request docstring says "The minted PR is found via the openings' pull_request_id", but the code at line 494 finds it via request_number, and test_reopen_shop_assembly_discards_pr_when_openings_do_not_reference_it exists because the openings-derived lookup was wrong. Update the docstring.

The reopen/discard transaction ordering (unlink + flush before delete), the reopenable_only filters, the request_number-based PR lookup, and the FK handling all check out - no blocking correctness defect there.

Issue #353, opened 2026-07-27
Relay: make the backend<->relay link hands-off (enrol once, never touch the workstation again)


Xilous:
## Goal

**Enrol the relay once; never need physical access to that machine again.**

Today the relay effectively needs re-setup every few days. This issue holds the full root-cause analysis and a **self-contained implementation plan** — an implementation agent can be pointed at this issue and execute it without re-deriving any design decision. Read `CLAUDE.md` at the repo root first, especially the GP/relay integration section.

### Reframe the target state first

**A WebSocket cannot be made immortal.** Every backend deploy restarts the container and kills the socket; networks blip. The achievable goal is not "never disconnects" — it is:

> **every disconnection is invisible, self-healing, and never requires a human.**

---

## Current state (verified 2026-07-26)

- `relayStatus.connected` = `false`.
- Exactly one enrolment row: `TAGGING3W10`, company `TUBC`, hostname `Tagging3W10`, `enrolledAt` = `2026-07-24T22:54:27.662369`, `lastSeenAt` **byte-identical** (never advanced).
- Backend log: `"WebSocket /relay-link" 403` every ~30s, unbroken, **no `[accepted]` line ever** — including straight through the enrolment instant.
- Because `enroll_install` calls `encrypt_secret`, a successful enrolment proves `RELAY_SECRET_ENC_KEY` was valid at that moment. A 403 seconds later therefore means **the secret mismatched, not the key**.

**Diagnosis:** `enroll` rewrote `[auth] shared_secret` in `config.toml`, but the already-running relay process kept dialling with its old in-memory secret. `config.toml` on disk already matches the DB — the mismatch is purely in-memory. The relay needs a *restart*, not another enrolment. **Nobody currently has access to `Tagging3W10`**, and the relay's inbound server binds `127.0.0.1:7321`, so there is no remote restart path — PR B below is what recovers this class of failure without physical access.

**Consequence:** all GP writes fail. `createReceive` is the only path that puts hardware into inventory, so inventory is empty and the shop-assembly E2E plan is blocked.

---

## Decisions that change the issue as originally written

[...]

PR #356, opened 2026-07-27
refactor(relay): store the relay secret hashed, not encrypted (#353 PR C)

Xilous:
Implements **PR C** of #353. Branched off `master` after PR B (#354) merged — strictly sequential, no stacking. Migration chain: `067` → `down_revision = "066"`.

## Why

`RELAY_SECRET_ENC_KEY` is currently a single point of **permanent** failure. Rotate it or lose it and every enrolled relay is orphaned with no recovery path, because the plaintext exists nowhere the backend can reach. Hashing takes that key out of the authentication path entirely.

## Design decisions

- **SHA-256 hex, no KDF, no salt.** A KDF exists to make *low-entropy* secrets expensive to brute-force. This credential is not a human password: the relay generates it as `secrets.token_urlsafe(32)` — 256 bits of CSPRNG entropy, which stretching cannot improve on — while per-handshake bcrypt/argon2 cost is a real downside during a reconnect storm. The same reasoning already governs `hash_token`. The justification is in the code and the migration docstring.
- **Indexed equality, then `compare_digest` as the confirm.** Today's O(n) decrypt-scan becomes a single row fetch, and the constant-time property the docstring claims stays true.
- **Upgrade-in-place on a legacy match: write `secret_hash` AND set `secret_encrypted = NULL`.** Nulling rather than keeping both — leftover ciphertext means the key still matters, so a later rotation would turn those rows into "undecryptable", which `authenticate_secret` correctly reports as a key problem, raising a false alarm about a relay that is perfectly healthy. Nulling also makes the migration observable: `SELECT count(*) FROM relay_installs WHERE secret_encrypted IS NOT NULL` → 0 means the key is dead weight.
- **A missing `RELAY_SECRET_ENC_KEY` is no longer fatal.** `has_encryption_key()` is checked once before the legacy scan; absent key ⇒ skip it entirely and log once, not per row. `_fernet()` still raises `CONFIG_ERROR` when actually called.

## Changes

- `app/crypto.py` — `hash_secret`, `has_encryption_key`, shared `_sha256_hex`; module docstring rewritten (it previously asserted the secret *must* be reversible, which is the misconception this PR kills).
- `app/models/relay_install.py` — `secret_hash: String(64)`, nullable, indexed.
- `app/repositories/relay_repository.py` — `set_install_secret` (the single writer introduced in PR B) now writes the hash and nulls the ciphertext, so enrolment **and** adoption moved with one edit. `authenticate_secret` becomes the dual read described above. The reject-path diagnostic keeps #352's shape and gains `hash_rows` / `legacy_rows` / `encryption_key_present` so "wiped DB" / "key gone" / "key rotated" / "stale relay secret" stay distinguishable. The presented secret is never logged.
- **Migration `067`** — add column + `ix_relay_installs_secret_hash`. **No data migration**: the plaintext is not recoverable at migration time, and the dual read upgrades each row on its next handshake (~30 s for a live relay).
- `main.py` `reset_data` — `secret_hash` added to both the SELECT and the INSERT column lists. Missing that would re-create exactly the orphaned-relay bug #352 fixed, through a different column.

## Downgrade honesty

`067`'s downgrade drops `secret_hash`, orphaning any install that has no legacy ciphertext left — i.e. anything enrolled or adopted from here on. CI's migration-integrity job runs against an empty database so it passes cleanly; on a populated one, recovery is an admin-armed adopt window (PR B) or a re-enrolment. That ordering is exactly why PR B landed first. The caveat is stated in the migration docstring.

## Tests — `backend/tests/test_relay_secret_hash.py`

1. Enrol → `secret_hash` set, `secret_encrypted is None`, authenticates.
2. Legacy row (constructed directly, since nothing writes that shape any more) authenticates **and** is upgraded in place; still authenticates afterwards via the hash path.
3. Key absent → a hash-only install still authenticates.
4. Key absent → a legacy-only install returns `None` **without raising**, and logs the cause.
5. Wrong secret matches neither representation; blank returns `None`.
6. Adoption also stores a hash (proving `set_install_secret` really is the single writer).
7. `test_relay_installs.py::test_provision_returns_token_and_stores_hash_only` gains `secret_hash is None` at provision time.

## Verification

- CI green including the migration-integrity job.
- After deploy the live relay reconnects (it dials continuously) → log shows `relay install upgraded to hashed secret`; `relayStatus.connected` stays true across the deploy.
- **Then prove the key is dead:** temporarily unset `RELAY_SECRET_ENC_KEY` in Railway, redeploy, confirm the relay still authenticates. If `SELECT count(*) FROM relay_installs WHERE secret_encrypted IS NOT NULL` is 0, delete the variable for good.

## Note on the CLAUDE.md update

The issue asks for the `RELAY_SECRET_ENC_KEY` entry in CLAUDE.md's Environment section to be downgraded to "legacy rows only". **The root `CLAUDE.md` is gitignored in this repo** (`.gitignore:32`), so that edit cannot land in a PR — it has been made locally only. If it should be tracked, that is a separate change to `.gitignore`.

### DIFF
--- a/backend/alembic/versions/067_relay_secret_hash.py
+++ b/backend/alembic/versions/067_relay_secret_hash.py
@@ -1,0 +1,46 @@
+"""Store the relay's Bearer secret hashed, not encrypted (#353 PR C)
+
+Revision ID: 067
+Revises: 066
+Create Date: 2026-07-27
+
+Storing the relay secret reversibly made `RELAY_SECRET_ENC_KEY` a single point of PERMANENT failure:
+rotate or lose it and every enrolled relay is orphaned, because the plaintext exists nowhere the
+backend can reach. Hashing removes the key from the authentication path entirely.
+
+SHA-256 hex, no KDF and no salt, deliberately. The credential is not a human password - the relay
+generates it as `secrets.token_urlsafe(32)`, 256 bits of CSPRNG entropy, which stretching cannot
+improve on, while per-handshake bcrypt/argon2 cost is a real downside during a reconnect storm. The
+same reasoning already governs the enrollment-token hash.
+
+**No data migration.** The plaintext is not recoverable here, so this only adds the column and its
+index; `relay_repository.authenticate_secret` reads both representations and upgrades a legacy row in
+place on its next successful handshake - which for a live relay is within ~30 seconds. Once
+`SELECT count(*) FROM relay_installs WHERE secret_encrypted IS NOT NULL` reaches 0,
+`RELAY_SECRET_ENC_KEY` is dead weight and can be deleted from the environment.
+
+**Downgrade honesty.** Dropping `secret_hash` orphans any install that has no legacy ciphertext left
+- i.e. anything enrolled or adopted from here on. CI's migration-integrity job runs against an empty
+database so it passes cleanly, but on a populated one the downgrade means the relay must be recovered
+with an admin-armed adopt window (066/#353 PR B) or re-enrolled by hand. That is precisely why the
+adopt window landed first.
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision = "067"
+down_revision = "066"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column("relay_installs", sa.Column("secret_hash", sa.String(length=64), nullable=True))
+    op.create_index("ix_relay_installs_secret_hash", "relay_installs", ["secret_hash"])
+
+
+def downgrade() -> None:
+    op.drop_index("ix_relay_installs_secret_hash", table_name="relay_installs")
+    op.drop_column("relay_installs", "secret_hash")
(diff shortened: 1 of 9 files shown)


Xilous:
Follow-up on the first CI run: `test_authenticate_secret_survives_a_rotated_encryption_key` (added by #352) failed, correctly.

That test asserted that rotating `RELAY_SECRET_ENC_KEY` makes a **correct** relay secret fail — true while the secret was stored reversibly, and precisely the property that made one key a single point of permanent failure. With the secret hashed, the key is not on that path at all, so rotation is a non-event. The test now pins the inverse (`found is not None`), with a docstring explaining the inversion.

The clean-rejection behaviour #352 added still matters for pre-`067` rows and is covered in `test_relay_secret_hash.py`: a legacy row with the key gone returns `None` **without raising**, and logs the real cause.

Also moved the "legacy rows only" note onto `app/config.py` next to the variable, since the root `CLAUDE.md` is gitignored here.

PR #378, opened 2026-07-28
feat(shop-assembly): allocation builder (owed/short) + flag-time replacement reservations (#377)

### LINKED ISSUE #377: Shop assembly: allocation builder (owed/short) + flag-time replacement reservations
two workstreams changing how shop assembly requests (SAR) claim inventory. design agreed 2026-07-27, not implemented.

A. replace the all-or-nothing creation gate with an allocation builder. requester assigns available inventory to door leaves, partially-covered leaves allowed into the request, requester makes the send/don't-send call. each checklist line keeps its full owed (schedule) quantity and gains an allocated quantity; the difference is the short bucket.
B. reserve replacement stock at deficiency flag time, not at PR-REPL approval time.

locked decisions, not open for reinterpretation
1. partially-covered leaves stay in the request. no auto-exclusion of partial leaves. the shop assembly workflow is explicitly NOT all-or-nothing, the requester decides whether to send.
2. leaves with zero allocation across all their lines auto-drop from the request (an empty cart has nothing to pull, stage, or assemble). wizard shows them grayed out.
3. checklist representation is owed + short. ShopAssemblyOpeningItem keeps the full schedule quantity and gains allocated_quantity. short = quantity - allocated_quantity, always derived, never stored.
4. only SHOP_HARDWARE-classified items enter the SAR checklist. already true today, ImportWizard.tsx:540 filters, unchanged.
5. backfill of short leaves deferred to the future reallocation module. its contract - the current gap is derived at reallocation time by comparing the leaf's requirements in the CURRENT schedule version against what is physically recorded on the leaf (installed OpeningItemHardware + deficient outstanding + replacement pending). the short list is evidence, never the authority. mirrors the snapshot philosophy in _handle_schedule_replacement, in-flight requests are never rewritten by a re-upload.
6. replacement reservations are project-scoped by construction, everything in the reservation/availability chain is keyed by project_id and no cross-project claim is expressible. reserve min(available, condemned) at flag time. top up when availability rises. notify only when fully covered.
7. leaf assignment lock, NO change. current #340 behavior stands - unassign-to-pool and manager reassign allowed mid-assembly, self-claim theft forbidden.

current state anchors
- creation gate + SAR creation - backend/app/repositories/import_repository.py, finalize_import_session SAR section (~line 1196), _gate_on_available_inventory (~line 336). refuses the whole finalize via InventoryShortfallError if the selection does not fit available = on-hand - deficient - active reservations.
- reservations - backend/app/models/inventory_reservation.py (check constraint ck_inventory_reservations_source_matches_request ties source to SAR/shipping FKs), backend/app/repositories/warehouse/reservations.py (create_reservations aggregates needs per combo, one row per combo per request).
- accept - shop_assembly_repository.accept_shop_assembly_request (~line 1052), mints PullRequest (SHOP_ASSEMBLY, PENDING) with one LOOSE line per SAO item, repoints ShopAssemblyOpening.pull_request_id.
- pull approval - warehouse/pull_requests.py approve_pull_request (~line 240), all-or-nothing, releases the holder's reservation then deducts FIFO under one set of row locks. INSUFFICIENT leaves the pull PENDING/blocked.
- holder resolution - warehouse/pull_requests.py find_reservation_holder (line 208), SAR via request_number match, shipping via pull_request_id, returns None for PR-REPL pulls today.
- staging - stage_pull_openings (~line 763), per-opening (per-leaf) cart confirmation, no inventory movement.
- progress - shop_assembly_repository.record_assembly_progress (~line 269), installed absolute, flag-deficient incremental.
- deficiency flag - backend/app/repositories/stock/deficiency.py report_deficiency_at_assembly (line 142), bumps quantity AND deficient_quantity on an InventoryLocation row (available unchanged), mints/merges a LOOSE line carrying leaf + sa_opening_item_id on a PENDING PR-REPL pull, new pull numbered PR-REPL-{basis}-N when the prior one is no longer PENDING.
- completion - shop_assembly_repository.complete_opening (~line 421), refuses while units unaccounted for, refuses all-deficient (zero installed), mints OpeningItem (state IN_INVENTORY, quantity 1, leaf-stamped) + OpeningItemHardware.
- SAO item constraint - backend/app/models/shop_assembly.py ~line 140, installed_quantity + deficient_quantity + replacement_pending_quantity <= quantity, all >= 0, completion requires equality.
- re-upload - import_repository._handle_schedule_replacement (~line 428), PENDING requests trimmed of vanished openings, reservations rebuilt from surviving items (line 492), auto-reject when empty, integrity_note set. APPROVED requests untouched except the note.
- replacement arrivals - warehouse/pull_requests.py _apply_replacement_arrivals (~line 399), routes arrived units by leaf status (deficient drops for PENDING/IN_PROGRESS, moves to replacement_pending for COMPLETED).
- unblock notification - warehouse/receiving.py notify_unblocked_replacement_pulls (~line 402), fires PULL_UNBLOCKED when a receive makes a blocked PR-REPL pull sufficient.
- deficiency review - stock/deficiency.py resolve_deficiency (line 354), REPAIR decrements deficient_quantity only (unit becomes available again), never touches PR-REPL lines and does not need to, the replacement pull can pull the repaired unit as free stock.
- wizard step - frontend/src/modules/import/ShopAssemblyStep.tsx, demand vs availability table, red shortfall alert, Next disabled on any shortfall. drafts derived in ImportWizard.tsx (~line 529), SHOP_HARDWARE filtered, grouped per leaf.
- cancel - warehouse/pull_requests.py cancel_pull_request (line 970), restocks, re-creates the source request's reservation (or flags via integrity_note if it no longer fits). discard_pending_pull_request handles PENDING pulls (reopen path).

workstream A - allocation builder + owed/short bucket

A1. data model
- add allocated_quantity (Integer, NOT NULL) to shop_assembly_opening_items.
- Alembic migration backfills existing rows with allocated_quantity = quantity, every pre-feature request was fully covered by construction (server_default then drop, or an UPDATE in the migration).
- replace the bucket check constraint with installed_quantity >= 0 AND deficient_quantity >= 0 AND replacement_pending_quantity >= 0 AND installed_quantity + deficient_quantity + replacement_pending_quantity <= allocated_quantity AND allocated_quantity >= 0 AND allocated_quantity <= quantity. keep ck_shop_assembly_opening_items_quantity_positive.
- allocated_quantity = 0 is legal, a line fully short on a leaf that has other allocated lines.
- short derived everywhere as quantity - allocated_quantity. no stored column, no separate state.

A2. GraphQL surface
- SAROpeningItemInput (backend/app/schemas/inputs.py line 141) - add allocated_quantity: int | None = None. None means fully allocated (= quantity) so non-wizard callers and existing tests keep working. wizard always sends it explicitly.
- Strawberry type ShopAssemblyOpeningItem (app/schemas/types.py) - add allocated_quantity, update the converter in app/schemas/converters.py.
- update the SDL reference backend/app/graphql/shop_assembly.graphql.
- frontend - add allocatedQuantity to the SAO item selections in frontend/src/graphql/shop-assembly.ts (GET_SHOP_ASSEMBLY_REQUESTS, RECORD_ASSEMBLY_PROGRESS, the my-work/assemble-list queries) and to PULL_STAGING_OPENING_FIELDS in frontend/src/graphql/warehouse.ts.
- no new enums, the two-enum-file sync rule is not in play for A.

A3. finalize (import_repository.finalize_import_session, SAR section)
- validate per line 0 <= allocated_quantity <= quantity, else ValidationError.
- refuse a leaf whose every line has allocated_quantity == 0 (ValidationError). the wizard auto-drops such leaves before submitting, the server check is a race/bug backstop.
- sar_needs becomes the ALLOCATED quantities, lines with allocated > 0 only. _gate_on_available_inventory itself unchanged, it now gates the allocation instead of the full demand. it still refuses the whole finalize on a race (availability shrank between wizard load and submit); the wizard catches InventoryShortfallError, refreshes availability, re-runs auto-assign, asks the user to review again.
- create_reservations from the allocated needs.
- SAO items created with quantity = owed (unchanged) + allocated_quantity from input (default = quantity when the input field is None).
- after a successful partial finalize, fire notification_service.notify_po_shortfall for the total short per combo. short combos are out of available stock by construction so this is always a genuine purchasing signal. today this notification fires only on the refusal path in app/schemas/imports.py; that path stays for races.

A4. accept (accept_shop_assembly_request)
- mint pull lines only for SAO items with allocated_quantity > 0, requested_quantity = allocated_quantity.
- reopen_shop_assembly_request, no changes. the reservation is held through a reopen and was minted from allocations, a re-accept re-mints the pull from the same allocations.
- everything else unchanged. because the pull asks for exactly what the reservation holds, approve_pull_request stays all-or-nothing by construction, no warehouse-side changes in this workstream.

A5. progress + completion
- record_assembly_progress - cap check becomes installed + deficient (+ replacement_pending) <= allocated_quantity, enforced by the new DB constraint. update the ValidationError message to name the allocated quantity.
- complete_opening - per-line accounting check becomes installed + deficient + replacement_pending == allocated_quantity. short units excused, they were never pulled.
- zero-installed guard unchanged, refuse completion when every line has installed_quantity == 0. covers both the all-deficient leaf and the all-short leaf, neither may mint an assembled OpeningItem with nothing on it.
- install_replacement and _apply_replacement_arrivals, no changes. they operate inside the deficient/replacement_pending buckets, which live inside allocated.

A6. re-upload reconciliation
- _handle_schedule_replacement line 492 - rebuild surviving reservations from item.allocated_quantity instead of item.quantity.

A7. shipping the short leaf takes a decision
- extend the incomplete-leaf gate at shipping finalize (import_repository ~line 925, get_awaiting_replacement_quantities -> ValidationError naming every flagged leaf -> acknowledge_incomplete_leaves) to also count short units. correlate the OpeningItem back to the ShopAssemblyOpening that produced it (latest_assembly correlation exists in shop_assembly_repository ~line 669) and sum quantity - allocated_quantity across its items.
- message names both kinds, "N unit(s) awaiting replacement, M unit(s) never pulled". acknowledge flow unchanged.

A8. allocator UI (rewrite ShopAssemblyStep.tsx, derivations in ImportWizard.tsx)
- reused inputs - openingDrafts (one per leaf, SHOP_HARDWARE aggregated), availabilityByCombo (reservation-aware, from projectInventoryAvailability).
- extract auto-assign as a pure function (e.g. frontend/src/modules/import/allocation.ts), (drafts, availabilityByCombo) -> per-leaf per-line allocations. iterate drafts in their existing order (schedule order), per line allocate min(owed, remaining pool for that combo), pool starts at availableQuantity per combo. deterministic, first-leaf-first.
- co-located Vitest tests (__tests__/allocation.test.ts) - full coverage, scarcity ordering, zero pool, multi-combo independence.
- auto-assign runs once on entry to the step. manual edits never clobbered by a silent re-run; re-run only via an explicit "Re-run auto-assign" button or after a race refusal.
- layout top to bottom
  1 request number field, unchanged.
  2 combo summary table, the existing demand table extended - per combo available / allocated across leaves / remaining pool / total short. short > 0 renders as a warning-colored count, not an error.
  3 per-leaf list, one card or table section per draft (opening number + leafSuffix + building/floor). each line - category, product code, owed, allocated stepper clamped to [0, min(owed, allocated + remaining pool)], short badge ("N short") when allocated < owed.
  4 leaf-level chip - Fully covered / Partial / Not covered, plus an include/exclude toggle (excluding a leaf returns its allocation to the pool). not-covered leaves (all lines zero) grayed out, labeled auto-dropped, toggle disabled, excluded from the payload.
- Next gating - request number non-empty AND at least one included leaf with allocated > 0 AND availability loaded without error. the shortfall alert becomes an informational summary of what will be sent short.
- on InventoryShortfallError at finalize - refetch availability, re-run auto-assign, surface an alert that availability changed and the allocation must be reviewed before resending.
- keep the same-derivation-not-a-parallel-one property. the drafts shown are exactly the SAROpeningInput list submitted, now with allocatedQuantity per item, minus auto-dropped and user-excluded leaves.

A9. downstream display
- accept UI (ShopAssemblyRequestsPage.tsx) - show allocated vs owed per line and a request-level short summary, so the acceptor sees exactly what they are approving.
- bench (AssemblyDetailModal.tsx, MyWorkPage, AssignmentBoard, openingFilters.assemblyProgress rollup) - add allocatedQuantity to the OpeningItem interfaces, remaining = allocated - installed - deficient - replacementPending, render short as a per-line "N never pulled" annotation so a leaf does not read as permanently unfinished.
- warehouse staging cart (per-opening cart view fed by PULL_STAGING_OPENING_FIELDS) - pick count = allocated_quantity. the pull LINES already carry allocated as requested_quantity per A4, this is the per-opening cart display. show short as context.
- pipeline (#344) - add a short-units rollup per request. backend assemblyPipeline query and assembly_pipeline_to_type / assembly_pipeline_summary_to_type converters (app/schemas/shop_assembly.py + app/schemas/converters.py). frontend PipelinePage.tsx.

workstream B - flag-time replacement reservations

B1. schema
- add REPLACEMENT_PULL to ReservationSource in BOTH backend/app/models/enums.py and backend/app/schemas/enums.py, the two enum files must stay in sync per repo convention.
- InventoryReservation - add pull_request_id FK (nullable, ondelete CASCADE) + index. update ck_inventory_reservations_source_matches_request to a three-way constraint, each source requires exactly its own FK and null others.
- Alembic migration.

B2. mint at flag time (report_deficiency_at_assembly)
- after the PR-REPL line is minted or merged, compute reservation-aware availability for the combo, reserve_qty = min(available, quantity flagged now). find an existing reservation row for (REPLACEMENT_PULL, replacement_pr.id, combo) and increment, else create.
- reserving less than condemned is expected, top-up hooks close the gap later.
- the reservation claims free stock only. availability is already net of everyone else's reservations, so a replacement can never eat stock another request has claimed, preserving the guarantee in find_reservation_holder's docstring.

B3. holder resolution + consumption
- find_reservation_holder - when a SHOP_ASSEMBLY pull's request_number matches no SAR, detect a replacement pull structurally (any line with sa_opening_item_id set, same test as find_pending_replacement_pulls) and return (REPLACEMENT_PULL, pr.id). update the docstring, the "replacement pulls hold no claim" case is superseded.
- release_reservations - support the new source keyed by pull_request_id (extend the lookup or add a sibling helper, executor's choice; approve/cancel/discard must all go through it).
- approve_pull_request consumes the replacement reservation exactly like the others, release holder then FIFO deduct, sufficiency checked with exclude_reservations_of = holder. a partially-reserved replacement pull whose remainder does not fit free stock stays INSUFFICIENT/blocked. PO backfill loop unchanged.

B4. top-up hooks
- factor a helper (e.g. top_up_replacement_reservations(session, project_id, combos)) that, for each PENDING replacement pull wanting one of the combos, raises its reservation toward its total need by min(need - reserved, free available), oldest pull first.
- call site 1, notify_unblocked_replacement_pulls (receiving) - top up FIRST, then run its sufficiency check with exclude_reservations_of = (REPLACEMENT_PULL, pr.id). the exclude is mandatory, without it the pull's own new reservation makes it read insufficient forever. notify only when fully covered, current behavior preserved.
- call site 2, resolve_deficiency for resolutions that raise availability (REPAIR, and SEND_TO_STOCK's project-side path) - same helper, so a repaired unit is claimed for the replacement pull before another request can take it.

B5. release paths
- discard_pending_pull_request - release (REPLACEMENT_PULL, pr.id).
- cancel_pull_request of an approved replacement pull - reservation already consumed at approval, restock only, re-create nothing, a replacement pull has no source request to hand back. PENDING replacement pulls go through discard, which now releases.
- resolve_deficiency needs no PR-REPL line unwinding, unchanged. a REPAIR feeds the unit back to availability and the top-up hook claims it for the replacement pull.

three things this issue deliberately does NOT build
- backfilling short leaves when stock arrives later, that is the reallocation module (future) under the derive-from-current-schedule contract in decision 5.
- any change to leaf assignment locking (decision 7).
- frames, excluded from the leaf model entirely per #311.

invariants that must hold after both workstreams
1. reservations never exceed reservation-aware availability at the moment they are written (creation gate, flag-time mint, top-up).
2. every pull line minted from a SAR equals its reservation, so approve_pull_request keeps its all-or-nothing property with zero changes to its logic in workstream A.
3. the three progress buckets partition allocated_quantity; allocated_quantity <= quantity; short is always derived.
4. one live OpeningItem per (opening_id, leaf), unchanged (#345 unique index).
5. a replacement reservation can only ever claim stock in the deficient leaf's own project, and never stock another request has reserved.

suggested PR breakdown
- PR 1 (backend, workstream A) - migration + model + finalize/accept/progress/completion/re-upload changes + PO shortfall notification + tests. ships dark, with no wizard changes every request still arrives fully allocated via the input default, so behavior is identical until PR 2.
- PR 2 (frontend, workstream A) - allocator step + downstream display (accept UI, bench, staging cart, pipeline).
- PR 3 (workstream A7) - shipping incomplete-leaf gate extension. small, may fold into PR 1.
- PR 4 (workstream B) - schema + flag-time mint + holder/consumption + top-up + release paths + tests.
- order 1 -> 2 -> 3/4 in either order.

backend tests, extend the existing suites rather than starting parallel ones
- test_inventory_sufficiency_gates.py - partial allocation reserves only allocated; race refusal at finalize with partial allocation; PO shortfall notification fired for short combos on successful send.
- test_accept_requests.py - accept mints lines only for allocated > 0 with requested_quantity = allocated; zero-line leaf refused at finalize.
- test_assembly_progress.py - progress capped at allocated (not owed); completion accounting against allocated.
- test_shop_assembly_completion_guards.py - completion excuses short units; zero-installed refusal still covers all-short.
- test_replacement_loop.py - flag mints reservation min(available, condemned); merge on repeated flags; approve consumes; blocked when partially reserved and free stock short; top-up on receive; top-up on REPAIR; discard releases; notify fires only when fully covered and excludes self.
- new test_sar_allocation.py for the finalize allocation matrix (defaults, validation bounds, re-upload rebuild from allocated).
- migration integrity upgrade/downgrade/upgrade + alembic check already runs in CI.

frontend tests
- Vitest, co-located per repo convention (__tests__/*.test.{ts,tsx}).
- allocation.test.ts - the auto-assign matrix from A8 (full coverage, scarcity ordering, zero pool, multi-combo independence, determinism).
- stepper clamp math and leaf-status derivation (Fully covered / Partial / Not covered), prefer extracting them as pure helpers so they are testable without component rendering.

deliberately left to the executor
- exact naming of new helpers, fields, and GraphQL selections, follow the neighboring code.
- module placement of top_up_replacement_reservations (receiving.py vs reservations.py).
- component structure inside the A8 layout (cards vs table sections) and MUI styling details.
- whether PR 3 folds into PR 1.

simulated user testing, Chrome DevTools MCP against Railway after deployment, frontend https://frontend-production-34fc.up.railway.app/
1. reset or use a scratch project. import a hardware schedule (Start a Task -> Import) with at least 2 openings x 2 leaves carrying SHOP_HARDWARE, classify at least two products as Shop.
2. receive inventory such that one combo is deliberately short (e.g. schedule needs 8 hinges total, receive 5) and another combo is fully covered.
3. Start a Task -> shop assembly flow -> allocator step. verify auto-assign fills leaves in schedule order until the hinge pool is exhausted; combo summary shows available/allocated/remaining/short; later leaves show a short badge; a leaf with nothing allocated (if constructed) is grayed out and labeled auto-dropped; Next is ENABLED despite the shortfall.
4. lower hinges on leaf 1, verify the freed units become assignable on a later leaf via its stepper clamp. exclude one coverable leaf via its toggle, verify its allocation returns to the pool.
5. send the request. verify a PO shortfall notification exists for the short combo.
6. Shop Assembly -> Requests, the pending request shows allocated vs owed per line and a short summary. accept it.
7. Warehouse -> Pull Requests, approve the pull. verify it approves (no INSUFFICIENT) and the pull lines show allocated quantities.
8. stage each opening, verify carts list allocated pick counts with short context.
9. assign a leaf, open the bench modal. lines show owed / allocated / short ("never pulled"); record progress up to allocated; verify the stepper/save refuses beyond allocated.
10. flag 1 unit deficient with a reason. check the availability view (or a second draft request's allocator), verify the replacement reservation reduced available for that combo in this project.
11. complete the leaf (all allocated units installed or deficient, at least one installed, aisle/row/bay filled). verify completion succeeds with short units outstanding and the OpeningItem appears in warehouse assembled inventory.
12. receive the replacement combo. verify the PULL_UNBLOCKED notification appears once fully covered, approve the PR-REPL pull, verify it consumes the reservation (available unchanged by the approve beyond the deduction) and the deficient unit's expectation is restored on the leaf (replacement pending on the completed leaf).
13. attempt a shipping-out request for the short leaf. verify the incomplete-leaf decision names both "awaiting replacement" and "never pulled" counts and requires acknowledgment.

Xilous:
closes #377.

shop assembly no longer all-or-nothing at creation. requester assigns available inventory to leaves, partially-covered leaves stay in the request, send/don't-send call is theirs.

workstream A - allocation builder + owed/short bucket

`ShopAssemblyOpeningItem` gains `allocated_quantity`.
`quantity` keeps meaning owed, never reduced by scarcity.
short = `quantity - allocated_quantity`, derived everywhere, stored nowhere - the authority on what a leaf is still missing is the current schedule against what is on the leaf, which is the reallocation module.
three progress buckets partition `allocated_quantity`, not `quantity`.

- migration 070 adds the column, backfills `allocated = quantity` for every existing row (all cleared the all-or-nothing gate, so fully covered by construction), rewrites the progress check constraint.
- `SAROpeningItemInput.allocated_quantity` defaults to None = fully allocated, non-allocator callers unchanged.
- finalize validates 0 <= allocated <= quantity, refuses a leaf with nothing allocated on any line, gates and reserves over the allocation, fires the PO shortfall notification for short combos on the success path.
- creation gate is now the race gate. refusal means availability moved, wizard refetches, re-runs auto-assign, asks the user to review.
- accept mints pull lines only for allocated > 0 at the allocated quantity, so the pull equals the reservation and `approve_pull_request` needs no change.
- `record_assembly_progress` caps at allocated, `complete_opening` accounts against allocated and excuses short units. zero-installed guard unchanged, now covers the all-short leaf.
- re-upload rebuilds surviving reservations from allocated, not owed.
- shipping incomplete-leaf gate counts short units as well as awaiting-replacement ones, names both per leaf - waiting fixes one and not the other.

allocator step - rewritten `ShopAssemblyStep.tsx` + pure `allocation.ts`
- auto-assign fills whole leaves in schedule order until each combo's pool runs out
- per-line steppers clamped to owed and remaining pool
- leaf include/exclude toggle returns allocation to the pool
- zero-coverage leaves greyed out and auto-dropped
- shortfall alert informational, no longer blocks Next

downstream display
- accept UI shows allocated vs owed plus a request-level short chip
- bench modal shows owed / pulled / "N never pulled"
- staging cart picks the allocated count
- pipeline gains short unit and short opening rollups

workstream B - flag-time replacement reservations

PR-REPL pulls now hold a claim from the flag instead of competing for stock at approval.

`ReservationSource.REPLACEMENT_PULL`
`inventory_reservations.pull_request_id` (CASCADE, indexed)
three-way source/FK check constraint
migration 071

- `report_deficiency_at_assembly` reserves `min(free stock, condemned)` at the flag. reserving less is the normal case - the shelf is usually empty, which is why a replacement is needed.
- `find_reservation_holder` detects a replacement structurally (any line with `sa_opening_item_id`) and returns the pull itself as holder. `approve_pull_request` consumes it like any other claim.
- `top_up_replacement_reservations` raises each PENDING replacement toward its need, oldest pull first, on a receive and on a REPAIR resolution.
- `notify_unblocked_replacement_pulls` tops up first, then checks coverability with the pull's own claim excluded - without the exclusion it reads insufficient forever.
- `discard_pending_pull_request` releases. cancel of an approved replacement releases nothing, approval already consumed it.
- partly-reserved replacement that comes up short is excluded from the RESERVED_PULL_SHORT integrity audit - it is the expected population, and flagging every one is how a real discrepancy gets missed.

two deviations from the spec

- spec listed SEND_TO_STOCK as a second availability-raising resolution. it is not - it decrements `quantity` and `deficient_quantity` together, so project available is unchanged. only REPAIR calls the top-up.
- spec said add REPLACEMENT_PULL to both enum files. `ReservationSource` is not in `app/schemas/enums.py` - DB-only enum, no GraphQL surface - so only `app/models/enums.py` changed.

state

- frontend 294 tests pass, lint and build clean
- backend ruff check and format clean
- DB-backed backend tests are skipped locally (no DATABASE_URL, no local postgres), so the new suites run for the first time in CI
- two `test_manufacturer_match` failures on this machine are pre-existing on master and unrelated
- simulated user testing against Railway not yet run

### DIFF
--- a/backend/app/models/shop_assembly.py
+++ b/backend/app/models/shop_assembly.py
@@ -172,7 +172,13 @@ class ShopAssemblyOpeningItem(Base):
     #
     # 0 is legal: a line can be fully short on a leaf that other lines do cover. A leaf with *every*
     # line at 0 has an empty cart and is dropped before it reaches here.
-    allocated_quantity: Mapped[int] = mapped_column(Integer, nullable=False, default=0, server_default="0")
+    #
+    # **No default, client-side or server-side.** Migration 070 uses a server default only to get the
+    # NOT NULL column onto a populated table and drops it in the same step, and mirroring a default
+    # here would put the footgun straight back: an insert that forgot this column would quietly write
+    # 0, pass the check constraint, and produce a line that reads as entirely short. Omitting it is
+    # meant to fail loudly.
+    allocated_quantity: Mapped[int] = mapped_column(Integer, nullable=False)
     # Per-item assembly progress (#340). The assembler records these incrementally at the bench;
     # completion reads them rather than taking an ephemeral checklist as input.
     #   installed_quantity - units physically fitted to the leaf. Absolute, editable both directions
--- a/backend/app/models/shop_assembly.py
+++ b/backend/app/models/shop_assembly.py
@@ -132,14 +132,17 @@ class ShopAssemblyOpeningItem(Base):
             "shop_assembly_opening_id",
         ),
         CheckConstraint("quantity >= 1", name="ck_shop_assembly_opening_items_quantity_positive"),
-        # Progress can never exceed the planned quantity, and no count can go negative (#340/#341).
-        # The three buckets partition the line: installed + deficient + replacement_pending <=
-        # quantity, and completion requires equality. That is what lets a replacement arriving on an
-        # already-completed leaf move a unit out of `deficient` without the leaf reading as
-        # un-dispositioned - the unit lands in replacement_pending, and the sum is unchanged.
+        # Progress can never exceed what was actually pulled for the line, and no count can go
+        # negative (#340/#341). The three buckets partition `allocated_quantity`, not `quantity`:
+        # installed + deficient + replacement_pending <= allocated, and completion requires
+        # equality. That is what lets a replacement arriving on an already-completed leaf move a unit
+        # out of `deficient` without the leaf reading as un-dispositioned - the unit lands in
+        # replacement_pending, and the sum is unchanged - while the short units (quantity -
+        # allocated) sit outside the partition entirely, because they were never pulled.
         CheckConstraint(
             "installed_quantity >= 0 AND deficient_quantity >= 0 AND replacement_pending_quantity >= 0 "
-            "AND installed_quantity + deficient_quantity + replacement_pending_quantity <= quantity",
+            "AND allocated_quantity >= 0 AND allocated_quantity <= quantity "
+            "AND installed_quantity + deficient_quantity + replacement_pending_quantity <= allocated_quantity",
             name="ck_shop_assembly_opening_items_progress_within_quantity",
         ),
         # The replacement-install work queue and the shipping deficiency guard both scan for lines
@@ -156,6 +159,20 @@ class ShopAssemblyOpeningItem(Base):
     hardware_category: Mapped[str] = mapped_column(String, nullable=False)
     product_code: Mapped[str] = mapped_column(String, nullable=False)
     quantity: Mapped[int] = mapped_column(Integer, nullable=False)
+    # What the schedule says this leaf is owed, vs what the requester could actually claim out of
+    # available inventory when the request was sent. Shop assembly is not all-or-nothing: a partially
+    # covered leaf still goes to the bench, and the requester makes the informed send/don't-send call.
+    #   quantity           - owed. The schedule's number, never reduced by scarcity.
+    #   allocated_quantity - what was reserved, pulled and will physically arrive on the cart.
+    # **Short is derived, never stored**: short = quantity - allocated_quantity. There is no separate
+    # state for it and nothing downstream may write one, because the authority on what is still
+    # missing is the *current* schedule compared against what is physically on the leaf - the short
+    # count here is evidence of what this request executed, not a backlog anybody works off.
+    # Backfilling a short leaf when stock turns up later belongs to the reallocation module.
+    #
+    # 0 is legal: a line can be fully short on a leaf that other lines do cover. A leaf with *every*
+    # line at 0 has an empty cart and is dropped before it reaches here.
+    allocated_quantity: Mapped[int] = mapped_column(Integer, nullable=False, default=0, server_default="0")
     # Per-item assembly progress (#340). The assembler records these incrementally at the bench;
     # completion reads them rather than taking an ephemeral checklist as input.
     #   installed_quantity - units physically fitted to the leaf. Absolute, editable both directions
@@ -163,8 +180,9 @@ class ShopAssemblyOpeningItem(Base):
     #   deficient_quantity - units found defective and already handed to the deficiency flow
     #     (returned to inventory flagged deficient + a PR-REPL replacement line). Only ever grows;
     #     undoing a deficiency is deficiency review's job, not the assembler's.
-    # Remaining (quantity - installed - deficient) is derived, never stored; completion is refused
-    # while any line still has remaining > 0.
+    # Remaining (allocated - installed - deficient) is derived, never stored; completion is refused
+    # while any line still has remaining > 0. It counts against `allocated_quantity`, not `quantity`:
+    # a short unit was never pulled, so there is nothing for the assembler to disposition.
     installed_quantity: Mapped[int] = mapped_column(Integer, nullable=False, default=0, server_default="0")
     deficient_quantity: Mapped[int] = mapped_column(Integer, nullable=False, default=0, server_default="0")
     # Units whose replacement hardware has arrived but is not on the leaf yet (#341). Only ever
--- a/backend/app/repositories/import_repository.py
+++ b/backend/app/repositories/import_repository.py
@@ -371,6 +371,44 @@ def _gate_on_available_inventory(
         )
 
 
+def _allocated_quantity(item_input: dict) -> int:
+    """What a checklist line actually claims out of inventory.
+
+    A missing / null `allocated_quantity` means **fully allocated**. That default is what keeps every
+    caller that predates the allocator - tests, scripts, a browser tab loaded before the deploy - on
+    the old all-or-nothing behaviour: it asks for the whole line, and the availability gate bounces
+    the finalize if the whole line does not fit. The allocator step always sends the number.
+    """
+    allocated = item_input.get("allocated_quantity")
+    return item_input["quantity"] if allocated is None else allocated
+
+
+def _validate_leaf_allocation(opening_number: str, sa_opening_input: dict) -> None:
+    """Per-line bounds, and the one leaf-level rule the allocator enforces client-side.
+
+    A leaf with nothing allocated on any line is an empty cart: nothing to pull, nothing to stage,
+    nothing to assemble, and `complete_opening` would refuse it at the far end for having zero
+    installed units. The wizard drops such leaves before submitting, so reaching here means a race
+    (availability collapsed between load and send) or a non-UI caller - both of which are better
+    refused than turned into a work unit that can never be finished.
+    """
+    items = sa_opening_input.get("items", [])
+    for item in items:
+        allocated = _allocated_quantity(item)
+        if allocated < 0 or allocated > item["quantity"]:
+            raise ValidationError(
+                f"{_opening_leaf_label(opening_number, sa_opening_input.get('leaf'))} {item['product_code']}: "
+                f"allocated quantity {allocated} must be between 0 and the {item['quantity']} unit(s) owed.",
+                field="allocated_quantity",
+            )
+    if items and all(_allocated_quantity(item) == 0 for item in items):
+        raise ValidationError(
+            f"{_opening_leaf_label(opening_number, sa_opening_input.get('leaf'))} has no hardware allocated to it, "
+            "so there is nothing to pull for it. Drop it from the request, or wait for stock.",
+            field="allocated_quantity",
+        )
+
+
 def _live_shop_assembly_requests(session: Session, project_id: uuid.UUID) -> list[SARModel]:
     """Shop-assembly requests in a project that are still in flight: PENDING, or APPROVED while the
     pull they minted has not been worked (PENDING). Once the warehouse approves the pull, inventory
@@ -489,7 +527,13 @@ def _handle_schedule_replacement(
                 sar.project_id,
                 ReservationSource.SHOP_ASSEMBLY_REQUEST,
                 sar.id,
-                [(item.hardware_category, item.product_code, item.quantity) for o in sar.openings for item in o.items],
+                # Allocated, not owed: the claim it held was minted from the allocation, and rebuilding
+                # from `quantity` would inflate it into a claim on hardware this request never had.
+                [
+                    (item.hardware_category, item.product_code, item.allocated_quantity)
+                    for o in sar.openings
+                    for item in o.items
+                ],
             )
         sar.integrity_note = SCHEDULE_CHANGED_DROPPED_NOTE if lost else SCHEDULE_CHANGED_NOTE
 
@@ -911,34 +955,46 @@ def finalize_import_session(
                 ).all()
             }
 
-        # Deficiency guard (#341). A leaf whose source checklist still has condemned-and-unreplaced
-        # units - or units whose replacement has arrived but is not fitted yet - is physically short
-        # of the hardware list it ships under. The business decision is warn + confirm, never silent
-        # and never a hard block: deliberate short-shipping is a real workflow (reallocation exists
-        # for it), but it has to be a decision someone made. So it is refused here unless the caller
-        # says explicitly that it knows, and the wizard's confirm is what sets that flag.
+        # Incomplete-leaf guard (#341). A leaf can be physically short of the hardware list it ships
+        # under for two unrelated reasons, and the shipper needs to see both because the remedies are
+        # not the same:
+        #   - **awaiting replacement** - a unit arrived and failed. A replacement pull already exists;
+        #     waiting is a real option.
+        #   - **never pulled** - the unit was never available when the request was sent, so it was
+        #     allocated 0 and no pull was ever raised for it. Waiting does nothing; that gap is
+        #     closed by purchasing and, later, by reallocation.
+        # The business decision is warn + confirm, never silent and never a hard block: deliberate
+        # short-shipping is a real workflow, it just has to be a decision someone made. Refused here
+        # unless the caller says explicitly that it knows, and the wizard's confirm sets that flag.
         #
-        # One grouped scalar aggregate for every referenced leaf, not a per-line count.
+        # Two grouped scalar aggregates for every referenced leaf, not a per-line count.
         if referenced_oi_ids and not acknowledge_incomplete_leaves:
             from app.repositories import shop_assembly_repository
 
-            flagged_leaves = shop_assembly_repository.get_awaiting_replacement_quantities(
-                session, list(referenced_oi_ids)
-            )
-            if flagged_leaves:
+            oi_ids = list(referenced_oi_ids)
+            flagged_leaves = shop_assembly_repository.get_awaiting_replacement_quantities(session, oi_ids)
+            never_pulled = shop_assembly_repository.get_never_pulled_quantities(session, oi_ids)
+            if flagged_leaves or never_pulled:
                 # Name every flagged leaf, not just the first: the user is being asked to make a
-                # decision, and they can only make it once if they can see the whole list.
-                detail = ", ".join(
-                    sorted(
-                        f"{_leaf_label(opening_items_by_id[oi_id])} ({qty} unit(s) awaiting replacement)"
-                        for oi_id, qty in flagged_leaves.items()
-                        if oi_id in opening_items_by_id
-                    )
-                )
+                # decision, and they can only make it once if they can see the whole list. Each leaf
+                # carries whichever of the two counts is non-zero, so "waiting will fix this" and
+                # "waiting will not" are distinguishable per leaf rather than lumped together.
+                labels = []
+                for oi_id in set(flagged_leaves) | set(never_pulled):
+                    oi = opening_items_by_id.get(oi_id)
+                    if oi is None:
+                        continue
+                    parts = []
+                    if oi_id in flagged_leaves:
+                        parts.append(f"{flagged_leaves[oi_id]} unit(s) awaiting replacement")
+                    if oi_id in never_pulled:
+                        parts.append(f"{never_pulled[oi_id]} unit(s) never pulled")
+                    labels.append(f"{_leaf_label(oi)} ({', '.join(parts)})")
+                detail = ", ".join(sorted(labels))
                 raise ValidationError(
-                    "These assembled leaves are incomplete - hardware on them is still awaiting a "
-                    f"replacement: {detail}. Confirm you want to ship them short, or wait for the "
-                    "replacements to be installed.",
+                    "These assembled leaves are incomplete - hardware they should carry is either "
+                    f"awaiting a replacement or was never pulled: {detail}. Confirm you want to ship "
+                    "them short, or wait for the hardware.",
                     field="opening_item_id",
                 )
 
@@ -1143,6 +1199,7 @@ def finalize_import_session(
                     "so it cannot be sent to shop assembly.",
                     field="shop_assembly_openings",
                 )
+            _validate_leaf_allocation(opening_number, sa_opening_input)
             opening_id_by_number[opening_number] = opening_id
             opening_leaf_specs.append((opening_number, opening_id, sa_opening_input.get("leaf")))
 
@@ -1194,12 +1251,21 @@ def finalize_import_session(
                 field="shop_assembly_openings",
             )
 
-        # Reservation gate: every checklist line claims fungible stock, so the whole request must fit
-        # inside what is available right now.
+        # Reservation gate, now over the ALLOCATION rather than the demand. The request no longer has
+        # to fit the schedule's full bill of hardware - the requester already decided what to claim
+        # and what to send short - so what is gated is exactly what will be reserved and pulled.
+        #
+        # `_gate_on_available_inventory` itself is unchanged and still refuses the whole finalize.
+        # It is not the shortfall gate any more; it is the **race** gate. The allocator built its
+        # numbers from a snapshot of availability, and if that snapshot went stale between load and
+        # send (another request took the stock, an admin wrote it off), reserving anyway would write
+        # a claim on hardware that is not there. The wizard catches the refusal, refetches
+        # availability, re-runs auto-assign and asks the user to look again.
         sar_needs = [
-            (item["hardware_category"], item["product_code"], item["quantity"])
+            (item["hardware_category"], item["product_code"], _allocated_quantity(item))
             for sa_opening_input in sar_openings_input
             for item in sa_opening_input.get("items", [])
+            if _allocated_quantity(item) > 0
         ]
         _gate_on_available_inventory(
             session,
@@ -1208,6 +1274,20 @@ def finalize_import_session(
             label="shop-assembly request",
             request_number=sar_request_number,
         )
+        # Everything the schedule owed but the allocation could not cover. Short combos are out of
+        # available stock by construction - the allocator only leaves a line short once the pool for
+        # that combo is exhausted - so this is always a genuine purchasing signal, not a claim
+        # somebody else is holding. It fires *after* the request is created, unlike the refusal path
+        # in app/schemas/imports.py, which stays for the race above.
+        sar_short_by_combo: dict[tuple[str, str], tuple[int, int]] = {}
+        for sa_opening_input in sar_openings_input:
+            for item in sa_opening_input.get("items", []):
+                short = item["quantity"] - _allocated_quantity(item)
+                if short <= 0:
+                    continue
+                key = (item["hardware_category"], item["product_code"])
+                owed, missing = sar_short_by_combo.get(key, (0, 0))
+                sar_short_by_combo[key] = (owed + item["quantity"], missing + short)
 
         sar = SARModel(
             id=uuid.uuid4(),
@@ -1247,12 +1327,17 @@ def finalize_import_session(
                         shop_assembly_opening_id=sa_opening.id,
                         hardware_category=item_input["hardware_category"],
                         product_code=item_input["product_code"],
+                        # Owed stays the schedule's number even when nothing could be claimed for it.
+                        # The checklist is what the leaf takes; the allocation is what turned up.
                         quantity=item_input["quantity"],
+                        allocated_quantity=_allocated_quantity(item_input),
                     )
                 )
 
         # The request holds its claim from here until the warehouse approves the pull that spends it
-        # (#342). One row per combo across the whole request, not per opening.
+        # (#342). One row per combo across the whole request, not per opening - and over the
+        # allocated quantities, so the pull the accept mints asks for exactly what is reserved and
+        # `approve_pull_request` keeps its all-or-nothing property with no warehouse-side change.
         warehouse_repository.create_reservations(
             session,
             project.id,
@@ -1260,6 +1345,26 @@ def finalize_import_session(
             sar.id,
             sar_needs,
         )
+
+        if sar_short_by_combo:
+            from app.services import notification_service
+
+            notification_service.notify_po_shortfall(
+                session,
+                project_id=project.id,
+                request_number=sar_request_number,
+                shortfalls=[
+                    warehouse_repository.Shortfall(
+                        hardware_category=cat,
+                        product_code=code,
+                        requested=owed,
+                        available=owed - missing,
+                        short=missing,
+                        reserved=0,
+                    )
+                    for (cat, code), (owed, missing) in sorted(sar_short_by_combo.items())
+                ],
+            )
 
     session.flush()
 
--- a/backend/app/repositories/import_repository.py
+++ b/backend/app/repositories/import_repository.py
@@ -967,28 +967,27 @@ def finalize_import_session(
         # short-shipping is a real workflow, it just has to be a decision someone made. Refused here
         # unless the caller says explicitly that it knows, and the wizard's confirm sets that flag.
         #
-        # Two grouped scalar aggregates for every referenced leaf, not a per-line count.
+        # One grouped scalar aggregate for every referenced leaf, not a per-line count and not one
+        # query per reading.
         if referenced_oi_ids and not acknowledge_incomplete_leaves:
             from app.repositories import shop_assembly_repository
 
-            oi_ids = list(referenced_oi_ids)
-            flagged_leaves = shop_assembly_repository.get_awaiting_replacement_quantities(session, oi_ids)
-            never_pulled = shop_assembly_repository.get_never_pulled_quantities(session, oi_ids)
-            if flagged_leaves or never_pulled:
+            shortfalls = shop_assembly_repository.get_leaf_shortfalls(session, list(referenced_oi_ids))
+            if shortfalls:
                 # Name every flagged leaf, not just the first: the user is being asked to make a
                 # decision, and they can only make it once if they can see the whole list. Each leaf
                 # carries whichever of the two counts is non-zero, so "waiting will fix this" and
                 # "waiting will not" are distinguishable per leaf rather than lumped together.
                 labels = []
-                for oi_id in set(flagged_leaves) | set(never_pulled):
+                for oi_id, shortfall in shortfalls.items():
                     oi = opening_items_by_id.get(oi_id)
                     if oi is None:
                         continue
                     parts = []
-                    if oi_id in flagged_leaves:
-                        parts.append(f"{flagged_leaves[oi_id]} unit(s) awaiting replacement")
-                    if oi_id in never_pulled:
-                        parts.append(f"{never_pulled[oi_id]} unit(s) never pulled")
+                    if shortfall.awaiting_replacement:
+                        parts.append(f"{shortfall.awaiting_replacement} unit(s) awaiting replacement")
+                    if shortfall.never_pulled:
+                        parts.append(f"{shortfall.never_pulled} unit(s) never pulled")
                     labels.append(f"{_leaf_label(oi)} ({', '.join(parts)})")
                 detail = ", ".join(sorted(labels))
                 raise ValidationError(
@@ -1279,15 +1278,21 @@ def finalize_import_session(
         # that combo is exhausted - so this is always a genuine purchasing signal, not a claim
         # somebody else is holding. It fires *after* the request is created, unlike the refusal path
         # in app/schemas/imports.py, which stays for the race above.
-        sar_short_by_combo: dict[tuple[str, str], tuple[int, int]] = {}
+        #
+        # Totals are accumulated over EVERY line of a combo, not only the short ones. Purchasing acts
+        # on the combo, so the number it needs is what this request wanted of that product against
+        # what it got; counting only the short lines would report a request that took 4 hinges on one
+        # leaf and missed 3 on another as "need 4, 1 available", which is true of neither.
+        sar_totals_by_combo: dict[tuple[str, str], list[int]] = {}
         for sa_opening_input in sar_openings_input:
             for item in sa_opening_input.get("items", []):
-                short = item["quantity"] - _allocated_quantity(item)
-                if short <= 0:
-                    continue
                 key = (item["hardware_category"], item["product_code"])
-                owed, missing = sar_short_by_combo.get(key, (0, 0))
-                sar_short_by_combo[key] = (owed + item["quantity"], missing + short)
+                totals = sar_totals_by_combo.setdefault(key, [0, 0])
+                totals[0] += item["quantity"]
+                totals[1] += _allocated_quantity(item)
+        sar_short_by_combo = {
+            key: (owed, allocated) for key, (owed, allocated) in sar_totals_by_combo.items() if owed > allocated
+        }
 
         sar = SARModel(
             id=uuid.uuid4(),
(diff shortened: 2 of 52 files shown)


backend/app/repositories/import_repository.py
```diff
@@ -911,34 +955,46 @@ def finalize_import_session(
...
         if referenced_oi_ids and not acknowledge_incomplete_leaves:
             from app.repositories import shop_assembly_repository
 
-            flagged_leaves = shop_assembly_repository.get_awaiting_replacement_quantities(
-                session, list(referenced_oi_ids)
-            )
-            if flagged_leaves:
+            oi_ids = list(referenced_oi_ids)
+            flagged_leaves = shop_assembly_repository.get_awaiting_replacement_quantities(session, oi_ids)
+            never_pulled = shop_assembly_repository.get_never_pulled_quantities(session, oi_ids)
```
Xilous: Never-pulled leaves become unshippable: the gate widened but the acknowledgment path did not.

`acknowledge_incomplete_leaves` is only ever set true by `ShippingPRsStep.tsx`, and its trigger is `leaf.awaitingReplacementQuantity > 0` (lines 107 and 265). That field comes from `get_awaiting_replacement_quantities` alone (`schemas/warehouse.py` opening_items / opening_item_details), so it is deficient + replacement_pending and carries nothing about never-pulled units.

Leaf assembled from a short request (owed 4, allocated 2, both installed, no deficiency):
- wizard sees `awaitingReplacementQuantity === 0` -> no chip, no confirm dialog, `onAcknowledgeIncompleteLeaf()` never fires
- finalize sends `acknowledgeIncompleteLeaves: false`
- this block finds it in `never_pulled` and raises ValidationError
- there is no UI affordance to set the flag, so the leaf can never be shipped

Either surface never-pulled on the OpeningItem type and fold it into the wizard's `incomplete` test, or scope this new refusal to leaves the wizard can already flag.

frontend/src/modules/shop-assembly/ShopAssemblyRequestsPage.tsx
```diff
@@ -88,9 +91,28 @@ export default function ShopAssemblyRequestsPage() {
...
         onChanged={refetch}
-        renderSummary={(req) => (
-          <Chip label={`${req.openings.length} opening(s)`} size="small" variant="outlined" />
-        )}
+        renderSummary={(req) => {
+          // The acceptor is approving a pull for the ALLOCATED quantities, so the short count has to
+          // be on the summary line, not buried in the per-leaf tables. Approving a request that is
+          // knowingly short is fine; approving one without knowing it is short is not.
+          const short = req.openings.reduce(
+            (sum, o) => sum + o.items.reduce((n, i) => n + (i.quantity - i.allocatedQuantity), 0),
```
Xilous, line 99: `allocatedQuantity` is never fetched here, so the whole accept-screen half of this slice is dead.

`GET_SHOP_ASSEMBLY_REQUESTS` (graphql/shop-assembly.ts, the `items { id hardwareCategory productCode quantity }` selection around line 247) was not updated. The hand-written `RequestOpeningItem` interface declares the field, so TypeScript does not catch it.

Runtime result on a short request:
- `i.allocatedQuantity` is `undefined` -> `short` is `NaN` -> `NaN > 0` is false -> the request-level short chip never renders
- the new "Allocated" cell renders empty and `item.quantity > undefined` is false, so the per-line short chip never renders

The acceptor approves a knowingly-short pull with nothing on screen saying so, which is exactly what the comment above says must not happen. Add `allocatedQuantity` to the items selection in `GET_SHOP_ASSEMBLY_REQUESTS`.

frontend/src/modules/shop-assembly/openingFilters.ts
```diff
@@ -18,32 +18,52 @@ export function isAvailableForAssignment(o: AssignableFields): boolean {
...
   const planned = items.reduce((sum, i) => sum + i.quantity, 0);
+  // Fall back to the owed quantity on a server that predates allocation, where every line was fully
+  // allocated by construction - the all-or-nothing gate could not create anything else.
+  const allocated = items.reduce((sum, i) => sum + (i.allocatedQuantity ?? i.quantity), 0);
   const installed = items.reduce((sum, i) => sum + i.installedQuantity, 0);
   const deficient = items.reduce((sum, i) => sum + i.deficientQuantity, 0);
   const replacementPending = items.reduce((sum, i) => sum + (i.replacementPendingQuantity ?? 0), 0);
-  const remaining = planned - installed - deficient - replacementPending;
-  return { planned, installed, deficient, replacementPending, remaining, complete: remaining === 0 };
+  const remaining = allocated - installed - deficient - replacementPending;
```
Xilous, line 57: `remaining` moved to the allocated basis but all three consumers still pair it with `planned`, so the unit counters now report work that never happened.

Every caller computes `planned - remaining` over `planned`:
- `AssembleListPage.tsx:277` -> `{progress.planned - progress.remaining}/{progress.planned} units`
- `MyWorkPage.tsx:81` -> `` `${planned - remaining}/${planned} units` ``
- `AssemblyDetailModal.tsx:350` -> `${draftProgress.planned - draftProgress.remaining}/${draftProgress.planned} units accounted for`

Owed 4, allocated 2, nothing installed: remaining = 2, so the label reads "2/4 units" with zero units dispositioned. Owed 4, allocated 2, both installed: remaining = 0, label reads "4/4 units" for a leaf carrying 2. The short units are silently counted as done.

`pipelineStages.ts::unitProgressLabel` was updated correctly (`done/allocatedUnitCount`); these three were not. They should be `allocated - remaining` over `allocated`, with `short` shown separately.

frontend/src/modules/import/ShopAssemblyStep.tsx
```diff
@@ -50,52 +68,90 @@ export default function ShopAssemblyStep({
...
+
+  // Auto-assign runs **once**, when the step first has real numbers to work with, and never again on
+  // its own. Re-running silently would throw away every manual move the moment an unrelated query
+  // refetched; the user re-runs it deliberately with the button, and a race refusal re-runs it while
+  // saying so. Waiting for the lookup to answer matters too - an empty availability map reads as
+  // "nothing available" and would allocate nothing to everything.
+  const seeded = useRef(false);
+  useEffect(() => {
+    if (seeded.current || availabilityLoading || availabilityError || openingDrafts.length === 0) return;
+    seeded.current = true;
```
Xilous: `seeded` lives in the step, which unmounts on every step change, so Back then Next silently discards every manual allocation.

`ImportWizard.tsx:1491` renders this as `{effectiveStepId === 'shop-assembly' && <ShopAssemblyStep ... />}`. Leaving the step unmounts the component and resets `seeded.current` to false. Coming back mounts it fresh, availability is cached so `availabilityLoading` is already false, the effect fires and `runAutoAssign()` overwrites `sarAllocation` and `includedLeafKeys`.

Repro: allocator step -> move hardware between leaves, exclude a leaf -> Back -> Next. Every stepper is back at the auto-assigned value and the excluded leaf is included again.

The state was deliberately lifted to the wizard for exactly this ("Held here rather than inside the step so stepping back and forward does not silently re-run auto-assign over the user's manual moves") but the seeded flag was not. Lift it too, or key the seed off `allocation.size === 0`.

frontend/src/modules/import/allocation.ts
```diff
@@ -0,0 +1,248 @@
...
+        rows.set(key, row);
+      }
+      row.owed += item.quantity;
+      row.allocated += allocatedFor(allocation, draft, item);
+    }
+  }
+  const result = Array.from(rows.values());
+  for (const row of result) {
+    row.remaining = Math.max(0, row.available - row.allocated);
+    row.short = Math.max(0, row.owed - row.allocated);
```
Xilous, line 248: `comboSummary` and `buildAllocatedDrafts` disagree about which leaves are in the request, so the short totals over-report what is actually sent.

`buildAllocatedDrafts` filters on `includedLeafKeys.has(...) && leafAllocatedTotal > 0`. `comboSummary` filters on `includedLeafKeys.has(...)` alone.

A leaf the user manually steppered down to 0 on every line stays in `includedLeafKeys` (`leafCoverage` turns NONE, the render greys it out and disables its toggle, but nothing removes the key). That leaf then:
- contributes its full `owed` to `row.owed` here, so `row.short` includes it
- drives `totalShort` and the "This request will be sent N unit(s) short" alert in `ShopAssemblyStep.tsx:252`
- is dropped from the payload by `buildAllocatedDrafts`, so the server's `sar_short_by_combo` and the PO notification never see it

The UI short count is larger than the one the request actually records. Apply the same `leafAllocatedTotal > 0` predicate in `comboSummary`, or drop the key when a leaf's coverage falls to NONE.

backend/app/repositories/warehouse/pull_requests.py
```diff
@@ -1283,6 +1325,7 @@ def discard_pending_pull_request(session: Session, pr_id: uuid.UUID | None) -> N
             f"Cannot reopen - the warehouse has already started this pull request ({pr.status.value}). "
             "Resolve it in the warehouse first."
         )
+    reservations.release_reservations(session, ReservationSource.REPLACEMENT_PULL, pr.id)
```
Xilous, line 1331: This release is unreachable, and it is the only release path a PENDING replacement claim has.

`discard_pending_pull_request` has two callers:
- `shop_assembly_repository.py:1271` (reopen a SAR, passes the SAR's own pull)
- `shipping_repository.py:769` (reopen a SOR, passes that request's pull)

Neither can ever be a PR-REPL pull. And `cancel_pull_request` refuses a PENDING pull outright ("This pull has not been approved yet..."), there is no soft-delete mutation, and `_handle_schedule_replacement` only rewrites PENDING SARs while a flagged leaf necessarily sits on an APPROVED one.

So a claim minted at flag time is held until somebody approves that pull, with no escape. A replacement that is never approved (leaf shipped short, request abandoned, deficiency handled off-system) permanently removes that stock from project availability, and this is the first version where a PR-REPL pull can do that.

`test_discarding_a_pending_replacement_pull_releases_its_claim` covers a path production cannot reach. Either give the warehouse a real way to abandon a PENDING replacement pull that routes here, or state where the claim is expected to be released.

backend/app/repositories/warehouse/reservations.py
```diff
@@ -253,3 +311,120 @@ def release_reservations(session: Session, source: ReservationSource, request_id
...
+                    InventoryReservationModel.source == ReservationSource.REPLACEMENT_PULL,
+                    InventoryReservationModel.pull_request_id == pr.id,
+                )
+            ).all()
+        }
+        for combo, need in sorted(needs.items()):
+            gap = need - held.get(combo, 0)
+            if gap <= 0:
+                continue
+            got = reserve_for_replacement_pull(session, pr.id, project_id, combo[0], combo[1], gap)
```
Xilous: Per-(pull, combo) query loop with a FOR UPDATE inside each iteration, on the receive path.

Each `reserve_for_replacement_pull` call runs `get_available_quantities(lock=True)`, which is a `SELECT ... FOR UPDATE` over `inventory_locations` plus a grouped aggregate over `inventory_reservations`, then a third SELECT for the existing reservation row and a `flush()`. That is four round trips per (pull, combo), and the outer loop already did one query per pull for `held`.

`notify_unblocked_replacement_pulls` calls this on every non-stock PO receive, then loops the same pulls again for `check_inventory_sufficiency`. A project with several pending replacements across several received combos turns one receive into dozens of sequential Railway round trips, several of them holding row locks for the rest of the transaction. CLAUDE.md's perf rules call this shape out specifically.

One `get_available_quantities` for the whole combo set up front, decremented in Python as claims are handed out, would collapse it to two queries plus the writes.

backend/app/repositories/import_repository.py
```diff
@@ -1261,6 +1346,26 @@ def finalize_import_session(
             sar_needs,
         )
 
+        if sar_short_by_combo:
+            from app.services import notification_service
+
+            notification_service.notify_po_shortfall(
```
Xilous, line 1357: Two problems with reusing `notify_po_shortfall` on the success path.

Wording: it renders `"Pull Request {request_number} couldn't be fulfilled - backfill needed."` (notification_service.py:105-106). The request was created and is proceeding, and there is no pull yet. The PO reads a live short request as a failed one.

Numbers: `sar_short_by_combo` only accumulates lines where `short > 0`, so `requested` is the owed total of the short lines alone and `available` is `owed - missing`, which is those lines' allocation rather than project availability. A request with HG-100 owed 4/allocated 4 on A01 and owed 4/allocated 1 on A02 reports "need 4, 1 available (short 3)" while the request actually claimed 5 of that combo. `test_short_across_two_leaves_is_one_notification_per_combo` asserts the short number and never the requested/available pair, so this is unpinned.

A distinct message (created-short vs could-not-fulfil) with the combo's real owed total would say what happened.

backend/app/models/shop_assembly.py
```diff
@@ -156,15 +159,30 @@ class ShopAssemblyOpeningItem(Base):
...
+    #   allocated_quantity - what was reserved, pulled and will physically arrive on the cart.
+    # **Short is derived, never stored**: short = quantity - allocated_quantity. There is no separate
+    # state for it and nothing downstream may write one, because the authority on what is still
+    # missing is the *current* schedule compared against what is physically on the leaf - the short
+    # count here is evidence of what this request executed, not a backlog anybody works off.
+    # Backfilling a short leaf when stock turns up later belongs to the reallocation module.
+    #
+    # 0 is legal: a line can be fully short on a leaf that other lines do cover. A leaf with *every*
+    # line at 0 has an empty cart and is dropped before it reaches here.
+    allocated_quantity: Mapped[int] = mapped_column(Integer, nullable=False, default=0, server_default="0")
```
Xilous: The model keeps `server_default="0"` that migration 070 deliberately drops, and `default=0` re-opens the footgun the migration says it closed.

070 does `alter_column(..., server_default=None)` with the rationale "the default is dropped immediately, so nothing later can insert a silently-zero row and have it read as fully short". This declaration contradicts both halves of that:
- `server_default="0"` no longer matches the DB. `compare_server_default` is off in `alembic/env.py` so `alembic check` will not catch it, but the next autogenerate against a fresh DB will re-add the default.
- `default=0` is a client-side default, so any future ORM insert that omits `allocated_quantity` writes 0 and the row reads as fully short with the constraint satisfied. That is exactly the silent-zero case, just via SQLAlchemy instead of the DB.

Drop both defaults and let the single construction site in `import_repository.py:1330` be the only writer, or keep them and drop the claim in the migration docstring.

frontend/src/modules/import/ImportWizard.tsx
```diff
@@ -1035,15 +1080,41 @@ export default function ImportWizard({ open, project, onClose }: ImportWizardPro
...
+        const fresh = new Map<string, number>();
+        for (const row of refreshed?.data?.projectInventoryAvailability ?? []) {
+          fresh.set(
+            itemGroupKey({ hardware_category: row.hardwareCategory, product_code: row.productCode }),
+            row.availableQuantity,
+          );
+        }
+        const next = autoAssign(shopAssemblyOpeningDrafts, fresh);
+        setSarAllocation(next);
+        setIncludedLeafKeys(
```
Xilous: The race recovery throws away the user's exclusions, and degrades badly when the refetch fails.

`includedLeafKeys` is rebuilt purely from `leafCoverage(next, draft) !== 'NONE'`, so every leaf the user deliberately toggled off comes back into the request. A race in availability is not a reason to re-include a leaf somebody chose to hold back; intersect with the previous `includedLeafKeys` instead.

Separately, `refetchAvailability().catch(() => null)` on line 1097 leaves `refreshed` null, `fresh` empty, and `autoAssign` then allocates 0 to everything and `includedLeafKeys` becomes empty. The step remount happens to re-seed from the cached `availabilityByCombo` and paper over it, which means the recovery is correct only by accident. Bail out of the rebuild when the refetch did not return data.

`allocationStale` also never resets on edit, only on a successful finalize, so the warning banner outlives the condition.

backend/app/repositories/warehouse/reservations.py
```diff
@@ -253,3 +311,120 @@ def release_reservations(session: Session, source: ReservationSource, request_id
...
+                hardware_category,
+                product_code,
+                claim,
+            )
+        )
+    session.flush()
+    return claim
+
+
+def top_up_replacement_reservations(
```
Xilous, line 375: Only two of the paths that raise project availability call this, so the "first refusal" guarantee is partial.

Wired up: PO receive (via `notify_unblocked_replacement_pulls`) and `resolve_deficiency` REPAIR. Not wired up, and each one frees project stock a waiting replacement should get first:
- `cancel_pull_request` restocking a cancelled pull's quantities
- `reject_shop_assembly_request` / `_handle_schedule_replacement` releasing a request's reservations
- allocating stock-pool inventory to a project

On any of those, a request created after the defect can still take the units the replacement is waiting on, which is the exact failure the docstring says this exists to end. Worth either hooking them or naming them as known gaps here, since the omission is invisible from the two call sites that are wired.

backend/app/repositories/shop_assembly_repository.py
```diff
@@ -691,22 +700,75 @@ def get_awaiting_replacement_quantities(
         .limit(1)
         .lateral("latest_assembly")
     )
-    outstanding = func.sum(
-        ShopAssemblyOpeningItem.deficient_quantity + ShopAssemblyOpeningItem.replacement_pending_quantity
-    )
+
+
+def _sum_over_source_checklist(session: Session, opening_item_ids: list[uuid.UUID], expression):
```
Xilous: The shipping guard now runs this lateral twice over the same id set.

`import_repository.py:975-976` calls `get_awaiting_replacement_quantities` then `get_never_pulled_quantities` back to back with the identical `oi_ids`. Each rebuilds `_latest_assembly_lateral()` and re-executes the correlated LATERAL join over every referenced leaf. The docstring above justifies one grouped aggregate on the grounds that this list is "the heaviest list in the module"; the same argument says it should not be run twice.

Two `func.sum` expressions in one select, returning both totals per leaf, keeps the callers' shape and halves the work. `_sum_over_source_checklist` could take a mapping of name to expression, or gain a sibling that returns both.

frontend/src/test/setup.ts
```diff
@@ -1 +1,27 @@
...
+ * internal "already suggested" flag without scheduling anything. Every real client built afterwards
+ * - including the ones `MockedProvider` builds internally, which take no devtools option - finds
+ * that flag already set.
+ *
+ * `window.top` is restored immediately, so nothing under test sees the frame-like window.
+ */
+const realTop = window.top;
+Object.defineProperty(window, 'top', { configurable: true, value: null });
+void new ApolloClient({ cache: new InMemoryCache(), link: ApolloLink.empty() });
+Object.defineProperty(window, 'top', { configurable: true, value: realTop });
```
Xilous: The restore does not restore: `Object.defineProperty` with only `value` and `configurable` defaults `writable: false` and `enumerable: false`.

`window.top` starts as an accessor on jsdom's Window. After this line it is a non-writable, non-enumerable data property for the rest of every test file. Anything under test that assigns `window.top`, or enumerates window, now behaves differently than in a browser. Add `writable: true, enumerable: true` at minimum.

Also worth noting this whole trick depends on an unversioned Apollo heuristic (it only arms the suggestion timer when the page looks like a top-level document). If Apollo changes that check the flag stops being set, nothing fails loudly, and the flaky 10-second unhandled ReferenceError comes back. `vi.stubGlobal('setTimeout', ...)` or a `fakeTimers` guard would fail visibly instead.

frontend/src/modules/import/ImportWizard.tsx
```diff
@@ -586,16 +614,28 @@ export default function ImportWizard({ open, project, onClose }: ImportWizardPro
...
+  // (docs/HARDWARE_IDENTITY_LIFECYCLE.md), so it reserves nothing.
   const requestedByCombo = useMemo(() => {
     const map = new Map<string, number>();
     if (purpose === 'assembly') {
-      for (const draft of shopAssemblyOpeningDrafts) {
+      for (const draft of allocatedShopAssemblyDrafts) {
         for (const item of draft.items) {
           const key = itemGroupKey({ hardware_category: item.hardwareCategory, product_code: item.productCode });
-          map.set(key, (map.get(key) ?? 0) + item.quantity);
+          map.set(key, (map.get(key) ?? 0) + item.allocatedQuantity);
```
Xilous: The assembly branch of `requestedByCombo` is now dead work.

`requestedByCombo` feeds only `availabilityShortfalls`, and after this PR `availabilityShortfalls` has exactly one consumer left: `ShippingPRsStep` (line 1522). `ShopAssemblyStep` computes its own totals from `comboSummary`. So for `purpose === 'assembly'` this loop runs on every allocation keystroke and its output is never read.

Narrowing `requestedByCombo` to the shipping branch removes a map rebuild per stepper click and removes the second derivation of the allocated totals, which is the drift risk the comment above `shopAssemblyOpeningDrafts` warns about.

Xilous (commented):
14 inline findings. Four are blocking.

1. never-pulled leaves are unshippable - the shipping gate widened, the acknowledge path did not
2. `GET_SHOP_ASSEMBLY_REQUESTS` never fetches `allocatedQuantity`, so the accept screen's short chip and Allocated column are dead
3. `assemblyProgress.remaining` moved to the allocated basis but three consumers still divide by `planned`, so short units read as work done
4. `seeded` ref lives in `ShopAssemblyStep`, which unmounts on step change - Back then Next silently wipes the user's allocation

The rest: UI/payload disagreement on which leaves count as short, an unreachable release path that leaves a PENDING replacement claim with no escape, a per-(pull, combo) FOR UPDATE loop on the receive path, PO shortfall wording/numbers on the success path, model defaults contradicting migration 070, race recovery re-including excluded leaves, top-up gaps on the other availability-raising paths, a doubled lateral in the shipping guard, a non-restoring `window.top` restore, and a dead `requestedByCombo` branch.

Xilous:
merge and test this after pr 379 is merged

Issue #367, opened 2026-07-27
pull requests - flesh out the workflow


Xilous:
shop assembly requests or shipping out requests, once accepted, trigger the creation of a pull request for them.

need to build out a proper user experience for the warehouse user user to perform the actual pull request.

all ears on design for this user experience, but current working idea so far is:

expanded list of 1 section per product code. each section within it contains all of the warehouse locations that contains that product code.

requirement is that this form can also be printed out on paper, so it needs to have a pdf output.

within each product code unique section, the user inputs on each location how much quantity of that product code they pulled from each location.

Xilous:
warehouse user dictates how much and from which location hardware items are deducted from

Xilous:
deduction moment moves approve -> explicit pick confirmation, dictated by the warehouse user, per location, per product code, with a printable paper sheet.

#342 invariant relocated to confirm-pick, source request's reservation consumed atomically at the moment of deduction.

decisions already settled
- no system suggestions anywhere. no suggested split button, no suggested column, not on screen and not on the PDF. Received date column stays so pickers can rotate stock themselves.
- all door leaves listed in full in every section. "+N more" is banned. identity is leaf-of-opening everywhere, e.g. 101A · L1, consistent with #311.
- stage carts view becomes dedicated sections per opening door leaf following the pick-screen pattern, hardware item rows beneath each section. #343 staging semantics otherwise unchanged.
- fetch check-offs on shipping pulls persist per leaf via fetched_at/fetched_by.
- short confirm allowed, deduct what was entered, remainder stays open, purchasing notified, integrity event if the pull held a reservation. a second confirm covers the remainder.
- no over-pull ever. staging and complete gated on pick confirmed.
- a local mock will be used as reference for implementation

backend

1. migration + models, one alembic revision

new table pull_pick_lines
id
pull_request_id FK indexed
hardware_category
product_code
inventory_location_id FK nullable ondelete=SET NULL
quantity check > 0
state enum pull_pick_line_state (DRAFT | APPLIED)
entered_by
entered_at
applied_at nullable

pull_requests gains picked_at nullable dt, picked_by nullable str.
pull_request_items gains fetched_at nullable dt, fetched_by nullable str for OPENING_ITEM fetch check-offs.
backfill UPDATE pull_requests SET picked_at = approved_at, picked_by = assigned_to WHERE approved_at IS NOT NULL, so every pull approved under the old model (already deducted) never presents a pick sheet and staging keeps working.
[...]

Issue #342, opened 2026-07-26
Inventory reservations: request creation reserves stock (shop assembly + shipping out)


Xilous:
Part 4 of 6. Architectural decision from the product owner: **creating a shop-assembly or shipping-out request RESERVES inventory.** The creator must abide by currently-available counts at creation time — if stock can't cover the selection, they refine the selection of openings. The downstream sufficiency checks stop being reactive point-in-time re-checks and become reservation consumption. This kills the accept-vs-pull TOCTOU race and the "blocked pull waiting on backfill" limbo for normal requests.

## Reservation model

- New `inventory_reservations` table (Alembic migration): `project_id`, `hardware_category`, `product_code`, `quantity`, source request discriminator + id (shop-assembly request | shipping-out request), `created_at`. Aggregate-level, NOT pinned to specific `InventoryLocation` rows (FIFO deduction stays unchanged at consumption time). InventoryLocation itself stays untouched — a hinge is still a hinge (docs/HARDWARE_IDENTITY_LIFECYCLE.md).
- **Available = on-hand − deficient − active reservations.** `check_inventory_sufficiency` (backend/app/repositories/warehouse/) grows a reservation-aware mode; every caller is audited and updated.

## Lifecycle

- **Create** (Start a Task, `import_repository` finalize, both request types): gate on available (reservation-aware) inventory — raise a typed shortfall error listing the short combos if insufficient — then write reservations in the same transaction. LOOSE shipping-out lines reserve; OPENING_ITEM lines do not (assembled leaves left fungible inventory at assembly).
- **Accept**: drop the reactive sufficiency re-check in `accept_shop_assembly_request` / `accept_shipping_out_request` — accept becomes a pure human approval gate that mints the PullRequest.
- **Pull approval** (`approve_pull_request`): consume the source request's reservations (delete/mark-consumed) and deduct FIFO as today, atomically. PR-REPL replacement pulls have no reservations (deficiencies can't be reserved in advance) — they keep the reactive check + PO-backfill-notification loop. Keep a defensive sufficiency check on the reserved path but treat failure as an integrity error, not normal flow.
- **Release** on: reject, reopen (#325), request/pull cancellation, and schedule re-upload handling (below).

## Also in scope (request-integrity guards)

[...]

Issue #344, opened 2026-07-26
Shop assembly slice 6: pipeline observability, notifications, housekeeping


Xilous:
Part 6 of 6. Depends on slices 1-5 (surfaces the states they introduce).

## Pipeline observability

- A per-request detail view answering "where is opening A01 leaf 2?": requested -> accepted -> pull staged x/y (or blocked) -> assigned (who) -> in progress (5/8 units) -> completed (location) -> shipped. Derived from existing state; scalar aggregate queries only (CLAUDE.md perf rules — this walks thousands of rows in the All-Projects view).
- Surface the flags earlier slices introduced: schedule-changed-since-request, awaiting-replacement, replacement-arrived-after-ship.

## Notifications (existing notification_service patterns)

- Pull opening staged / pull completed -> shop assembly manager (board has work).
- Replacement arrived -> the leaf's assignee.
- PR-REPL blocked-pull unblocked by receiving -> warehouse (the backfill retry loop for replacement pulls, which have no reservations).

## Housekeeping

- Delete any remaining dead enum values / states after slices 1-5 (audit PullStatus, PullRequestStatus, AssemblyStatus usage).
- Reconcile all backend/app/graphql/*.graphql SDL reference docs with the actual Strawberry schema for touched domains.
- refetch.ts disjoint refetch/evict lists extended for every new mutation.
- docs/HARDWARE_IDENTITY_LIFECYCLE.md enforcement table updated with the new code paths; testing/CLAUDE.md updated with the full new flow map.

PR #388, opened 2026-07-28
User-dictated per-location pull picking (#367)

Xilous:
issue #367. deduction moves out of approve into an explicit confirm-pick the warehouse user dictates, per location, per product code, against a printable paper sheet. #342 invariant relocated, the source request's reservation is consumed atomically at deduction, now confirm-pick not approve.

- start pick, PENDING -> IN_PROGRESS, pull assigned to the picker. no inventory movement, no sufficiency gate.
- save draft writes DRAFT `pull_pick_lines`, replace-all not merge, availability not validated. a draft is a note, not a claim.
- confirm pick consumes the claim per combo -> deducts the exact rows named. one APPLIED pick line + one PULL_DEDUCTION audit per row. stamps `picked_at` / `picked_by` once every combo is covered.
- short confirm deducts what was entered, pull stays In Progress and un-picked, purchasing notified once, remainder keyed in later.
- fetch check-offs on OPENING_ITEM lines persist per leaf.
- confirm pick enforces three ceilings:
the row's own available units
what the pull asked for
what is free once other requests' reservations are counted, this pull's own holder excluded for self-coverage
- the third one is a CONFLICT, and it is what keeps #342 intact. a claim that holds only until somebody physically reaches the shelf first is not a claim.
- `get_pick_sheet` exposes the same number as `claimableQuantity` / `claimableShortfall`. PickSection and the printed sheet warn about contention before the walk, not after it.

migration 072 adds:
`pull_pick_lines` (DRAFT/APPLIED, `inventory_location_id` ON DELETE SET NULL)
`pull_requests.picked_at` / `picked_by`
`pull_request_items.fetched_at` / `fetched_by`
backfill sets `picked_at = approved_at`, so nothing in flight at deploy time comes back presenting a pick sheet for hardware already on a cart

these replace `approve_pull_request`:
`start_pull_request_pick`
`get_pick_sheet`
`save_pick_draft`
`confirm_pick`
`set_pull_item_fetched`

- `reservations.consume_reservations` does partial per-combo consumption, deletes rows that hit zero, ignores overshoot.
- `stage_pull_openings` and `complete_pull_request` gate on `picked_at`.
- cancel restocks to the exact rows the units came off, reading the APPLIED pick lines. per-combo fallback for legacy pulls and null FKs. releases any surviving claim before re-creating the request's, since a pull cancelled before its pick still holds all of it.
- `pullPickSheet` query plus four mutations, all behind `require_user`.
- `approvePullRequest` and `ApproveResult` removed. `PickOutcome` (PICKED / SHORT) replaces `ApproveOutcome`.
- `get_pick_sheet` is a fixed five queries whatever the pull covers, `partiallyPicked` is one grouped query for the whole queue page.

- new route `pull-requests/:id/pick`.
- PickSection renders one section per product code, every leaf listed in full, per-location ledger rows with Location / Received / Available / Pulled input. no suggested split, no suggested column, no autofill. Received is there so the picker can rotate stock themselves.
- gauges for codes / required / entered / remaining, over and short alerts, confirm behind a dialog. confirm enabled only when balanced, short variant when something is entered and nothing is over, disabled when any row is over.
- PDF pick sheet with blank write-in boxes, full leaf lists, fetch list, Picked by / Date / Keyed into Nexus by signature footer.
- queue Phase cell replaces the Status and Staging columns -> Pending / Picking / Short / Staging n of m / Completed.
- detail modal offers Start pick and Resume pick. approve button, confirm dialog, availability lookup, Available Qty / Status columns removed.
- stage carts becomes a section per opening door leaf, hardware as ledger rows beneath.
- layout follows the local ui mock.

one thing to look at in review:
- no sufficiency gate at approve any more. an unfillable pull is picked short, raising the same PO backfill signal. the check moved into confirm pick, where it runs against what was actually entered.

- `backend/tests/test_pick_flow.py` covers start, draft semantics, exact-row deduction with audits and reservation consumption, all three over-pull refusals, a plain shortage not being reported as contention, the short path with integrity event, second confirm, the staging and completion gates, pure fetch pulls, fetch toggles, all three cancel restock shapes.
- `tests/pick_helpers.pick_pull` reproduces the old oldest-row-first behaviour, existing `approve_pull_request` call sites sweep mechanically and assert the same inventory outcome.
- `PickSheet.test.tsx` covers the entry arithmetic and PickSection validation states, staging and modal tests updated.
- backend tests not run locally, no Postgres on this machine. CI covers the DB-backed suite and migration integrity.

these pass locally:
ruff check
ruff format
tsc
eslint
vite build
323-test vitest suite


backend/app/repositories/warehouse/pull_requests.py
```diff
@@ -263,171 +270,879 @@ def find_reservation_holder(session: Session, pr: PullRequestModel) -> tuple[Res
...
-                        "newQuantity": row.quantity,
-                        "deducted": deduct,
-                        "pullRequestNumber": pr.request_number,
-                        "hardwareCategory": cat,
-                        "productCode": code,
-                    },
-                )
+    return ConfirmPickResult(
+        pull_request=pr,
+        outcome="SHORT",
```
Xilous, line 1133: short confirm leaves the pull IN_PROGRESS forever, out of the #344 backfill loop.

`find_pending_replacement_pulls` (receiving.py:347) filters `status == PENDING`
`notify_unblocked_replacement_pulls` and `top_up_replacement_reservations` are both built on it

before #367 an insufficient approve left the pull PENDING, so the backfill topped it up and raised a PULL_UNBLOCKED bell.

now: replacement pull short-picks 1 of 4 -> IN_PROGRESS -> PO backfills 3 hinges -> nothing tops up its claim, no notification fires. the received units are free for the next request to reserve, and the picker is never told the remainder is available.

frontend/src/modules/warehouse/PickPage.tsx
```diff
@@ -0,0 +1,354 @@
...
+        <PickSection
+          key={`${section.hardwareCategory}|${section.productCode}`}
+          section={section}
+          entries={entries}
+          onChange={handleEntryChange}
+          editable={isOpen && !confirming}
+        />
+      ))}
+
+      <FetchListPanel items={fetchItems} editable={isOpen} onChanged={() => refetch()} />
```
Xilous: `editable={isOpen}`, `isOpen = IN_PROGRESS && !pickedAt` - every fetch check-off goes read-only the moment the pick is confirmed.

a pure fetch pull (shipping out assembled leaves) enables confirm from first render with zero sections, and `onCompleted` navigates back to the queue on PICKED. confirm before ticking, or leave one leaf unticked, and there is no route back. `set_pull_item_fetched` only requires IN_PROGRESS, but no UI reaches it. same on a mixed pull whose loose lines finish first.

frontend/src/modules/warehouse/pick.ts
```diff
@@ -0,0 +1,195 @@
...
+    if (totals.over || totals.anyRowOver) over = true;
+  }
+  return {
+    codeCount: sections.length,
+    required,
+    applied,
+    entered,
+    remaining,
+    over,
+    balanced: remaining === 0 && !over,
```
Xilous, line 166: `sectionTotals` / `pickTotals` never read `claimableQuantity`, so `balanced` and `over` come off the row ceiling and the request ceiling only.

section needs 10, rows hold 10, another request reserved 4 -> `claimableQuantity` 6 -> UI reads balanced -> Confirm pick enabled -> `confirm_pick` throws ConflictError on the whole submission. the picker loses every entry across every section, including the ones that were fine, and re-keys.

`PickSection` renders the contention line, nothing gates on it.

backend/app/repositories/warehouse/pull_requests.py
```diff
@@ -263,171 +270,879 @@ def find_reservation_holder(session: Session, pr: PullRequestModel) -> tuple[Res
...
+#   start_pull_request_pick   the pull is claimed and opened for picking. Nothing moves.
+#   confirm_pick              the picker dictates a quantity per location, and *that* deducts.
+#
+# The #342 invariant is preserved, just relocated: the source request's claim is consumed at the
+# exact moment of the deduction, atomically, under the same row locks - which is now confirm rather
+# than approve. What changes is that the consumption can be partial, because a pick can be
+# confirmed short and the un-picked remainder is still owed.
+
+
+def start_pull_request_pick(session: Session, pr_id: uuid.UUID, started_by: str) -> PullRequestModel:
```
Xilous, line 290: no `deleted_at` check here, `_pickable_pull` has one. `lock_rows` does not filter soft-deleted rows.

soft-deleted PENDING pull -> started -> status IN_PROGRESS -> `session.commit()` -> resolver's `get_pull_request_details` (which does filter `deleted_at`) raises NotFoundError. the caller sees an error, the write is already committed, and the pull is stuck IN_PROGRESS with no way to pick, complete or cancel it.

backend/app/repositories/warehouse/pull_requests.py
```diff
@@ -263,171 +270,879 @@ def find_reservation_holder(session: Session, pr: PullRequestModel) -> tuple[Res
...
+            PullPickLineModel.state == PullPickLineState.DRAFT,
+        )
     )
+    session.flush()
 
-    # 4. If short: leave the PR PENDING (blocked, not cancelled), keep the reservations (the claim
-    #    outlives the blocked attempt), and notify the PO for backfill.
-    if not result.sufficient:
+    covered = {combo: already.get(combo, 0) + entered_by_combo.get(combo, 0) for combo in required}
+    short_combos = {combo: required[combo] - picked for combo, picked in covered.items() if picked < required[combo]}
```
Xilous, line 1072: `confirm_pick` with `lines: []` on a pull that has LOOSE requirements deducts nothing, returns SHORT, and fires an INVENTORY_SHORTFALL notification at purchasing for the full requirement.

`canConfirm` needs `balanced || entered > 0`, but the mutation is directly reachable. the dedupe on `has_unread_notification_for_pull` then suppresses the real shortfall signal from the later genuine short confirm until somebody reads the bogus one.

backend/app/repositories/warehouse/pull_requests.py
```diff
@@ -263,171 +270,879 @@ def find_reservation_holder(session: Session, pr: PullRequestModel) -> tuple[Res
...
+        # Only speak of contention when reservations are actually the cause. A combo can also come
+        # up short here simply because the project does not hold that much, and saying "another
+        # request has claimed this" with `reserved = 0` would send somebody hunting for a competing
+        # request that does not exist. That case falls through to the per-row check below, which
+        # names the bin and the number - the more useful answer, and the one the picker can act on.
+        #
+        # Falling through is safe rather than a hole: with no reservations, free is exactly on-hand,
+        # so entering more than the combo has means at least one row was entered past its own
+        # available units, which is precisely what that check catches.
+        contested = [s for s in contention.shortfalls if s.reserved > 0]
```
Xilous: `contested = [s for s in contention.shortfalls if s.reserved > 0]` still blames contention when the shortage is mostly plain scarcity.

on-hand 5, another request reserved 1, picker enters 10 -> available 4, reserved 1 -> "Another request has already claimed this stock ... 10 entered but only 4 free (1 claimed by other requests). The units are on the shelf but they are spoken for." five of those units do not exist at all.

a35832f fixed the `reserved == 0` case, the mixed case still sends the picker hunting for a competing request that accounts for 1 unit.

frontend/src/modules/warehouse/PickSheetDocument.tsx
```diff
@@ -0,0 +1,202 @@
...
+            Printed by {printedBy} on {printedAt}
+          </Text>
+        </View>
+
+        {sections.length === 0 && fetchItems.length === 0 && (
+          <Text style={styles.empty}>This pull has nothing to pick.</Text>
+        )}
+
+        {sections.map((section) => (
+          <View key={`${section.hardwareCategory}|${section.productCode}`} wrap={false}>
```
Xilous: `wrap={false}` on the per-section View, and again on the fetch list View at line 152. react-pdf cannot split a `wrap={false}` block, so anything taller than one A4 page is clipped rather than continued.

a section with 40 leaves and 20 candidate locations, or a shipping pull with 30 assembled leaves, silently loses the overflow. breaks the "every leaf, never truncated" rule, on the paper the pick actually happens on.

frontend/src/modules/warehouse/PullRequestDetailModal.tsx
```diff
@@ -366,19 +283,33 @@ export default function PullRequestDetailModal({
...
-        onClick={() => setConfirmApproveOpen(true)}
-        disabled={approveLoading || !allLooseSufficient}
+        onClick={() => startPick({ variables: { id: pr.id, startedBy: displayName } })}
+        disabled={startLoading}
       >
-        {approveLoading ? 'Approving...' : 'Approve and Start'}
+        {startLoading ? 'Starting...' : 'Start pick'}
       </Button>
     );
+  } else if (isInProgress && !isPicked) {
```
Xilous: this branch runs before `isAssignedToCurrentUser`, so an IN_PROGRESS un-picked pull offers Resume pick to every user regardless of `assignedTo`. the modal renders the "Locked to {assignedTo}" banner at the same time.

before this PR such a pull fell through to `canCancel` and got no confirming action. `confirm_pick` does not check assignment either - two pickers can work the same sheet.

backend/app/schemas/warehouse.py
```diff
@@ -313,17 +331,56 @@ def pull_requests(
...
+
+        A query of its own rather than a field on `PullRequest`, for the same reason
+        `pullRequestOpenings` is: a resolver field would run once per row of the pull-request queue,
+        and the queue never needs it - only the pick page does.
+
+        Open to any signed-in user. It exposes real per-location inventory for one project, which is
+        the same shape of information the warehouse inventory views already carry."""
+        require_user(info)
+        with SessionLocal() as session:
+            sheet = warehouse_repository.get_pick_sheet(session, uuid.UUID(str(pull_request_id)))
```
Xilous, line 384: `save_pick_draft` already ends with `return get_pick_sheet(session, pr.id)` (pull_requests.py:712), and the resolver discards it and calls `get_pick_sheet` again after the commit. every draft save runs the six-query sheet read twice.

drop the return from the repository function, or reuse it here and refresh only what the commit changed.

frontend/src/graphql/warehouse.ts
```diff
@@ -545,7 +552,29 @@ const PULL_REQUEST_FIELDS = `
   id requestNumber projectId source status requestedBy assignedTo
   createdAt updatedAt approvedAt completedAt cancelledAt cancelledBy cancellationReason
   stagingStatus stagedOpeningCount totalOpeningCount
-  items { id pullRequestId itemType openingNumber openingItemId leaf hardwareCategory productCode requestedQuantity }
+  pickedAt pickedBy partiallyPicked
```
Xilous, line 555: `partiallyPicked` is in `PULL_REQUEST_FIELDS`, selected by
COMPLETE_PULL_REQUEST
STAGE_PULL_OPENINGS
CANCEL_PULL_REQUEST

all three resolvers call `pull_request_to_type(pr, staging)` with no `partially_picked` argument -> null -> Apollo writes null over a cached `true` on the normalized PullRequest. `pullPhase` reads status first for completed and cancelled pulls, so it is invisible today; any future reader of `partiallyPicked` on those rows gets "not evaluated" when the pull really is part-picked.

frontend/src/modules/warehouse/FetchListPanel.tsx
```diff
@@ -0,0 +1,142 @@
...
+                component="tr"
+                key={item.pullRequestItemId}
+                data-testid={`fetch-row-${item.pullRequestItemId}`}
+                sx={{ '& td': { borderTop: '1px solid', borderColor: 'divider', py: 0.75 } }}
+              >
+                {editable && (
+                  <Box component="td">
+                    <Checkbox
+                      checked={Boolean(item.fetchedAt)}
+                      disabled={loading}
```
Xilous: `loading` is the single mutation-hook flag shared by every row - ticking one leaf disables all thirty checkboxes until the round trip returns. `onChanged={() => refetch()}` on PickPage.tsx:336 re-runs the whole six-query pick sheet on every tick. a thirty-leaf shipping pull is thirty full sheet reads and thirty UI freezes.

the mutation returns the updated `PullRequestItem` with its id - write it into the cached `fetchItems` entry instead of refetching, and key the disabled state to the item in flight.

backend/app/repositories/warehouse/pull_requests.py
```diff
@@ -582,10 +1297,24 @@ def _apply_replacement_arrivals(session: Session, pr: PullRequestModel) -> None:
         )
 
 
+def _require_picked(pr: PullRequestModel, action: str) -> None:
```
Xilous, line 1414: `_require_picked` is a gate bolted onto two call sites for a state the enum does not model. IN_PROGRESS now means either "opened, nothing moved" or "picked, stock gone", and only those two functions know the difference.

`shop_assembly_repository.py:62` already carries a comment that the `_APPROVED_PULL_STATUSES` filter exists to exclude "a pull that never got off the ground (PENDING, nothing deducted)" - IN_PROGRESS-unpicked is now exactly that and is not excluded. every future `status == IN_PROGRESS` reader has to remember the extra column.

a distinct status value, or a helper the status checks route through, would make the two states impossible to conflate.

frontend/src/modules/warehouse/FetchListPanel.tsx
```diff
@@ -0,0 +1,142 @@
...
+            {editable && <Box component="th" sx={{ width: 48 }} />}
+            <Box component="th">Leaf</Box>
+            <Box component="th">Location</Box>
+            <Box component="th">State</Box>
+            <Box component="th">Collected</Box>
+          </Box>
+        </Box>
+        <Box component="tbody">
+          {items.map((item) => {
+            const bin = [item.aisle, item.row, item.bay].filter(Boolean).join('-');
```
Xilous: `[item.aisle, item.row, item.bay].filter(Boolean).join('-')` is re-implemented here and again in PickSheetDocument.tsx:174. `pick.ts` already exports `locationLabel` doing exactly this, and PickSheetDocument imports it for the location table three lines away. backend `_location_label` is the same rule again.

call `locationLabel` in both places.

Issue #315, opened 2026-07-23
Register PO in GP unusable for CAD vendors: relay missing `list_tax_details`, tax detail can't be selected


Xilous:
Register PO to GP unusable for CAD vendors. relay does not implement `list_tax_details`; tax-detail selection (required for CAD) can't be made.

reproduced full-lifecycle E2E on Railway 2026-07-23:
- draft PO on GP job 80003 (TUBC), 2 SARGENT line items
- open Register in GP dialog
- pick GP vendor "ASSA ABLOY OF CANADA" -> currency CAD
- cost code -> buyer's designated 210-200 / 310-000, selected 210-200
- tax detail (required) dropdown stays DISABLED, never populates
- console error `[GraphQL error]: unknown op 'list_tax_details'`

CAD PO registration to GP fully blocked.
tax detail CAD-only (GP TX00201, TXDTLTYP=2).
USD vendors blank tax schedule, unaffected.

deployed relay build lacks `list_tax_details` (relay/backend op mismatch).

register-to-GP for CAD needs:
- implement `list_tax_details` in relay -> query GP TX00201 (TXDTLTYP=2) for company/currency, return
- Nexus Register-in-GP dialog selects tax detail (UI wired)
- manual tax-detail entry fallback
- guard relay/backend op parity -> missing op surfaces clear error

Xilous:
stale-relay deployment mismatch, not missing code.

list_tax_details shipped in #257 (commit 98580e1) across:
- relay: channel.py _OPS + econnect.list_tax_details
- backend: gpTaxDetails resolver in app/schemas/relay.py
- frontend: GET_GP_TAX_DETAILS, wired in GpPurchaseOrderDialog.tsx

that merge auto-published relay-v0.1.0-build.29; latest is build.30. both built from master after #257, both carry the op.

the box runs app-relay-v0.1.0-build.28 (the #306 onedir fix), predates #257. update-state.json shows no self-update since 2026-07-21. running relay has no list_tax_details op -> answers unknown_op -> surfaces `[GraphQL error]: unknown op 'list_tax_details'`. tax-detail dropdown stays empty and disabled, no explanation.

live bug fixed by updating the box to build.30 (owner doing this manually via the relay app Update button). code work below = two hardening asks.

branch feature/315-relay-op-parity-tax-fallback

#4 op-parity guard

A1 reactive mapping (backend).
[...]

Issue #314, opened 2026-07-23
Requirements mismatch: adoptGpJob lets Nexus create projects with no GP backing (projects must be GP-owned)


Xilous:
A Nexus project must always correspond to a real GP job. GP owns jobs; a project that is not in GP does not exist. This holds for BOTH entry points:

- Adopt an existing GP job: only jobs that already exist in GP may be adopted.
- Create a job through Nexus: Nexus must also create it in GP (eConnect) so it exists in GP. If the GP create fails, the Nexus project must not be created.

Invariant: no Nexus project without a corresponding GP job. A Nexus-only project is invalid state.

Current behavior (mismatch): `adoptGpJob(jobNumber, jobName)` writes a Nexus project row directly. It trusts the caller, does not call the relay, and does not check GP. A direct GraphQL call can adopt any jobNumber, including one with no GP counterpart, producing a Nexus-only project GP never heard of. There is also no path that creates the job in GP, so Nexus cannot legitimately originate a job today.

The UI masks the gap. The "Adopt GP Job" dialog picks from the live read-only gpJobs query, so through the UI you can only adopt jobs that already exist in GP. That is UI convention, not a backend guard - the mutation itself has no GP-existence check.

Reproduced during Railway E2E on 2026-07-23: adopted "Cowichan 80001" by calling adoptGpJob directly (jobNumber 80001, jobName "Cowichan Dist Hospital") with the relay not connected, and the Nexus project was created. A fabricated jobNumber behaves identically.

Consequences
- Nexus holds projects with no GP backing, violating the GP-owned requirement.
- POs, inventory, assembly, and shipping can attach to a job GP has no record of.
- Orphan and divergent state between Nexus and GP.

Fix direction
- Adopt path: verify jobNumber against GP (relay gpJobs) before writing the Nexus row; reject unknown jobs.
- Create path: if Nexus is allowed to originate a job, create it in GP first (eConnect), and only persist the Nexus project on GP success.
- Enforce both at the resolver so direct GraphQL calls get the same guard as the UI.
- Open question: offline / relay-down behavior, since both guards need a live GP round-trip.

Xilous:
adopt guard in PR #375. create-a-job-in-GP half is feasible, not built.

read-only discovery against the TUBC sandbox found a supported path:
wsiJCJobMaster
wsiJCJobMasterPre
wsiJCJobMasterPost

[...]

Issue #187, opened 2026-07-06
error handling - surface GP server errors or invalidations through relay -> nexus -> nexus end user


Xilous:
end user error screenshots will be the main form of communication for end to end errors

PR #390, opened 2026-07-28
feat(gp): create a GP job from Nexus + auto-adopt every GP job (#380)

### LINKED ISSUE #380: create a GP job from Nexus (wsiJCJobMaster + create-job form)
split out of #376. build it with a full create-job form, not defaulting the non-hardware fields, not adopt-only.

Nexus can already ADOPT an existing GP job (#314, guard shipped in PR #375). this originates a job in GP from Nexus.

wsiJCJobMaster - WennSoft eConnect-style proc for the JC job master, same family as wsiWSCreateUpdatePurchaseOrderIntegration the relay already uses for PO lines. body encrypted. HAS_PERMS_BY_NAME reports EXECUTE = 1. zero taJob* procs; JC00102 is a WennSoft table keyed on WS_Job_Number, not core GP.

both blockers named in #376 cleared.

parameter set resolved by OnlyValidate=1 dry runs against TUBC on 2026-07-28, each wrapped in BEGIN TRAN / ROLLBACK. nothing committed - JC00102 still holds its original 15 jobs and no dry-run row.

217 parameters; all 215 input parameters carry defaults, caller supplies only what it sets. empty call demanded nothing.

minimum required set - 8 of 217, found by validating clean then dropping each field in turn

@I_vWSJobNumber - job number
@I_vWSJobName - blank gives "The description field is blank."
@I_vDivisions - must be a division with division accounts set up; TUBC has exactly one, VANCOUVER
@I_vCustomerNumber - must exist in the customer setup table
@I_vJobAddressCode - must exist in the customer master address table for that customer
@I_vJobBilltoAddressCode - billing address, same table
@I_vTaxScheduleID - must exist in the tax schedule table
@I_vCreatedDate - must fall inside an open fiscal period

explicitly not required

EstimatorID
WSManagerID
WSProjectNumber
BillCustomerNumber
UseTaxSchedule
ScheduleStartDate
ScheduledCompletionDate
BidDueDate

closed-period check reads the supplied CreatedDate, not the clock. TUBC's fiscal calendar stops at FY2025 (SY40101 has 2024 and 2025 only, the latter ending 2025-09-30) - today's date fails with "Job cannot be created within a closed period". dry runs there need dates inside FY2025. a production company with a current calendar will not hit this; the sandbox always will until someone sets up FY2026.

Nexus carries the job number and description. it does not carry customer, address codes, tax schedule or division. those come from live GP reads through the relay, same pattern as the Register in GP dialog for vendors, cost codes and tax details:

customer -> RM00101
job + bill-to address codes -> RM00102, scoped to the chosen customer
tax schedules -> tax schedule table
divisions -> only those with division accounts set up

to build

- relay op wrapping wsiJCJobMaster, following the eConnect orchestration pattern in ops.py, ErrorState / ErrString read back
- create-job form collecting the 8 required fields, each populated from a live GP read
- run OnlyValidate=1 first and surface the proc's own message
- adopt stays as it is, additional path

Xilous:
closes #380.

adds create-job flow, removes manual adoption.

gp_job_sync creates a Nexus project for every JC00102 job the relay reports, triggered by
- ~5 min timer
- every relay reconnect
- new admin syncGpJobs mutation

removed
Adopt GP Job dialog
adoptGpJob
AdoptGpJobInput
tests/test_adopt_gp_job_guard.py

the invariant that guard test defended - only real GP jobs become projects - is now structural: both remaining writers of a project's identity (the sync and createGpJob) take the job number from GP itself.

createGpJob is admin-gated, GP-first, no outbox fallback. the form cannot even be composed without live GP reads, so a queued submit could never have been assembled while the relay was down, and a failed submit is retried from the still-open dialog. proc error message passed through.

eight optional fields behind a Show optional fields toggle
Estimator
WS Manager
WS Project
Bill-to Customer
Use Tax Schedule
Scheduled Start
Scheduled Completion
Bid Due

untouched optional field not sent at all, GP keeps its own default rather than being handed a blank string.

GP discovery (read-only, TUBC)
- divisions with accounts = JCDivisionSETP filtered by EXISTS in JCDivisionAccountsSETP. neither table has a description column, division code is the label. TUBC qualifies exactly one, VANCOUVER.
- @I_vUseTaxSchedule char(15), JC00102.USETAXSCHID char(15) -> tax schedule ID, reuses the tax-schedule picker.
- RM00102 gives ADRSCODE + ADDRESS1/CITY/STATE for dropdown labels.
- job number char(17), project number char(17), job name char(31), everything else char(15), dates datetime. all 16 fields confirmed on the live proc (217 params).

relay
- econnect.create_job wraps wsiJCJobMaster, assembles the EXEC from only the fields the caller set. all 215 inputs carry defaults, explicit NULL is not the same as left defaulted.
- only_validate drives @I_vOnlyValidate, dry run and real call go through one function so they cannot drift on how parameters are built.
- create_job_op runs job_exists pre-check -> OnlyValidate=1 -> real call. duplicate -> job_already_exists instead of a mid-proc failure, which also makes a retry after an ambiguous failure safe.
- four new reads
list_customers (RM00101)
list_customer_addresses scoped to CUSTNMBR (RM00102)
list_tax_schedules (TX00101), distinct from existing list_tax_details (TX00201)
list_divisions
- five ops registered in _OPS. hello frame advertises sorted(_OPS), out-of-date relay fails fast on the existing "update the relay" path (#315).

EConnectError gains an opt-in proc_message. errors.econnect_error_body resolves error_state through DYNAMICS.taErrorCode, which is right for the taXxx procs whose states ARE taErrorCode entries. the WennSoft job proc numbers its states independently, so that lookup would have replaced "Job cannot be created within a closed period" with an unrelated GP description. only create_job sets proc_message, PO error surfacing unchanged.

backend
- live reads
gpCustomers
gpCustomerAddresses
gpTaxSchedules
gpDivisions
- createGpJob replaces adoptGpJob, CreateGpJobInput carries 8 required + 8 optional fields.
- services/gp_job_sync.py modeled on gp_outbox_worker. every iteration wrapped so no error kills the loop, GP_JOB_SYNC_ENABLED kill switch, wake() called from /relay-link registration, task started in lifespan. commits per row so one bad job does not discard the rest, swallows ConflictError per row since a race with a concurrent create is benign.
- adopts all JC00102 jobs, no status filter for now. TUBC has 15. a filter can be added when a production company with closed or historic jobs arrives.

frontend
- CreateGpJobDialog replaces AdoptGpJobDialog. relay status header and read-only company from the old dialog, live GP dropdowns and relay-down gating from GpPurchaseOrderDialog.
- both address selects disabled until a customer is picked, re-fetch on customer change, because an address code is only valid under its own customer.
- Sync from GP button in Admin -> Projects.
- testing/CLAUDE.md updated, the ungated adoptGpJob seeding trick no longer exists, seeding a project without a relay is no longer possible.

no schema change, no migration.

all of test_gp_job_sync.py and one case in test_create_gp_job.py are DB-backed and first execute in CI - there is no local Postgres on this machine.

rollout
- rebuild the relay and self-update the installed tray app before end-to-end testing, otherwise the hello op-set lacks create_job and every call fails the "update the relay" path.
- first relay connect after deploy triggers the sync's first pass, 15 TUBC JC00102 jobs become projects.
- TUBC fiscal calendar ends 2025-09-30, creating with today's date fails with the proc's closed-period message. success path needs a FY2025 CreatedDate.


frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -0,0 +1,533 @@
...
+  // read-only rather than picked. skip: !open so a hidden dialog doesn't poll.
+  const relay = useRelayStatus({ skip: !open });
+  const company = relay.company ?? '';
+  const relayConnected = relay.connected === true;
+
+  const readsSkipped = !open || !relayConnected || !company;
+
+  const { data: customersData, loading: customersLoading } = useQuery<{ gpCustomers: GpCustomerOption[] }>(
+    GET_GP_CUSTOMERS,
+    { variables: { company }, skip: readsSkipped, fetchPolicy: 'cache-first' },
```
Xilous: None of the four GP reads destructure `error`, so a failed read is indistinguishable from an empty result.

The rollout note in this PR says the relay has to be rebuilt before anything works. Until it is, `list_customers` / `list_customer_addresses` / `list_tax_schedules` / `list_divisions` are not in the relay's advertised op set, all four queries fail with `RELAY_OP_UNSUPPORTED`, and the dialog renders: relay chip green, no warning banner, Division / Tax schedule / Customer empty, Create Job permanently disabled because `requiredComplete` can never become true. Same outcome for a relay timeout or a SQL error.

`GpPurchaseOrderDialog` does the opposite (`error: gpTaxDetailsError` -> `isRelayOpUnsupported` -> update-the-relay banner + manual fallback), and the `AdoptGpJobDialog` this replaces surfaced `jobsFetchError` as "Failed to load jobs from GP". Both signals are gone here.

backend/app/schemas/project.py
```diff
@@ -62,64 +66,91 @@ def project_ship_to(self, project_id: strawberry.ID) -> ProjectShipTo | None:
...
+            "bid_due_date": input.bid_due_date.isoformat() if input.bid_due_date else None,
+        }
+
+        try:
+            result = await relay_gateway.relay_call(company, "create_job", payload)
+        except RelayCallError as e:
+            # GP said no - a closed fiscal period, an address code that isn't on the customer, a
+            # division without accounts. The proc words those better than we could, so the message is
+            # passed through to the dialog rather than replaced with a generic failure.
+            raise ValidationError(str(e.message)) from e
```
Xilous: `ValidationError(str(e.message))` drops `e.detail`, which is the only thing `ErrorHandlerExtension` looks at to emit `extensions.relayError` (`detail = getattr(e, "detail", None)` in main.py).

The dialog renders the failure through `GpErrorAlert`, whose whole job is the relay detail table: GP proc, error state, taErrorCode description, plus the `Copy details` payload. With `detail` stripped, every one of those rows is empty and the alert degrades to a bare message string. That is the #187 screenshot channel this component exists for.

Either re-raise with the detail attached (a `ValidationError` subclass that carries `.detail`, or set `err.detail = e.detail` before raising) or let the `RelayCallError` through unchanged.

backend/app/schemas/project.py
```diff
@@ -62,64 +66,91 @@ def project_ship_to(self, project_id: strawberry.ID) -> ProjectShipTo | None:
...
+                project = project_repository.adopt_gp_job(session, job_number=job_number, job_name=job_name)
                 session.commit()
                 session.refresh(project)
                 return project_to_type(project)
 
-        # Off the event loop: the /relay-link read loop runs on it and must not block on Postgres.
-        return await asyncio.to_thread(_persist)
+        try:
+            # Off the event loop: the /relay-link read loop runs on it and must not block on Postgres.
+            return await asyncio.to_thread(_persist)
```
Xilous, line 176: GP-committed-but-Nexus-missing has no recovery path, despite the docstring's claim that queuing "would add an ambiguous-write class that does not otherwise exist here".

Two ways in:
- `relay_call` uses the 30s `DEFAULT_TIMEOUT_SECONDS`. `create_job_op` runs `job_exists` plus two full passes of a 217-parameter WennSoft proc on one connection. A slow GP blows the timeout after the real pass committed -> `RelayTimeoutError`, no project row.
- `_persist` raises `ConflictError` if `gp_job_sync` adopted the same job first (its 300s timer, or an admin hitting Sync from GP, landing between the relay commit and this `to_thread`).

Either way the retry the docstring recommends hits the relay's `job_exists` pre-check and comes back as `job_already_exists` -> `RelayCallError` -> `ValidationError`. The user is now stuck with a dialog that can only produce errors, and the project appears up to 5 minutes later with no feedback.

`gp_job_sync._persist_missing` already swallows `ConflictError` as a benign race with a concurrent create. The create side needs the mirror image: treat an existing Nexus row (and a relay `job_already_exists`) as success and return that project.

backend/app/services/gp_job_sync.py
```diff
@@ -0,0 +1,127 @@
...
+
+def _persist_missing(jobs: list[dict]) -> tuple[int, int]:
+    """Create a project for every reported job that doesn't have one. Returns (total, adopted).
+
+    Committed per row rather than in one batch: one bad job (a number too long for the column, a
+    duplicate racing a create) must not discard the other fourteen. A ConflictError means someone
+    else - create_gp_job, or an earlier pass - got there first, which is the expected outcome on every
+    pass after the first and not worth logging."""
+    with SessionLocal() as session:
+        existing = {pid for pid in session.scalars(select(ProjectModel.project_id)).all()}
```
Xilous, line 62: Unfiltered adoption of every JC00102 row is irreversible from the product: there is no `deleteProject` mutation anywhere in the backend or frontend, and `project_id` is immutable.

TUBC's 15 jobs make this invisible. Pointed at a real company, the first relay connect creates a project for every historic and closed job in the job master, each one landing in `GET_PROJECTS` and therefore in every project picker in import, PO, warehouse, shop assembly and shipping. The only way back is direct SQL.

A status or date filter on `list_jobs` (or on this loop) before the first production relay connect is much cheaper than the cleanup. `list_divisions` in the same PR already takes exactly this position, filtering to divisions that can actually be used.

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -0,0 +1,533 @@
...
+  open: boolean;
+  onClose: () => void;
+}
+
+/** GP column widths, so an over-length value is caught in the field rather than by the proc. */
+const MAX = { jobNumber: 17, jobName: 31, projectNumber: 17, id: 15 };
+
+function todayIso(): string {
+  return new Date().toISOString().slice(0, 10);
+}
```
Xilous, line 71: `toISOString()` is UTC, so this is not today's date west of UTC. In Vancouver the default flips to tomorrow at 17:00 PDT / 16:00 PST.

That matters here specifically because `createdDate` is the field GP validates against the fiscal calendar. On the last day of an open period, an evening submit silently proposes a date in the next period and the user gets "Job cannot be created within a closed period" for a date the form told them was today.

```suggestion
function todayIso(): string {
  // Local calendar date. toISOString() is UTC, which reads as tomorrow west of UTC after ~16:00 -
  // and this is the date GP checks against the fiscal calendar.
  const d = new Date();
  return `${d.getFullYear()}-${String(d.getMonth() + 1).padStart(2, '0')}-${String(d.getDate()).padStart(2, '0')}`;
}
```

relay/src/ucnexus_relay/econnect.py
```diff
@@ -637,6 +645,141 @@ def list_cost_codes(conn, job_number: str) -> list[dict]:
...
+# GP - so the four reads below feed its dropdowns and create_job runs the WennSoft job proc.
+
+
+def list_customers(conn) -> list[dict]:
+    """Read-only: the customer master RM00101, for the create-job customer picker. GP-first, same as
+    every other picker in this file: the options are whatever the company has set up."""
+    rows = conn.cursor().execute(
+        "SELECT RTRIM(CUSTNMBR) AS customer_number, RTRIM(CUSTNAME) AS customer_name "
+        "FROM dbo.RM00101 ORDER BY CUSTNAME"
+    ).fetchall()
```
Xilous: RM00101 is read unfiltered, so inactive customers land in the create-job customer picker.

That contradicts both neighbours in this file: `list_vendors` filters `VENDSTTS = 1` with the rationale "the sync only wants vendors usable on a new PO", and `list_divisions` two functions below filters on division accounts precisely so the dropdown cannot offer a choice that can only fail validation. An inactive customer is the same class of choice.

```suggestion
    rows = conn.cursor().execute(
        "SELECT RTRIM(CUSTNMBR) AS customer_number, RTRIM(CUSTNAME) AS customer_name "
        "FROM dbo.RM00101 WHERE INACTIVE = 0 ORDER BY CUSTNAME"
    ).fetchall()
```

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -0,0 +1,533 @@
...
+  }>(GET_GP_TAX_SCHEDULES, { variables: { company }, skip: readsSkipped, fetchPolicy: 'cache-first' });
+
+  // Addresses are per-customer: the proc validates a code against THAT customer's addresses, so this
+  // re-fetches on every customer change and the two address selects stay disabled until one is picked.
+  const { data: addressesData, loading: addressesLoading } = useQuery<{
+    gpCustomerAddresses: GpCustomerAddressOption[];
+  }>(GET_GP_CUSTOMER_ADDRESSES, {
+    variables: { company, customer: customer?.customerNumber ?? '' },
+    skip: readsSkipped || !customer,
+    fetchPolicy: 'cache-first',
```
Xilous, line 164: All four reads are `cache-first` and there is no refresh control, which the `AdoptGpJobDialog` this replaces did have (the `RefreshCw` IconButton calling `refetchJobs()`).

A customer, address code, tax schedule or division set up in GP after the dialog first opened in this tab will never appear: reopening the dialog re-reads the same Apollo cache entry, and only a full page reload clears it. That is the common case for this form, because a new customer is usually created in GP right before the job that needs it.

`refetch` on one of these plus a small refresh button restores the behaviour.

relay/src/ucnexus_relay/ops.py
```diff
@@ -216,6 +216,38 @@ def create_po_op(conn, *, company: str, request: models.CreatePoRequest) -> mode
...
+    if econnect.job_exists(conn, request.job_number):
+        raise RelayOpError(
+            "job_already_exists",
+            f"job '{request.job_number}' already exists in GP company {company} (JC00102)",
+        )
+
+    econnect.create_job(conn, only_validate=True, **fields)
+    econnect.create_job(conn, only_validate=False, **fields)
+
+    return models.CreateJobResponse(
```
Xilous, line 256: The response echoes `request.job_number` / `request.job_name` back. Nothing here reads JC00102 after the create, so the reply is the caller's own input, not GP's record of what it wrote.

That makes the comment on the backend side wrong (`# GP's own record of what it created, not the input echoed back`, project.py) and quietly retires the #314 invariant the deleted `test_snapshots_gps_own_job_name_not_the_callers` pinned: the project description used to be GP's `WS_Job_Name`, now it is whatever the client typed. GP pads to char(31) and can normalise on write, so the two can differ.

A `SELECT WS_Job_Name FROM JC00102 WHERE WS_Job_Number = ?` after the real pass would make the comment true again.

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -0,0 +1,533 @@
...
+            component="button"
+            type="button"
+            underline="hover"
+            onClick={() => setOptionalOpen((v) => !v)}
+            sx={{ alignSelf: 'flex-start' }}
+          >
+            {optionalOpen ? 'Hide optional fields' : 'Show optional fields'}
+          </Link>
+
+          <Collapse in={optionalOpen} unmountOnExit>
```
Xilous, line 531: `unmountOnExit` unmounts the eight optional inputs but their state lives in this component, so collapsing the section hides the values without clearing them.

Fill in Estimator ID and Bid due, decide against them, click "Hide optional fields", submit: both still ride on the mutation and both still reach `wsiJCJobMaster`. The section header reads as a disclosure, so nothing tells the user the hidden fields are still in play.

Either clear the optional state when the section collapses, or drop `unmountOnExit` and let the filled values stay visible.

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -0,0 +1,533 @@
...
+                </MenuItem>
+              ))}
+            </TextField>
+            <TextField
+              select
+              label="Bill-to address"
+              value={billtoAddressCode}
+              onChange={(e) => setBilltoAddressCode(e.target.value)}
+              required
+              disabled={disabled || !customer || addressesLoading}
```
Xilous: The Bill-to address select is fed from `addresses`, which is scoped to the job `customer` only. When the optional Bill-to customer is set to a different customer, the bill-to address codes on offer still belong to the job customer.

The file's own reasoning ("an address code is only valid under its own customer", and `list_customer_addresses` scoping to CUSTNMBR because "wsiJCJobMaster validates both codes against this table FOR THAT CUSTOMER") points at the bill-to pair needing the same treatment: when `billCustomer` is set, `billtoAddressCode` should come from that customer's RM00102 rows, and should reset when it changes.

backend/tests/test_create_gp_job.py
```diff
@@ -0,0 +1,168 @@
...
+
+def test_persists_the_project_from_gps_own_answer(monkeypatch, _migrate_database):
+    _as_admin(monkeypatch)
+    # GP's reply, not the input, is what the project is built from
+    _relay(monkeypatch, result={"job_number": "NEXUS-380-T9", "job_name": "Name GP Kept"})
+
+    project = _create(job_number="NEXUS-380-T9", job_name="What The Caller Typed")
+
+    assert project.project_id == "NEXUS-380-T9"
+    assert project.description == "Name GP Kept"
```
Xilous, line 232: This commits a real `NEXUS-380-T9` project through the resolver's own `SessionLocal()` and never removes it, so it is not covered by the `db_session` rollback fixture.

A second run against the same database fails: `adopt_gp_job` finds the row and raises `ConflictError`. `test_gp_job_sync.py` in this same PR added `_clean_up_test_projects` for exactly this reason; this module has no equivalent. CI hides it because the Postgres service is fresh each run, local reruns do not.

relay/src/ucnexus_relay/econnect.py
```diff
@@ -637,6 +645,141 @@ def list_cost_codes(conn, job_number: str) -> list[dict]:
...
+    leaving them defaulted, and there is no reason to find out which ones differ. Unset optional fields
+    are simply absent from the statement.
+
+    only_validate drives @I_vOnlyValidate, so the caller's dry run and the real create go through this one
+    function and cannot drift on how the parameters are built - the dry run has to exercise the exact
+    statement the real call will make, or it validates something else.
+
+    @I_vReturnErrorText=1 makes the proc fill @oErrString; that text is carried on the raised error as
+    proc_message so it survives to the user verbatim (see EConnectError)."""
+    supplied = {param: fields[field] for field, param in _CREATE_JOB_PARAMS if fields.get(field) is not None}
```
Xilous: `fields.get(field) is not None` applies to required fields too, so a required field arriving as None is silently dropped from the EXEC instead of raising, and the proc runs on its own default for it. The `unknown` guard on the next line catches the opposite mistake but there is no guard for a missing required one.

Worst case, `supplied` is empty and the f-string produces `EXEC dbo.wsiJCJobMaster\n        ,\n        @I_vOnlyValidate = ?`, a syntax error rather than a readable failure.

`CreateJobRequest` protects the one caller that exists today, which is exactly why this will not be noticed when a second caller appears. A required-key check beside the `unknown` check costs one line.

backend/app/services/gp_job_sync.py
```diff
@@ -0,0 +1,127 @@
...
+    while True:
+        try:
+            if relay_gateway.connected and relay_gateway.company:
+                total, adopted = await run_once()
+                if adopted:
+                    logger.info("gp job sync: adopted %s of %s GP jobs", adopted, total)
+        except asyncio.CancelledError:
+            raise
+        except Exception:  # noqa: BLE001
+            logger.exception("gp job sync iteration failed")
```
Xilous, line 125: `run_once` raises `RelayUnavailableError` when the relay is gone, and the guard above it is a separate read of `relay_gateway.connected` - so every relay drop that lands mid-pass (including `relay_call` finding `_socket is None`) falls through to this line and writes a full ERROR traceback.

A relay restart or a flapping connection is routine, not an iteration failure. `gp_outbox_worker` handles `RelayUnavailableError` explicitly rather than letting it reach the catch-all. Catching it here and continuing quietly keeps the error log meaningful.

frontend/src/graphql/import.ts
```diff
@@ -135,9 +135,47 @@ export const FINALIZE_IMPORT_SESSION = gql`
   }
 `;
 
-export const ADOPT_GP_JOB = gql`
-  mutation AdoptGpJob($input: AdoptGpJobInput!) {
-    adoptGpJob(input: $input) {
+// Issue #380: the create-job form's live GP reads. Nexus stores none of this - customer, address
```
Xilous, line 129: `GET_GP_JOBS` (line 98) had exactly one consumer, `AdoptGpJobDialog`, deleted in this PR. It is now dead, and so is the `gpJobs` GraphQL field behind it, which no longer has a frontend caller at all now that the sync reads `list_jobs` straight off the relay.

Worth deleting the export here; the `gpJobs` resolver is a judgement call since it is still reachable through the API.

Issue #382, opened 2026-07-28
finish RELAY_SECRET_ENC_KEY retirement: delete the variable, drop the leftover references


Xilous:
split out of #376. the retirement itself is done and proven - the variable is blank on Railway, the backend redeployed, and the relay re-handshook with "WebSocket /relay-link" [accepted] and no 403, so the install authenticates on secret_hash alone with no key present.

what is left is deleting the variable outright rather than leaving it blank. blocked on permissions, not on anything technical.

railway variable delete RELAY_SECRET_ENC_KEY --service backend --environment production

after that, `RELAY_SECRET_ENC_KEY` can also come out of backend/.env.example and the config.py comment that describes it as legacy. crypto.py's encrypt_secret/decrypt_secret and has_encryption_key stay useful only if a pre-067 row could ever reappear; it cannot, since set_install_secret NULLs secret_encrypted on every enroll and adopt.

Xilous:
done, closing.

variable deleted, not blank - gone from the backend production variable list. `railway variable delete` has no `--skip-deploys`, redeploy came back SUCCESS.

code merged as d6d0ac5, PR #391
- `app/config.py` `RELAY_SECRET_ENC_KEY` constant removed with its legacy comment, imported nowhere; crypto.py reads `os.getenv` directly
- `backend/.env.example` entry removed
- retirement comments moved to past tense in
  crypto.py docstring
  `relay_repository.delete_install`
  `schemas/relay.delete_relay_install`
  testing/CLAUDE.md
- testing/CLAUDE.md claimed `enroll_install` calls `encrypt_secret`, so a successful enrolment proved the Fernet key valid. untrue since 067, enrolment stores a hash. conclusion was right, the reason given for it was not.

live proof
- deploy succeeded 00:22:01Z, `lastSeenAt` 2026-07-29T00:22:02.422739, 1.4s later. `last_seen_at` is written only by `authenticate_secret` on a successful handshake - the relay re-authenticated against the new instance on `secret_hash` alone, no key present anywhere.
- `relayStatus` connected true, company TUBC, build relay-v0.1.0-build.41
- Admin -> Relay Installs renders RELAY CONNECTED, single install ENROLLED / CONNECTED, Remove disabled with "This relay is connected right now."

kept
`encrypt_secret`
`decrypt_secret`
`has_encryption_key`
`secret_encrypted` column

[...]

Issue #376, opened 2026-07-27
Consolidated follow-ups: relay key retirement + outbox proof, GP job creation, pull-request workflow


Xilous:
consolidates leftovers of #353, #314, #367, all closed in favour of this. each was mostly delivered; what remains is operational or blocked on a decision, and was not worth three open issues.

from #353, relay hands-off

code complete, PRs #352, #354, #355, #356, #364, #365 merged.

steps 1 and 3 verified on TAGGING3W10, 2026-07-27. adopt window recovered the drifted secret, relay.log shows secret_rejected at 10:53 and 12:18 then clean connects from 12:19, /health reports channel connected true. poller bootstrapped and self-updated build.32 -> build.36 unattended at 12:17.

step 2, retire RELAY_SECRET_ENC_KEY. unblocked by #366: stale pre-067 rows kept secret_encrypted populated forever and could not be deleted until install removal shipped.
- remove stale installs via Admin -> Relay Installs -> Remove
- confirm `SELECT count(*) FROM relay_installs WHERE secret_encrypted IS NOT NULL` returns 0
- unset RELAY_SECRET_ENC_KEY on Railway
- redeploy
- confirm the relay still authenticates
- delete the variable
- backend log "relay install upgraded to hashed secret" confirms the live row migrated

step 4, exercise the GP write outbox. submit a receive with the relay disconnected, confirm Queued, reconnect, confirm it drains within about a second of /relay-link accepting.

Windows Service stays abandoned, reboot-with-nobody-logged-in caveat stands, per #353.

from #314, create a job in GP from Nexus

adopt guard shipped in PR #375. adopt_gp_job requires a signed-in caller, verifies the job number against the connected relay's live job master before writing, uses GP's job name not the caller's. that closes the invariant for the path actually in use.

originating a job from Nexus is not built, and nothing does it today. feasible - read-only discovery on TUBC found wsiJCJobMaster with Pre and Post siblings, WennSoft eConnect-style integration proc for the JC job master, same family as wsiWSCreateUpdatePurchaseOrderIntegration already used by the relay for PO lines. HAS_PERMS_BY_NAME reports EXECUTE = 1. body encrypted, same as the PO proc. zero taJob procs, because JC00102 is a WennSoft table keyed on WS_Job_Number, not core GP.

217 parameters, including
WSJobNumber
WSProjectNumber
WSJobName
Divisions
CustomerNumber
JobAddressCode
EstimatorID
WSManagerID
UpdateIfExists
[...]

PR #391, opened 2026-07-28
chore(relay): drop the dead RELAY_SECRET_ENC_KEY wiring (#382)

Xilous:
finishes RELAY_SECRET_ENC_KEY retirement started in #353 PR C, carried through #376. key left the auth path at migration 067.

removed
- `RELAY_SECRET_ENC_KEY` constant and its legacy comment in `app/config.py`, imported nowhere; crypto.py reads `os.getenv` directly
- its entry in `backend/.env.example`

corrected
- `crypto.py` docstring - retirement done, a legacy row cannot come back; `set_install_secret` is the only writer of `secret_encrypted` and NULLs it on every enroll and adopt
- `relay_repository.delete_install` - pointer moved from `app/config.py` to crypto.py
- `schemas/relay.delete_relay_install` - gated-count sentence to past tense
- `testing/CLAUDE.md` - enrolment stores a hash, `enroll_install` does not call `encrypt_secret`; a 403 cadence running through the enrolment instant still means a stale secret

kept
`encrypt_secret`
`decrypt_secret`
`has_encryption_key`
`secret_encrypted` column

reachable only if the variable is set again. deleting them and dropping the column is a migration, separate job.

no behaviour change - every read was `os.getenv(NAME, "")`, blank and absent are the same value. variable blank on Railway since 2026-07-28 01:14Z, relay authenticating on `secret_hash` alone since 01:15Z.

ruff check and ruff format --check clean over backend/. four relay test files run 13 passed, 27 skipped locally, skips are the Postgres-backed ones that only run in CI.

one step is not in this PR and needs a hand with Railway permissions - the variable is blank, not deleted.

    railway variable delete RELAY_SECRET_ENC_KEY --service backend --environment production


backend/app/crypto.py
```diff
@@ -6,12 +6,17 @@
...
 
-    SELECT count(*) FROM relay_installs WHERE secret_encrypted IS NOT NULL;  -- 0 means it is unused
+    SELECT count(*) FROM relay_installs WHERE secret_encrypted IS NOT NULL;  -- 0
 
-No install enrolled or adopted from 067 on reads that key for anything."""
+With no key present `has_encryption_key` is False and the legacy decrypt path is skipped outright, so
+in practice nothing below `hash_token` runs outside the tests. A legacy row also cannot come back:
+`set_install_secret` is the only writer of `secret_encrypted` and it NULLs the column on every enroll
+and adopt. These three functions are kept only so that a row restored from an old backup could still
+be read by setting the variable again - deleting them (and the column) is a separate job."""
```
Xilous: "setting the variable again" only works if someone still has the retired value. fernet ciphertext is readable only with the exact key that produced it; a freshly generated one decrypts nothing.

this PR deletes both places that recorded the variable:
config.py comment
.env.example

both carried the `Fernet.generate_key()` line. PR body's last step is `railway variable delete RELAY_SECRET_ENC_KEY --service backend --environment production`, after which the value is unrecoverable.

archive the key value somewhere durable before that delete runs and name the place in this docstring, or say the functions are kept only for the tests and drop the restore claim. as written it is a guarantee that goes false the moment the PR is finished.

backend/app/crypto.py
```diff
@@ -6,12 +6,17 @@
...
+retirement is finished (#382): the gating count reached 0, and `RELAY_SECRET_ENC_KEY` is set in no
+environment any more - not on Railway, not in `.env.example`.
 
-    SELECT count(*) FROM relay_installs WHERE secret_encrypted IS NOT NULL;  -- 0 means it is unused
+    SELECT count(*) FROM relay_installs WHERE secret_encrypted IS NOT NULL;  -- 0
 
-No install enrolled or adopted from 067 on reads that key for anything."""
+With no key present `has_encryption_key` is False and the legacy decrypt path is skipped outright, so
+in practice nothing below `hash_token` runs outside the tests. A legacy row also cannot come back:
+`set_install_secret` is the only writer of `secret_encrypted` and it NULLs the column on every enroll
```
Xilous: `set_install_secret` is not the only writer of `secret_encrypted`. `backend/main.py`'s `/admin/reset-data` snapshots the column and writes it straight back after the schema rebuild:
SELECT at main.py:329 lists `secret_encrypted`
INSERT at main.py:350-352 binds `:secret_encrypted` from that snapshot

a legacy row comes back across a reset, verbatim, on any database that still has one. testing/CLAUDE.md:65 names that endpoint as the one that preserves relay_installs, and TESTING_ENABLED is on for the Railway test backend. the claim holds only because the gating count is already 0, not because the code prevents it.

qualify it, or name main.py as the second writer.

backend/app/crypto.py
```diff
@@ -6,12 +6,17 @@
...
+`relay_repository.authenticate_secret` upgrades to a hash in place on their next handshake. That
+retirement is finished (#382): the gating count reached 0, and `RELAY_SECRET_ENC_KEY` is set in no
+environment any more - not on Railway, not in `.env.example`.
 
-    SELECT count(*) FROM relay_installs WHERE secret_encrypted IS NOT NULL;  -- 0 means it is unused
+    SELECT count(*) FROM relay_installs WHERE secret_encrypted IS NOT NULL;  -- 0
 
-No install enrolled or adopted from 067 on reads that key for anything."""
+With no key present `has_encryption_key` is False and the legacy decrypt path is skipped outright, so
+in practice nothing below `hash_token` runs outside the tests. A legacy row also cannot come back:
```
Xilous: inverted. in file order `hash_secret` sits below `hash_token` (crypto.py:65), and it is the busiest function in the module:
- `set_install_secret` calls it on every enroll and adopt
- `authenticate_secret` calls it on every handshake

you mean nothing above `_sha256_hex` runs in production. as written, the sentence tells the reader the production hash path is test-only.

backend/app/crypto.py
```diff
@@ -6,12 +6,17 @@
 enrolled relay is orphaned with no recovery path, because the plaintext exists nowhere else.
 
 `encrypt_secret` / `decrypt_secret` survive only to read pre-067 rows, which
-`relay_repository.authenticate_secret` upgrades to a hash in place on their next handshake. Once no
-`secret_encrypted` remains, `RELAY_SECRET_ENC_KEY` is dead weight and can be deleted:
+`relay_repository.authenticate_secret` upgrades to a hash in place on their next handshake. That
+retirement is finished (#382): the gating count reached 0, and `RELAY_SECRET_ENC_KEY` is set in no
```
Xilous, line 10: the past-tense sweep missed the migration that created the gate. `backend/alembic/versions/067_relay_secret_hash.py:19-20` still reads "Once `SELECT count(*) FROM relay_installs WHERE secret_encrypted IS NOT NULL` reaches 0, `RELAY_SECRET_ENC_KEY` is dead weight and can be deleted from the environment" - future tense, and the origin of the very count this PR is closing out.

these four got updated, 067 did not:
crypto.py
relay_repository.py
schemas/relay.py
testing/CLAUDE.md

067 is the one someone reads next time they trace where the gate came from.

testing/CLAUDE.md
```diff
@@ -46,10 +46,10 @@ with `relayInstalls { label enrolled enrolledAt lastSeenAt }` and the Railway ba
...
   whole duration, so look for a *gap*: an unbroken 403 cadence with no `[accepted]` line means the
   channel never came up. Read it across the whole life of the deployment, not a sample window.
-- **A 403 cadence that runs unbroken *through* the enrolment instant means a stale secret, and it
-  also exonerates the Fernet key.** `enroll_install` calls `encrypt_secret`, so a successful enrolment
-  proves `RELAY_SECRET_ENC_KEY` was valid at that moment; if auth still 403s seconds later, the
-  mismatch is in the secret, not the key. Usual cause: enrolment rewrote `[auth] shared_secret` in
+- **A 403 cadence that runs unbroken *through* the enrolment instant means a stale secret.**
+  `enroll_install` stores a SHA-256 hash of whatever secret the relay generated, so a successful
+  enrolment cannot leave the row wrong about it; if auth still 403s seconds later, the mismatch is in
+  the secret the relay is presenting. Usual cause: enrolment rewrote `[auth] shared_secret` in
```
Xilous, line 52: the bullet at testing/CLAUDE.md:62 was left behind by this edit and now misleads. it tells the debugger to read `hash_rows` / `legacy_rows` / `encryption_key_present` because those "name the actual cause", but after this PR `encryption_key_present` is permanently `false` in every environment. anyone following the runbook reads a false flag on a healthy backend and diagnoses a config problem.

say the field is expected to be false now, leaving two causes:
`no enrolled installs`
`secret mismatch`

backend/app/crypto.py
```diff
@@ -6,12 +6,17 @@
...
+environment any more - not on Railway, not in `.env.example`.
 
-    SELECT count(*) FROM relay_installs WHERE secret_encrypted IS NOT NULL;  -- 0 means it is unused
+    SELECT count(*) FROM relay_installs WHERE secret_encrypted IS NOT NULL;  -- 0
 
-No install enrolled or adopted from 067 on reads that key for anything."""
+With no key present `has_encryption_key` is False and the legacy decrypt path is skipped outright, so
+in practice nothing below `hash_token` runs outside the tests. A legacy row also cannot come back:
+`set_install_secret` is the only writer of `secret_encrypted` and it NULLs the column on every enroll
+and adopt. These three functions are kept only so that a row restored from an old backup could still
```
Xilous: `has_encryption_key` runs in production today, not the backup case. `relay_repository.authenticate_secret:168` calls it unconditionally on every rejected handshake to build the `encryption_key_present` field and pick the `cause` string in the warning log.

"These three functions" also does not match the four defined below:
`_fernet`
`has_encryption_key`
`encrypt_secret`
`decrypt_secret`

scope the sentence to `encrypt_secret` / `decrypt_secret`.

backend/app/crypto.py
Xilous: when the separate job lands, the column drop breaks more than this module. `backend/main.py`'s `/admin/reset-data` hardcodes `secret_encrypted` in both raw statements:
SELECT column list (main.py:329)
INSERT column list plus its bind (main.py:350-352)

drop the column -> both fail with UndefinedColumn -> reset endpoint down, and the testing workflow depends on it.

name main.py here so whoever picks up the migration finds it. this docstring currently points them at `set_install_secret` as the only place that touches the column.

backend/app/repositories/relay_repository.py
```diff
@@ -119,9 +119,9 @@ def delete_install(session: Session, install_id: uuid.UUID) -> dict | None:
 
     Hard delete, not a soft-delete flag. A flag would have to be honoured by `authenticate_secret` too or
     the "deleted" install keeps authenticating - a new filter on the handshake hot path and a new way to
-    be silently wrong - and it would leave `secret_encrypted` populated, which is the very thing blocking
-    the RELAY_SECRET_ENC_KEY retirement (see app/config.py). Nothing references relay_installs by foreign
-    key, so this is a clean single-row delete.
+    be silently wrong - and it would leave `secret_encrypted` populated, which is the very thing that held
+    up the RELAY_SECRET_ENC_KEY retirement until #382 (see app/crypto.py). Nothing references
```
Xilous: `authenticate_secret` at :167 does work on every rejected handshake for nothing. it runs `select(RelayInstall).where(RelayInstall.secret_encrypted.is_not(None)).all()` before checking `key_present`, materializing ORM rows whose only surviving consumer is `len(legacy)` in the log extras and the `cause` string. key unset in every environment -> loop underneath can never run -> guaranteed-empty query on the failure path, which is the reconnect-storm path where a whole fleet re-dials with a stale secret.

gate it on `has_encryption_key()` first, or make it a `select(func.count())` so it stops building objects.

backend/app/repositories/relay_repository.py
```diff
@@ -119,9 +119,9 @@ def delete_install(session: Session, install_id: uuid.UUID) -> dict | None:
 
     Hard delete, not a soft-delete flag. A flag would have to be honoured by `authenticate_secret` too or
     the "deleted" install keeps authenticating - a new filter on the handshake hot path and a new way to
-    be silently wrong - and it would leave `secret_encrypted` populated, which is the very thing blocking
-    the RELAY_SECRET_ENC_KEY retirement (see app/config.py). Nothing references relay_installs by foreign
-    key, so this is a clean single-row delete.
+    be silently wrong - and it would leave `secret_encrypted` populated, which is the very thing that held
```
Xilous: the first half of this justification is dead, not just past tense. a soft-deleted row created since 067 has `secret_encrypted` NULL anyway - `set_install_secret` nulls it - so "it would leave `secret_encrypted` populated" only ever applied to pre-067 rows, and there are none left.

lead with the reason in the sentence before it:
- the row stays a live credential `authenticate_secret` accepts
- a soft-delete flag puts a new filter on the handshake hot path

same for `schemas/relay.py:229-230`.

PR #394, opened 2026-07-29
fix(gp): estimator/manager pickers, and stop claiming a create that did not happen (#392)

### LINKED ISSUE #392: create-job follow-up: estimator/manager pickers, and the duplicate-submit toast lies
follow-up to #380, both found during end-to-end testing on TUBC with relay build.41. neither is a regression, neither blocks the feature.

1. Estimator ID and WS Manager ID free text, GP validates them

EstimatorID "EST1" -> wsiJCJobMaster -> "The estimator does not exist in the payroll master table.", error state 51117. the OnlyValidate pass catches it so nothing is written, but the user has no way to discover a valid value from inside Nexus.

#380 made Estimator/WS Manager/WS Project free text on the grounds that no GP list existed in this build. that is wrong for two of the three - UPR00100, the payroll employee master, is the list. EMPLOYID char(15) matches Estimator_ID and WS_Manager_ID on JC00102.

TUBC has two employees, both active:
IANB (Ian Brown)
JONATHANR (Jonathan Ruballos)

no separate WennSoft estimator table; JC90102 carries WS_Estimator_Name and WS_Manager_Name but that is denormalized reporting output, not a master.

to build
- list_employees relay read over UPR00100, filtered INACTIVE = 0
  EMPLOYID
  FRSTNAME
  LASTNAME
- gpEmployees backend query following gp_customers
- Estimator ID and WS Manager ID become pickers in CreateGpJobDialog
- WS Project Number stays free text, it is not validated against anything

the INACTIVE = 0 filter is the same rule as list_vendors' VENDSTTS, list_customers' INACTIVE and list_divisions' accounts filter - a picker must not offer a choice that can only fail validation.

2. duplicate-submit toast says "created in GP" when nothing was created

createGpJob adopts the existing job rather than dead-ending when GP already holds that job number. that is the retry-after-ambiguous-failure path and it is correct - verified live, no second GP job created, existing job name not overwritten.

frontend unconditionally shows "Job X created in GP." on success, so resubmitting an existing job number reports a creation that did not happen. the mutation knows which path it took, the frontend needs to be able to tell.

Xilous:
closes #392. both fixes out of #380 end-to-end testing.

1. estimator and WS Manager are pickers

wsiJCJobMaster validates both against the payroll master. typed value comes back as "The estimator does not exist in the payroll master table", error state 51117, and there was no way to discover a valid one from inside Nexus. #380 made them free text on the assumption that no GP list existed in this build - UPR00100 is the list, EMPLOYID is char(15) like Estimator_ID and WS_Manager_ID on JC00102.

- list_employees relay read over UPR00100 (EMPLOYID, FRSTNAME, LASTNAME), filtered INACTIVE = 0, same as list_customers, list_vendors, list_divisions - a picker must not offer a choice that can only fail validation
- gpEmployees backend query + GpEmployee type
- Estimator and WS Manager become selects in CreateGpJobDialog
- WS Project Number stays free text, nothing validates it

TUBC has two employees, IANB and JONATHANR, both active.

2. the duplicate-submit toast no longer claims a create that did not happen

createGpJob adopts the existing job rather than dead-ending when GP already holds that job number. that is the retry-after-ambiguous-failure path and it is correct. but the dialog showed "Job X created in GP." unconditionally on success, so resubmitting an existing number reported a creation that never happened.

createGpJob now returns {created, project} instead of a bare Project. created=false drives "already existed in GP and is now a project".

created stays true when the sync wins the persist race - GP did create the job on that call, only the Nexus row was written by someone else first.

createGpJob return type changes. the mutation shipped in #380, CreateGpJobDialog is its only consumer.

relay needs rebuilding - list_employees is a new op, installed build must be updated before the estimator picker populates. an old relay trips the "relay is too old to create jobs" banner rather than showing an empty dropdown.


frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -190,21 +214,28 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
   // A failed read is NOT an empty list. Without this the deploy-before-relay-rebuild window (the four
   // ops are not in an older relay's advertised op-set) renders a green relay chip over three empty
   // required dropdowns and a button that can never enable, with nothing saying why.
-  const readError = customersError ?? divisionsError ?? taxSchedulesError ?? addressesError ?? null;
-  const readsUnsupported = [customersError, divisionsError, taxSchedulesError, addressesError].some((e) =>
-    isRelayOpUnsupported(e),
-  );
+  const readError =
+    customersError ?? divisionsError ?? taxSchedulesError ?? employeesError ?? addressesError ?? null;
+  const readsUnsupported = [
```
Xilous: employeesError in readsUnsupported -> any relay built for #380 but not #392 renders "The connected relay is too old to create jobs". that relay supports create_job, only list_employees is missing.

readsUnsupported does not disable the form (disabled = !relayConnected || loading). admin fills every required field and creates the job while being told the relay cannot.

split the required reads
customers
divisions
tax schedules
addresses

from the optional-only employees read, so an unsupported list_employees degrades the estimator picker instead of claiming the whole feature is dead.

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -530,25 +569,45 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
 
           <Collapse in={optionalOpen} unmountOnExit>
             <Stack spacing={2}>
+              {/* #392: GP validates both against the payroll master, so typing a name only ever
+                  earns "The estimator does not exist in the payroll master table". */}
               <Stack direction="row" spacing={2}>
                 <TextField
```
Xilous: dropdown with no fallback replaces #380's free-text estimator/WS manager. between the backend deploy and every relay install being rebuilt, and any time list_employees errors, neither estimator nor WS manager can be set at all.

gpError.ts:44-46 states the house pattern for RELAY_OP_UNSUPPORTED - "show an 'update the relay' banner and fall back (e.g. manual tax-detail entry) instead of leaving a dropdown silently empty".

keep the free-text input as the fallback when employeesError is set.

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -306,7 +337,15 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
           },
         },
       });
-      showToast(`Job ${jobNumber.trim()} created in GP.`, 'success');
+      // #392: GP may not have created anything - when it already held the job number the mutation
+      // adopts it instead, and saying "created" there would report something that did not happen.
+      const created = response.data?.createGpJob?.created !== false;
```
Xilous: `!== false` gives created=true for every value other than false, including undefined and null. absent or partial data -> toast says "created in GP" for a job GP did not create, which is the bug this PR fixes.

```suggestion
      const created = response.data?.createGpJob?.created === true;
```

backend/app/schemas/relay.py
```diff
@@ -137,6 +139,15 @@ async def gp_customer_addresses(
         result = await relay_gateway.relay_call(company, "list_customer_addresses", {"customer": customer})
         return [gp_customer_address_to_type(a) for a in result["addresses"]]
 
+    @strawberry.field
+    async def gp_employees(self, info: strawberry.Info, company: str) -> list[GpEmployee]:
+        """Live payroll employee master (UPR00100) via the connected relay, for the create-job
+        estimator and WS manager pickers (#392). The job proc validates both against this master, so
+        they cannot be free text."""
+        require_user(info)
```
Xilous: require_user exposes the full GP payroll employee master, every employee id and name, to any signed-in Nexus user
warehouse
shop assembly
shipping

only consumer is CreateGpJobDialog, and createGpJob is require_admin. gate this to match its caller.

```suggestion
        require_admin(info)
```

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -150,6 +163,16 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
     skip: readsSkipped,
     fetchPolicy: 'cache-first',
   });
+  const {
+    data: employeesData,
+    loading: employeesLoading,
+    error: employeesError,
+    refetch: refetchEmployees,
+  } = useQuery<{ gpEmployees: GpEmployeeOption[] }>(GET_GP_EMPLOYEES, {
```
Xilous, line 228: both consumers sit inside the Collapse behind "Show optional fields", which starts closed. the query still fires on every dialog open, one extra GP round-trip over the relay per open.

`skip: readsSkipped || !optionalOpen` removes it and keeps the optional read out of the readsUnsupported banner for anyone who never opens the section.

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -530,25 +569,45 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
...
                   size="small"
                   sx={{ flex: 1 }}
-                  slotProps={{ input: { sx: monoSx }, htmlInput: { maxLength: MAX.id } }}
-                />
+                >
+                  <MenuItem value="">
+                    <em>None</em>
+                  </MenuItem>
+                  {employees.map((e) => (
+                    <MenuItem key={e.employeeId} value={e.employeeId}>
```
Xilous: the TextFields this replaces carried `slotProps={{ input: { sx: monoSx } }}`, nothing here is mono. EMPLOYID is an identifier.

the two pickers next to it keep the mono voice
Division `<MenuItem sx={monoSx}>`
Customer `slotProps={{ listbox: { sx: monoSx } }}`

DESIGN.md line 205 - "The Two-Voice Rule. Source Sans 3 for UI, Plex Mono for identifiers ... Anything that names a thing in the system (a PO, a code, a bin) is mono".

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -530,25 +569,45 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
...
+                  disabled={disabled || employeesLoading}
                   size="small"
                   sx={{ flex: 1 }}
-                  slotProps={{ input: { sx: monoSx }, htmlInput: { maxLength: MAX.id } }}
-                />
+                >
+                  <MenuItem value="">
+                    <em>None</em>
+                  </MenuItem>
+                  {employees.map((e) => (
```
Xilous: byte-identical to the Estimator select above apart from label, value and setter, and rebuilds the same MenuItem array a second time. one `<EmployeeSelect label value onChange employees disabled />` covers both.

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -190,21 +214,28 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
...
+    taxSchedulesError,
+    employeesError,
+    addressesError,
+  ].some((e) => isRelayOpUnsupported(e));
 
   const refreshReads = useCallback(() => {
     void refetchCustomers();
     void refetchDivisions();
     void refetchTaxSchedules();
+    void refetchEmployees();
```
Xilous, line 290: refreshReads refetches employees too. tooltip at line 404 still reads "Re-read customers, divisions and tax schedules from GP".

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -190,21 +214,28 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
   // A failed read is NOT an empty list. Without this the deploy-before-relay-rebuild window (the four
```
Xilous: stale. five ops now, not four - four required dropdowns plus one optional. employees is not a required dropdown, which is why it should not feed the same banner.

same "four reads" wording in CreateGpJobDialog.test.tsx:197.

relay/src/ucnexus_relay/models.py
```diff
@@ -203,6 +203,17 @@ class TaxSchedulesResponse(BaseModel):
     tax_schedules: list[TaxScheduleOut]
 
 
+class EmployeeOut(BaseModel):
```
Xilous, line 206: EmployeeOut and EmployeesResponse are never constructed or validated. _run_list_employees returns econnect.list_employees' raw dicts straight out.

same as
CustomersResponse
TaxSchedulesResponse
DivisionsResponse

a second description of the wire shape that nothing enforces, free to drift from the SELECT that actually produces it.

relay/src/ucnexus_relay/econnect.py
```diff
@@ -668,6 +668,36 @@ def list_customers(conn, *, active_only: bool = True) -> list[dict]:
     return [{"customer_number": r.customer_number, "customer_name": r.customer_name or None} for r in rows]
 
 
+def list_employees(conn, *, active_only: bool = True) -> list[dict]:
```
Xilous, line 672: five reads now.

section header at line 651 still says "the four reads below feed its dropdowns"
models.py:171 says "The four reads that feed the create-job form's dropdowns"

relay/src/ucnexus_relay/channel.py
```diff
@@ -190,6 +190,13 @@ def _run_list_tax_schedules(company: str, payload: dict) -> dict:
     return {"company": company, "tax_schedules": rows}
 
 
+def _run_list_employees(company: str, payload: dict) -> dict:
```
Xilous, line 193: payload is ignored, so list_employees' active_only is unreachable through the channel, its only caller. _run_list_vendors forwards `payload.get("active_only", True)`. tests/test_job_reads.py:120 pins a branch nothing can reach.

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -530,25 +569,45 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
 
           <Collapse in={optionalOpen} unmountOnExit>
             <Stack spacing={2}>
+              {/* #392: GP validates both against the payroll master, so typing a name only ever
+                  earns "The estimator does not exist in the payroll master table". */}
               <Stack direction="row" spacing={2}>
                 <TextField
-                  label="Estimator ID"
+                  select
+                  label="Estimator"
```
Xilous, line 637: plain Select over a master table with no bound. TUBC has two employees, a production payroll master runs to hundreds, each rendered as an unsearchable MenuItem in a half-width field inside a maxWidth="sm" dialog.

Customer, the other master-table picker here, is an Autocomplete for exactly that reason.

frontend/src/graphql/import.ts
```diff
@@ -164,14 +176,19 @@ export const GET_GP_DIVISIONS = gql`
...
 export const CREATE_GP_JOB = gql`
   mutation CreateGpJob($input: CreateGpJobInput!) {
     createGpJob(input: $input) {
-      id
-      projectId
-      description
-      client
-      jobSiteName
+      created
+      project {
```
Xilous, line 187: the dialog reads only `created`. the project payload is unused, the list refreshes via `refetchQueries: [GET_PROJECTS]`. CLAUDE.md - "Don't request fields you don't use."

relay/src/ucnexus_relay/econnect.py
```diff
@@ -668,6 +668,36 @@ def list_customers(conn, *, active_only: bool = True) -> list[dict]:
...
+
+    INACTIVE = 0 by default, the same rule as list_customers / list_vendors / list_divisions: never
+    offer a choice that can only fail validation. There is no WennSoft-specific estimator master;
+    JC90102 carries WS_Estimator_Name and WS_Manager_Name, but that is denormalized reporting output."""
+    sql = (
+        "SELECT RTRIM(EMPLOYID) AS employee_id, RTRIM(FRSTNAME) AS first_name, "
+        "RTRIM(LASTNAME) AS last_name FROM dbo.UPR00100 "
+    )
+    if active_only:
+        sql += "WHERE INACTIVE = 0 "
```
Xilous, line 693: wsiJCJobMaster validates the estimator against UPR00100, not UPR00100 WHERE INACTIVE = 0. this picker is narrower than what the proc accepts.

a job recreated or backdated against an estimator since marked inactive cannot be expressed through the form, and there is no override.

list_customers has the same shape but its filter matches a real business rule (a customer nobody can invoice). a terminated employee is still a valid historical estimator.

PR #412, opened 2026-07-29
feat(admin): pick GP buyer identity from GP's buyer master, register new ones from Nexus (#409)

### LINKED ISSUE #409: URGENT - admin user management - gp buyer id
when admin is setting a user's gp buyer id, it needs to be a dropdown of all registered gp buyer id's in the targeted company within GP.

or

at least give the admin a workflow within nexus to actually create a new buyer id registration, manage gp buyer id registrations, so that they can assign it to a user that they need to assign it to.

currently admins are required to open GP locally just to see existing gp buyer ids and to create one. this can be done through Nexus through the relay instead

Xilous:
Closes #409.

Admin -> User Management GP Buyer ID goes free text -> dropdown fed live from GP's buyer master, with inline registration for a buyer that isn't there yet. a typo was previously invisible until much later, when taPoHdr rejected the unregistered BUYERID with error 269.

read-only GP discovery against TUBC found:
- `taCreateBuyer` exists, our login has EXECUTE. params `@I_vBUYERID char(15)`, `@I_vDSCRIPTN char(30)`, RequesterTrx, USRDEFND1-5, plus the two OUTPUT error params.
- no `taDeleteBuyer`, no update proc. taCreateBuyer plus its Pre/Post hooks only.
- POP00101 = `BUYERID char(15)`, `DSCRIPTN char(31)`, NOTEINDX, DEX_ROW_ID.
- DYNAMICS.taErrorCode states for it: 2681 BUYERID empty, 2684 duplicate in POP00101, 2683 insert failed, 2679 null input.
- TUBC has `mira` (Accounting), `donr` (Don Roberton). TUCSH ~100, including ids with spaces (`Anna Wyzynski`, `Nazanin T.`). BUYERID is not a GP user id, so the register form is free text.

relay adds two channel ops to `_OPS` in channel.py, advertised in the hello frame:
- `list_buyers_detailed` - POP00101 with descriptions, separate from the existing `list_buyers`.
- `create_buyer` - `create_buyer_op` pre-checks POP00101 -> EXECs taCreateBuyer -> reads the row back. the proc rejects a duplicate itself (2684) but from inside itself, so pre-empting it is what turns a re-register into a sentence the dialog can show, and what makes a retry after an ambiguous failure safe. same reasoning as create_job_op's job_exists check.

backend adds two admin-gated resolvers:
`Query.gpBuyersDetailed(company)`
`Mutation.createGpBuyer(buyerId, description)`

admin-gated because the descriptions are a staff roster by another name (in production, every buyer's email) - same reasoning that made gpEmployees admin-only.
buyerId/description bounded against GP's char widths before the call.
existing `gpBuyers` (bare ids, require_user) untouched, still backs the Create PO dropdown.

frontend adds `useGpBuyers` + `GpBuyerSelect`, shared by both admin screens, covering three states:
list
relay down
relay too old

- User Management free-text field -> dropdown with a pinned "+ Register new GP buyer..." entry (VendorSelect pattern). an already-linked id still displays while the relay is down.
- Buyers page gains a "Registered in GP" panel over the buyer master with a Register button, add-assignment dialog buyer field -> same dropdown, assignments for ids GP never registered carry a warning marker.
- relay down or too old -> field DISABLED, no free-text fallback. deliberately the opposite of the create-job dialog's employee pickers: an estimator id is rejected by the proc during the same submit, but a buyer id is written to Clerk and looks correct until someone tries to raise a PO.

no removal path, stays a GP-side task: eConnect exposes no proc, and a direct DELETE on POP00101 would bypass the business logic the eConnect-only rule exists to protect.
no edit path for an existing registration's description.

no DB migration.
installed relay needs a rebuild -> update through the relay app before the new ops go live.
until then the admin UI shows its disabled / "update the relay" state, existing #315 gating.

relay 304 pass, ruff clean.
backend ruff/format clean, new tests pass.
frontend 358 pass, lint 0 errors, tsc and build clean.

### DIFF
--- a/backend/app/graphql/user.graphql
+++ b/backend/app/graphql/user.graphql
@@ -19,7 +19,9 @@ type ClerkUser {
   # (app/auth.SHOP_ASSEMBLY_MANAGER_ROLE). Free-form in Clerk, so this is not an enum.
   roles: [String!]!
   # The GP BUYERID this account acts as (#216). Null until an admin links it; createPo and
-  # registerPoInGp enforce that the caller has one.
+  # registerPoInGp enforce that the caller has one. Since #409 the admin picks it from GP's live
+  # buyer master (Query.gpBuyersDetailed) rather than typing it, so an id that was never registered
+  # in GP can no longer be saved here - taPoHdr would have rejected it at PO time anyway (error 269).
   gp_buyer_id: String
   image_url: String!
 }
--- a/backend/app/graphql/user.graphql
+++ b/backend/app/graphql/user.graphql
@@ -20,8 +20,10 @@ type ClerkUser {
   roles: [String!]!
   # The GP BUYERID this account acts as (#216). Null until an admin links it; createPo and
   # registerPoInGp enforce that the caller has one. Since #409 the admin picks it from GP's live
-  # buyer master (Query.gpBuyersDetailed) rather than typing it, so an id that was never registered
-  # in GP can no longer be saved here - taPoHdr would have rejected it at PO time anyway (error 269).
+  # buyer master (Query.gpBuyersDetailed) rather than typing it, so the UI no longer offers an id GP
+  # never registered - taPoHdr would reject that one at PO time anyway (error 269). Note the guard is
+  # the picker, not this mutation: updateUserGpBuyerId still accepts any string, and rows written as
+  # free text before #409 are unmigrated, which is why the Buyers grid flags unregistered ids.
   gp_buyer_id: String
   image_url: String!
 }
--- a/backend/app/schemas/relay.py
+++ b/backend/app/schemas/relay.py
@@ -7,7 +7,14 @@ import strawberry
 
 from app.auth import require_admin, require_user
 from app.database import SessionLocal
-from app.errors import ConflictError, NotFoundError, RelayCallError, RelayUnavailableError, ValidationError
+from app.errors import (
+    ConflictError,
+    NotFoundError,
+    RelayCallError,
+    RelayUnavailableError,
+    ValidationError,
+    validation_error_from_relay,
+)
 from app.models.relay_install import RelayInstall
 from app.repositories import relay_repository
 from app.services import relay_adopt
@@ -297,10 +304,8 @@ class RelayMutations:
         except RelayCallError as e:
             # GP said no, or the id is already registered. Either way the relay words it better than a
             # generic failure would, and the detail body carries the proc + error state the dialog's
-            # error alert renders.
-            error = ValidationError(str(e.message))
-            error.detail = e.detail
-            raise error from e
+            # error alert renders - which is what validation_error_from_relay preserves.
+            raise validation_error_from_relay(e) from e
 
         # GP's stored row, read back from POP00101 by the relay rather than the request echoed back.
         # That row is what gpBuyersDetailed will serve on its next refetch, so answering with anything
--- a/backend/app/schemas/relay.py
+++ b/backend/app/schemas/relay.py
@@ -7,13 +7,14 @@ import strawberry
 
 from app.auth import require_admin, require_user
 from app.database import SessionLocal
-from app.errors import ConflictError, NotFoundError
+from app.errors import ConflictError, NotFoundError, RelayCallError, RelayUnavailableError, ValidationError
 from app.models.relay_install import RelayInstall
 from app.repositories import relay_repository
 from app.services import relay_adopt
 from app.services.relay_gateway import gateway as relay_gateway
 
 from .converters import (
+    gp_buyer_to_type,
     gp_cost_code_to_type,
     gp_customer_address_to_type,
     gp_customer_to_type,
@@ -26,6 +27,7 @@ from .converters import (
 )
 from .inputs import EnrollRelayInstallInput
 from .types import (
+    GpBuyer,
     GpCostCode,
     GpCustomer,
     GpCustomerAddress,
@@ -104,6 +106,19 @@ class RelayQueries:
         require_user(info)
         result = await relay_gateway.relay_call(company, "list_buyers")
         return result["buyers"]
+
+    @strawberry.field
+    async def gp_buyers_detailed(self, info: strawberry.Info, company: str) -> list[GpBuyer]:
+        """Registered GP buyers (POP00101) WITH their descriptions, via the connected relay, for the
+        admin screens that link a Nexus account to a GP buyer identity (#409).
+
+        Admin-gated where gp_buyers above is not, on the same reasoning as gp_employees: the bare id
+        list is ordinary working data every PO creator needs, but the descriptions are a staff roster
+        by another name ('Don Roberton', and in the production company every buyer's email), and the
+        only consumers are admin-only screens."""
+        require_admin(info)
+        result = await relay_gateway.relay_call(company, "list_buyers_detailed")
+        return [gp_buyer_to_type(b) for b in result["buyers"]]
 
     @strawberry.field
     async def gp_cost_codes(self, info: strawberry.Info, company: str, job: str) -> list[GpCostCode]:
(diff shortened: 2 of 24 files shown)


relay/src/ucnexus_relay/ops.py
```diff
@@ -260,6 +260,42 @@ def create_job_op(conn, *, company: str, request: models.CreateJobRequest) -> mo
...
+    if econnect.buyer_exists(conn, request.buyer_id):
+        raise RelayOpError(
+            "buyer_already_exists",
+            f"buyer '{request.buyer_id}' is already registered in GP company {company} (POP00101)",
+        )
+
+    econnect.create_buyer(conn, buyer_id=request.buyer_id, description=request.description)
+
+    created = next(
+        (b for b in econnect.list_buyers_detailed(conn) if b["buyer_id"] == request.buyer_id),
```
Xilous: read-back matches the buyer id with a python `==` against a full-table `list_buyers_detailed`, while every other lookup on this path normalizes through SQL: `buyer_exists` is `WHERE RTRIM(BUYERID) = ?` (case-insensitive under GP's default collation) and `create_job_op` reads back through the targeted `econnect.get_job`, also `RTRIM(...) = ?`.

consequences:
- any case normalization GP applies to BUYERID -> read-back misses -> `EConnectError("reported success but buyer ... is not in POP00101")` -> `_run_create_buyer` rolls back a create that actually succeeded, and the admin gets an internal-sounding error for a working write
- registering 'DONR' when 'donr' exists is refused by the pre-check (case-insensitive) but the read-back would compare case-sensitively, so the two gates disagree about identity
- one row costs a full scan of the buyer master

fix: add `econnect.get_buyer(conn, buyer_id)` shaped like `get_job` (`SELECT ... WHERE RTRIM(BUYERID) = ?`, arg stripped) and use it for the read-back.

frontend/src/modules/admin/BuyersPage.tsx
```diff
@@ -229,19 +257,80 @@ export default function BuyersPage() {
...
+            startIcon={<Plus size={18} strokeWidth={1.75} />}
+            onClick={() => setRegisterOpen(true)}
+            disabled={gpBuyers.unavailable}
+            sx={{ flexShrink: 0 }}
+          >
+            Register GP Buyer
+          </Button>
+        </Box>
+
+        {gpBuyers.unavailable ? (
```
Xilous, line 286: this panel has no `relayStatus === null` branch, but `GpBuyerSelect` does ('Checking the GP relay…').

`useGpBuyers` sets `unavailable = !relayConnected || error`, and `relayConnected` is `relay.connected === true`, so it is false while the first `relayStatus` round-trip is in flight. every load of the Buyers page therefore renders the warning alert 'The GP relay is not connected, so the buyer master cannot be read or added to' and disables Register, then flips, even when the relay is up the whole time.

same three-way state, two different renderings. give the panel the `relayStatus === null` case GpBuyerSelect already has.

frontend/src/modules/admin/RegisterGpBuyerDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const handleClose = useCallback(() => {
+    reset();
+    onClose();
+  }, [reset, onClose]);
+
+  const [createGpBuyer, { loading }] = useMutation<{ createGpBuyer: { buyerId: string } }>(CREATE_GP_BUYER, {
+    // The dropdowns this feeds read gpBuyersDetailed, so they have to see the new row before the
+    // caller's onRegistered tries to select it.
+    refetchQueries: [{ query: GET_GP_BUYERS_DETAILED, variables: { company } }],
+    awaitRefetchQueries: true,
```
Xilous: `awaitRefetchQueries: true` folds the refetch's failure into the mutation's. apollo resolves the mutate promise only after the refetch settles, so a refetch that errors rejects the mutation -> `onError` fires -> `onCompleted` never runs.

failure: relay drops right after `taCreateBuyer` commits (the window a GP write is slowest and most likely to lose the socket). the buyer IS registered, the admin is shown 'Could not register the buyer', `onRegistered` never fires, and the dropdown never gets the row.

the stated reason for awaiting - the dropdowns must see the row before `onRegistered` selects it - is not load-bearing: `GpBuyerSelect.selected` already falls back to `{ buyerId: value, description: null }` for an id not in the list, so selection renders correctly whether or not the refetch has landed.

frontend/src/modules/admin/RegisterGpBuyerDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+    onCompleted: (data) => {
+      showToast(`Buyer '${data.createGpBuyer.buyerId}' registered in GP`, 'success');
+      onRegistered?.(data.createGpBuyer.buyerId);
+      handleClose();
+    },
+    onError: (err) => setGpError(extractGpError(err) ?? { message: err.message }),
```
Xilous: the error path never refetches the buyer list, which dead-ends the exact retry the relay's pre-check was built for.

sequence: register -> relay commits -> reply lost (timeout) -> admin retries -> `create_buyer_op` answers `buyer_already_exists` -> shown here as a red GpErrorAlert. `refetchQueries` only runs on success, so `gpBuyersDetailed` still holds the pre-write list and the buyer that IS in GP is still not pickable. the admin has to reload the page to get out.

`create_gp_job` handles the same class by adopting on `job_already_exists` instead of surfacing it. at minimum refetch `GET_GP_BUYERS_DETAILED` when the relay error code is `buyer_already_exists`.

frontend/src/modules/admin/useGpBuyers.ts
```diff
@@ -0,0 +1,71 @@
...
+  const unsupported = isRelayOpUnsupported(error);
+
+  return {
+    buyers,
+    loading,
+    company,
+    relayConnected,
+    relayStatus: relay.connected,
+    unsupported,
+    unavailable: !relayConnected || Boolean(error),
```
Xilous, line 65: apollo's `error` is sticky and nothing retries this query, so one transient read failure disables the buyer field for the life of the page.

`GET_GP_BUYERS_DETAILED` has no `pollInterval` and the returned `refetch` is called by nobody. `useRelayStatus` polls every 10s, but a relay that stays connected through a failed read (GP timeout, a SQL error) leaves `error` set forever -> `unavailable` true -> field disabled, Register disabled, no way back except a remount.

the skip chain only re-runs the query on a disconnect->reconnect transition, which is not the failure being described here.

backend/app/schemas/relay.py
```diff
@@ -238,6 +253,65 @@ async def suggest_vendor_for_manufacturer(
...
+
+        try:
+            result = await relay_gateway.relay_call(
+                target, "create_buyer", {"buyer_id": cleaned_id, "description": cleaned_description}
+            )
+        except RelayCallError as e:
+            # GP said no, or the id is already registered. Either way the relay words it better than a
+            # generic failure would, and the detail body carries the proc + error state the dialog's
+            # error alert renders.
+            error = ValidationError(str(e.message))
```
Xilous: this is `app/schemas/project.py::_validation_error_from` re-implemented verbatim, comment and all:

```python
error = ValidationError(str(e.message))
error.detail = e.detail
return error
```

two copies of the one rule that keeps `extensions.relayError` alive across the RelayCallError -> ValidationError conversion. hoist it (app/errors.py is the natural home, both schema modules already import from it) and call it from both.

backend/tests/test_create_gp_buyer.py
```diff
@@ -0,0 +1,168 @@
...
+
+    assert "already registered" in str(exc.value)
+    # The detail body rides along: ErrorHandlerExtension publishes it as extensions.relayError, which
+    # is what the dialog's error alert renders.
+    assert exc.value.detail == detail
+
+
+def test_a_gp_refusal_surfaces_with_its_detail_intact(monkeypatch):
+    _as_admin(monkeypatch)
+    detail = {"error": "econnect_error", "message": "taCreateBuyer failed", "proc": "taCreateBuyer", "error_state": 1}
```
Xilous: this detail body is a shape the relay never produces. `relay/errors.py::error_body` always returns `{error, message, context: {...}}` and `econnect_error_body` puts `proc` / `error_state` / `error_description` inside `context`.

so `exc.value.detail["proc"]` asserts against a top-level key that only exists in this fixture, and `GpErrorAlert` reads `error.relay?.context?.proc`. the test passes while proving nothing about the payload the alert actually renders.

nest them under `context` to match `error_body`.

backend/app/graphql/user.graphql
```diff
@@ -19,7 +19,9 @@ type ClerkUser {
   # (app/auth.SHOP_ASSEMBLY_MANAGER_ROLE). Free-form in Clerk, so this is not an enum.
   roles: [String!]!
   # The GP BUYERID this account acts as (#216). Null until an admin links it; createPo and
-  # registerPoInGp enforce that the caller has one.
+  # registerPoInGp enforce that the caller has one. Since #409 the admin picks it from GP's live
+  # buyer master (Query.gpBuyersDetailed) rather than typing it, so an id that was never registered
+  # in GP can no longer be saved here - taPoHdr would have rejected it at PO time anyway (error 269).
```
Xilous: 'an id that was never registered in GP can no longer be saved here' is not true of this field. `update_user_gp_buyer_id` (app/schemas/user.py:30) takes `gp_buyer_id: str | None` straight to `user_repository.update_user_gp_buyer_id` with no GP check and no `require_admin`. the picker is the only thing enforcing it, and it is client-side.

existing rows written as free text are also unmigrated, which is why BuyersPage needs the unregistered-warning triangle at all.

either word the comment as a UI guarantee or put the check on the mutation.

frontend/src/modules/admin/GpBuyerSelect.tsx
```diff
@@ -0,0 +1,136 @@
...
+          <TextField
+            {...params}
+            label={label}
+            size="small"
+            helperText={unavailable ? unavailableText : helperText}
+          />
+        )}
+        sx={sx}
+      />
+      <RegisterGpBuyerDialog
```
Xilous, line 134: `GpBuyerSelect` owning a `RegisterGpBuyerDialog` means every page embedding it gets a hidden second copy. BuyersPage mounts one here (inside the edit dialog's select) and one at page level for the panel's Register button, so two dialogs with two live `useMutation` + refetch configs sit in the tree at once.

mounting it only when open (`{registerOpen && <RegisterGpBuyerDialog ... />}`) is the cheap version; hoisting it to the page and passing an `onRegisterRequest` callback is the clean one.

frontend/src/modules/admin/useGpBuyers.ts
```diff
@@ -0,0 +1,71 @@
...
+    error: error ?? null,
+    refetch: () => {
```
Xilous, line 67: `refetch` is dead: returned, wrapped in a closure, and called by no consumer. either wire it (the Buyers page warning alert is the obvious place for a Retry, which would also fix the sticky-error lockout above) or drop it from the interface.

frontend/src/modules/admin/GpBuyerSelect.tsx
```diff
@@ -0,0 +1,136 @@
...
+  label = 'GP Buyer ID',
+  disabled,
+  fullWidth,
+  sx,
+  helperText,
+}: GpBuyerSelectProps) {
+  const [registerOpen, setRegisterOpen] = useState(false);
+  const { buyers, loading, company, relayConnected, relayStatus, unsupported, unavailable } = state;
+
+  const options: (GpBuyerOption | { buyerId: typeof REGISTER_NEW_ID; description: null })[] = useMemo(
```
Xilous: a held value that is not in `buyers` is never added to `options`, so MUI logs `The value provided to Autocomplete is invalid. None of the options match with "..."` on every render of that path - which is precisely the relay-down-with-a-stored-id case this component exists to support, and the one the UserManagementPage test exercises.

append the synthetic `selected` to `options` when `buyers` does not already contain it.

frontend/src/modules/admin/GpBuyerSelect.tsx
```diff
@@ -0,0 +1,136 @@
...
+                isRegister
+                  ? { fontWeight: 600, borderTop: 1, borderColor: 'divider', color: 'text.primary' }
+                  : monoSx
+              }
+            >
+              {isRegister ? '+ Register new GP buyer…' : buyerLabel(option)}
+            </Box>
+          );
+        }}
+        renderInput={(params) => (
```
Xilous, line 121: the field lost its monospace treatment. both call sites previously carried `'& .MuiInputBase-input': { fontFamily: FONT_MONO }` (UserManagementPage dropped the `FONT_MONO` import entirely in this PR), and `renderOption` still applies `monoSx` to the list rows, as does the BuyersPage grid cell.

so the same BUYERID renders mono in the dropdown and in the grid, and in the body face once selected.

relay/src/ucnexus_relay/models.py
```diff
@@ -153,6 +153,48 @@ class BuyersResponse(BaseModel):
...
+
+class CreateBuyerRequest(BaseModel):
+    """Max lengths are taCreateBuyer's own parameter widths - char(15) BUYERID, char(30) DSCRIPTN -
+    so an over-length value is rejected here rather than silently truncated on the way into GP.
+
+    Note DSCRIPTN is char(31) on POP00101 but char(30) on the proc; the proc is the narrower gate and
+    therefore the binding one."""
+
+    company: str
+    buyer_id: str = Field(..., max_length=15)
```
Xilous: pydantic runs `Field(max_length=...)` before a `mode="after"` validator, so the widths are checked against the untrimmed value while the docstring frames them as taCreateBuyer's parameter widths.

`"  fifteenchars15  "` is rejected as `invalid_payload` even though it normalizes to a legal 15-char BUYERID. unreachable from the backend today only because `create_gp_buyer` strips first, but this model is the relay's own gate and the HTTP path does not.

move the length checks into `normalize`, after the strips.

Xilous (commented):
code review: 13 findings on the #409 buyer-registration slice.

Issue #407, opened 2026-07-29
Railway MCP server holds a stale token while the CLI stays authenticated


Xilous:
this project's Claude Code Railway MCP server holds a stale token. every mcp__Railway__* tool call fails with "Unauthorized. Please run `railway login` again". railway CLI is fine, railway whoami succeeds - the CLI refreshes its token on use; the MCP server authenticated once, never refreshes.

agent sessions fall back to the CLI plus raw GraphQL against backboard.railway.com, accessToken read from ~/.railway/config.json. that token expires mid-session, needs a CLI poke to refresh.

three fix directions exist:
- re-authenticate the MCP server
- configure it to read the CLI's config so it picks up refreshed tokens
- check whether a newer Railway MCP server version handles token refresh

Xilous:
the MCP server is the CLI itself; the global entry in ~/.claude.json is {"command": "railway", "args": ["mcp"]}. it reads ~/.railway/config.json once at spawn. CLI commands refresh the token on use and write it back -> long-lived mcp process never re-reads -> token stale mid-session. a session spawned against an already-expired config is unauthorized from the first call.

whoami and list_projects succeeded this session - the process spawned fresh against a recently-refreshed token. failure is intermittent by process lifetime.

installed CLI 4.66.0, latest v5.30.1. the MCP server ships inside the CLI binary, so upgrading is the only vector for an actual token-refresh fix.

1. upgrade CLI to v5.30.1 via the same install channel. confirm railway whoami and the MCP tools afterwards.
2. let a session sit past token expiry, then call an MCP tool. if it succeeds, done.
3. if v5 still holds a stale token, standing recovery on the first Unauthorized from an MCP tool is railway whoami (refreshes the token on disk) -> /mcp reconnect Railway (respawns the process, re-reads the config). file an upstream issue on railwayapp/cli. raw-GraphQL-against-backboard fallback stays last resort only.
4. update "Railway CLI is installed (v4.25.3)" in the global CLAUDE.md to the post-upgrade version and add the one-line recovery.

no repo change; machine/tooling config only.

Xilous:
CLI on v5.30.1, was 4.66.0 at filing. the MCP server ships inside the CLI binary, so the upgrade was the only real fix vector.

`railway whoami` succeeds this session. these MCP tools work:
list_projects
list_variables
set_variables

no Unauthorized since the upgrade. whether v5.30.1 still holds a stale token past expiry is unobserved, needs an idle session.

[...]

PR #413, opened 2026-07-29
feat(reset): preserve setup tables across reset, release stranded IN_PO hardware

### LINKED ISSUE #402: Repair migration for HardwareItem rows stranded IN_PO by pre-#398 PO cancels
#398 stops cancel_po and the register-time line-edit path stranding HardwareItem rows in IN_PO against dead POs. go-forward only, rows stranded before it landed remain in the db:
- double-count required_quantity in project purchasing progress rollups
- inflate the admin Hardware Items stat
- (opening, product, category, leaf) keys block finalize from recreating those items as AVAILABLE

fix with a data migration or one-off admin action releasing two populations to state=AVAILABLE with po_line_item_id nulled - the exact end-state #398 now produces at cancel time:
- rows whose po_line_item_id points at a line item of a CANCELLED or soft-deleted PO
- rows state=IN_PO with po_line_item_id NULL (register-edit orphans)

low urgency on current dev data, do before any real data exists - the zombies are invisible in every UI and only show up as subtly wrong numbers.

### LINKED ISSUE #410: reset-data: preserve non-dev config tables, re-link buyer assignments, sync GP jobs after rebuild
POST /admin/reset-data (the "DevAction: drop and rebuild schema" button, backend/main.py) drops the entire public schema and restores only relay_installs (#352). admin-entered config is lost.

these tables persist across a reset (snapshot before drop -> restore after alembic upgrade head, same pattern as relay_installs):
relay_installs (already preserved)
warehouses
buyer_assignments + buyer_assignment_projects (#216)
manufacturer_vendor_map (#232)
po_document_settings (#230; get-or-create reseeds defaults, but admin edits are lost today)

everything else stays wiped as dev data:
projects
openings
opening items
hardware
POs
inventory locations
reservations
pull requests
pick lines
receiving
shipping
shop assembly
stock items
notifications
audit log
GP outbox
idempotency rows

local vendors table also wiped on purpose. contact fields duplicate GP (PM00200/PM00300/SY01200); only Nexus-only column is free-text notes. if draft POs need vendor contact display, widen the relay's PM00200 read (currently only VENDORID/VENDNAME/VNDCLSID/VENDSTTS/CURNCYID) - separate follow-up issue, not this one.

gp_job_sync (#380) re-adopts every GP job after the reset:
5-minute poll
relay reconnect
admin Sync from GP button

reset endpoint additionally calls gp_job_sync.run_once() at the end when a relay is connected, skips silently when not.

GP vendors, buyers, cost codes, tax details are fetched live over the relay per request, never persisted - nothing to reload.

buyer_assignment_projects FKs projects.id; re-adopted projects get new UUIDs, so a naive row restore breaks. snapshot each assignment's linked GP job numbers (projects.schedule_id at snapshot time) -> restore buyer_assignments rows -> post-reset sync pass -> re-attach project links by job number. links to jobs gone from GP are dropped.

Nexus-owned project fields (ship-to address, GC contact info) do not come back after re-adoption. acceptable for a dev reset.

generalize the existing relay_installs snapshot/restore in reset_data into a preservation list of tables. plain SQL snapshot to mappings before the drop -> restore after upgrade head. tolerate absent tables on a half-migrated DB, same as the relay_installs code today. add the buyer-assignment job-number re-link after the sync pass.

### LINKED ISSUE #411: derive /admin/reset-data relay_installs snapshot columns from RelayInstall model metadata
/admin/reset-data preserves relay_installs across the drop-and-rebuild so a dev wipe never orphans the enrolled relay. snapshot SELECT and restore INSERT (backend/main.py, reset_data) each hardcode the 13 column names in raw SQL.

nothing enforces sync with the RelayInstall model. a migration that adds a relay_installs column without updating reset_data still resets fine, the new column silently comes back NULL on every preserved row. depending on the column that is a broken credential or lost audit data, discovered much later.

derive the column list from RelayInstall.__table__.columns. alternatively (or additionally) a test comparing the hardcoded list against the model catches the drift in CI.

Xilous:
three issues, one branch.

#411 + #410 - reset_data preservation

preservation list of models in app/services/reset_preservation.py, restored in order:
RelayInstall
Warehouse
BuyerAssignment
ManufacturerVendorMap
PODocumentSettings

column list per model = Model.__table__.columns intersected with the live table's columns, in model declaration order. the intersection keeps a half-migrated DB from crashing the SELECT, same tolerance the has_table guard gives.

statements built through SQLAlchemy Core, not string SQL - deviation from the plan. BuyerAssignment.cost_codes and four PODocumentSettings columns are JSON; psycopg2 adapts a Python list to a Postgres array -> a text() bind param fails the restore with "column is of type json but expression is of type text[]". relay_installs alone never hit this, it has no JSON column. Core also falls back to the model's Python-side default for a column the snapshot is missing, instead of inserting NULL.

buyer_assignment_projects cannot be row-restored - FKs projects.id, re-adopted projects get new UUIDs. snapshotted as (buyer_assignment_id, job_number) pairs, re-linked by job number after the sync.

GP job sync after restore via new gp_job_sync.run_once_blocking(). _loop captured at run_forever start, next to _wake_event. reset_data is a sync def in the threadpool -> hands run_once() to that loop via asyncio.run_coroutine_threadsafe -> blocks on the result, 90s timeout. sync disabled, no relay, relay vanished mid-pass, or timeout all return None and skip.

response JSON now carries:
per-table preserved counts
relay_installs_preserved, kept
jobs_adopted, null when no pass ran
buyer_links_restored
buyer_links_dropped
message, carrying the summary

PODocumentSettings restore lands before its get-or-create seed.

crypto.py docstring updated - dropping secret_encrypted no longer has to touch the reset-data SQL.

#402 - migration 073

two idempotent UPDATEs, no-op on clean data, applied on every deploy by the entrypoint's alembic upgrade head.

rows pointing at a CANCELLED or soft-deleted PO -> state AVAILABLE, po_line_item_id NULL, no state filter
IN_PO rows with po_line_item_id NULL -> state AVAILABLE

downgrade is a no-op.

#407 is not in this PR, no repo change. Railway CLI upgraded 4.66.0 -> 5.30.1 on this machine, global CLAUDE.md updated with the version and the whoami -> /mcp reconnect recovery.

14 new tests in backend/tests/test_reset_preservation.py, no Postgres needed. ruff check and ruff format clean. alembic heads reports 073 as the single head.

Closes #410, closes #411, closes #402

### DIFF
--- a/backend/alembic/versions/073_release_stranded_in_po_hardware.py
+++ b/backend/alembic/versions/073_release_stranded_in_po_hardware.py
@@ -1,0 +1,63 @@
+"""Release hardware stranded IN_PO by a dead purchase order (#402)
+
+Revision ID: 073
+Revises: 072
+Create Date: 2026-07-29
+
+Before #398, cancelling a PO left its hardware rows behind: the PO went to CANCELLED (or was
+soft-deleted via `deleted_at`) but the `hardware_items` rows it had claimed kept `state = 'IN_PO'`
+and kept pointing at the now-dead `po_line_items` row. `cancel_po` releases them properly now -
+back to `state = 'AVAILABLE'` with `po_line_item_id = NULL` - but that only fixes cancels from here
+on. Every environment still carries the rows the old cancels stranded, and nothing in the app can
+reach them: they are attached to a PO no screen will ever open again.
+
+Three symptoms come out of that, all of them from the same stranded rows:
+
+- required_quantity rollups double-count. The stranded row still reads as ordered, and the
+  replacement the user raised afterwards reads as ordered too.
+- the admin Hardware Items stat is inflated by exactly the stranded count.
+- finalize cannot recreate the items as AVAILABLE, because the stranded row already holds the
+  (opening, product, category, leaf) key it needs.
+
+This is a data migration rather than a one-off admin action on purpose. The backend entrypoint runs
+`alembic upgrade head` on every deploy, so it repairs each environment exactly once, needs no UI,
+and no-ops from then on.
+
+Both statements are idempotent and safe on an empty database. `downgrade()` is a no-op: the pre-#398
+state is a bug, and there is no record of which rows were stranded to put back.
+"""
+
+from alembic import op
+
+revision = "073"
+down_revision = "072"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Anything pointing at a line item of a cancelled or soft-deleted PO, whatever state it claims to
+    # be in. No `state` filter here on purpose: a dead PO cannot own hardware, so both fields get
+    # normalized regardless of how the row got there.
+    op.execute(
+        """
+        UPDATE hardware_items SET state = 'AVAILABLE', po_line_item_id = NULL
+        WHERE po_line_item_id IN (
+          SELECT li.id FROM po_line_items li JOIN purchase_orders po ON po.id = li.po_id
+          WHERE po.status = 'CANCELLED' OR po.deleted_at IS NOT NULL)
+        """
+    )
+
+    # Register-edit orphans: the link was cleared but the state was not, so the row is claimed by a
+    # PO that is not named anywhere.
+    op.execute(
+        """
+        UPDATE hardware_items SET state = 'AVAILABLE'
+        WHERE state = 'IN_PO' AND po_line_item_id IS NULL
+        """
+    )
+
+
+def downgrade() -> None:
+    """No-op. Re-stranding the rows would restore the defect, and which rows were stranded is not
+    recorded anywhere."""
--- a/backend/alembic/versions/073_release_stranded_in_po_hardware.py
+++ b/backend/alembic/versions/073_release_stranded_in_po_hardware.py
@@ -23,9 +23,23 @@ This is a data migration rather than a one-off admin action on purpose. The back
 `alembic upgrade head` on every deploy, so it repairs each environment exactly once, needs no UI,
 and no-ops from then on.
 
-Both statements are idempotent and safe on an empty database. `downgrade()` is a no-op: the pre-#398
-state is a bug, and there is no record of which rows were stranded to put back.
+Both statements are idempotent. `downgrade()` is a no-op: the pre-#398 state is a bug, and there is
+no record of which rows were stranded to put back.
+
+The empty-table guard in `upgrade()` is not an optimization. `AVAILABLE` is added to
+`hardware_item_state` by `ALTER TYPE ... ADD VALUE` back in 026, and PostgreSQL refuses to *use* a
+label added by a transaction that has not committed yet. `alembic/env.py` wraps the entire
+`upgrade head` in one transaction, so on a fresh database 026 and this migration share it and the
+UPDATE cannot even be parsed. 071 hit the same wall and got out of it by comparing `source::text`,
+which works for a read but not here - an UPDATE has to produce a real enum value, and every route to
+one goes through the same check. Skipping is the honest answer rather than a dodge: a database with
+no `hardware_items` rows has nothing stranded in the first place. On a deployed database 026
+committed deploys ago, so the label is long since safe and the repair runs normally.
+
+The reads still cast to text (071's pattern), which keeps them clear of the same trap.
 """
+
+import sqlalchemy as sa
 
 from alembic import op
 
@@ -36,6 +50,11 @@ depends_on = None
 
 
 def upgrade() -> None:
+    # Nothing to repair, and on a fresh database the statements below could not run anyway - see the
+    # module docstring for why the enum label is off limits inside this transaction.
+    if op.get_bind().execute(sa.text("SELECT 1 FROM hardware_items LIMIT 1")).first() is None:
+        return
+
     # Anything pointing at a line item of a cancelled or soft-deleted PO, whatever state it claims to
     # be in. No `state` filter here on purpose: a dead PO cannot own hardware, so both fields get
     # normalized regardless of how the row got there.
@@ -44,7 +63,7 @@ def upgrade() -> None:
         UPDATE hardware_items SET state = 'AVAILABLE', po_line_item_id = NULL
         WHERE po_line_item_id IN (
           SELECT li.id FROM po_line_items li JOIN purchase_orders po ON po.id = li.po_id
-          WHERE po.status = 'CANCELLED' OR po.deleted_at IS NOT NULL)
+          WHERE po.status::text = 'CANCELLED' OR po.deleted_at IS NOT NULL)
         """
     )
 
@@ -53,7 +72,7 @@ def upgrade() -> None:
     op.execute(
         """
         UPDATE hardware_items SET state = 'AVAILABLE'
-        WHERE state = 'IN_PO' AND po_line_item_id IS NULL
+        WHERE state::text = 'IN_PO' AND po_line_item_id IS NULL
         """
     )
 
--- a/backend/app/services/gp_job_sync.py
+++ b/backend/app/services/gp_job_sync.py
@@ -33,7 +33,19 @@ logger = logging.getLogger(__name__)
 # covers "I just made one and want it now".
 POLL_SECONDS = 300.0
 
+# How long /admin/reset-data waits for its one forced sync pass. Generous next to relay_call's own 30s
+# because this pass also writes a project row per GP job, and the reset is a deliberate manual action
+# that nobody is timing - overshooting costs a few seconds, giving up early costs every buyer link.
+RESET_SYNC_TIMEOUT_SECONDS = 90.0
+
 _wake_event: asyncio.Event | None = None
+
+# The loop `run_forever` was scheduled on, which is the loop the relay websocket lives on. Captured so
+# a *synchronous* caller in the threadpool can still run a sync pass: `/admin/reset-data` is a sync def
+# and must re-adopt every GP job right after it rebuilds the schema, but `run_once` talks to the relay
+# over that socket and may only be awaited on its own loop (#410). None until the lifespan task starts,
+# and it never starts when `enabled()` is False - callers treat that as "no sync available" and skip.
+_loop: "asyncio.AbstractEventLoop | None" = None
 
 
 def enabled() -> bool:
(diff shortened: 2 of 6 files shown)


backend/app/services/reset_preservation.py
```diff
@@ -0,0 +1,158 @@
...
+def restore(conn: Connection, snap: ResetSnapshot) -> dict[str, int]:
+    """Insert the snapshotted rows back into the rebuilt schema, original UUIDs intact.
+
+    Runs before anything can lazily seed a default row - `PODocumentSettings` in particular is read
+    through a get-or-create helper, so restoring first is what stops a defaults row being seeded
+    alongside the admin-edited one."""
+    for model in PRESERVED_MODELS:
+        rows = snap.rows.get(model.__table__.name)
+        if rows:
+            conn.execute(insert(model.__table__), rows)
```
Xilous: blocking - this insert collides with the migration seeds on every reset past 033.

alembic upgrade head replays 033_add_warehouses -> seeds warehouses with ('Warden','WRD') and ('VP','VP'). the snapshot taken before the drop holds those same two rows under their old UUIDs, so this executemany violates uq_warehouses_name and uq_warehouses_code.

restore is one transaction with conn.commit() after the loop, RelayInstall inserted first -> IntegrityError rolls the relay rows back too. /admin/reset-data 500s, schema already dropped and rebuilt, relay_installs gone - the exact GP outage this feature was written to prevent. fires on the first reset on any environment.

clear the table before restoring it, so the snapshot is authoritative over whatever the rebuild seeded.

backend/app/services/reset_preservation.py
```diff
@@ -0,0 +1,158 @@
...
+    return snap.counts
```
Xilous: one table failing discards the rest. this loop runs in a single transaction main.py commits after restore returns -> IntegrityError on any later model rolls back the relay_installs rows inserted first.

wrap each table in conn.begin_nested(), and return the count actually restored per table instead of snap.counts, so a bad table costs its own rows and nothing else.

backend/main.py
```diff
@@ -347,24 +338,42 @@ def reset_data(request: Request):
...
-                row,
-            )
+        preserved = reset_preservation.restore(conn, snap)
+        conn.commit()
+
+    # Re-adopt every GP job before re-linking, not after: the links are matched on job number, so a
+    # projects table that is still empty here drops every one of them. Skipped silently when no relay
+    # is connected - see run_once_blocking; the buyer rows and their cost codes survive either way, and
+    # the background poll re-adopts the projects later (it just has nothing left to re-link them to).
+    sync_result = gp_job_sync.run_once_blocking()
```
Xilous, line 348: snap.buyer_project_pairs only exists inside this request. run_once_blocking returns None (no relay connected, sync disabled, 90s timeout) -> relink_buyer_projects runs against an empty projects table -> drops every pair -> pairs gone. the background poll re-adopts the projects a poll interval later, nothing re-links them; buyer/project assignments lost with manual re-entry as the only recovery.

persist the pairs, or defer the relink to the sync pass that actually adopts the projects.

backend/app/services/gp_job_sync.py
```diff
@@ -102,11 +114,39 @@ async def run_once() -> tuple[int, int]:
...
+    `/admin/reset-data` is a sync def, so FastAPI runs it in the threadpool - but it has to re-adopt
+    every GP job the moment it has rebuilt the schema. Waiting out POLL_SECONDS is not an option there:
+    the buyer/project links the reset restores are matched by job number, so until the projects are
+    back there is nothing to match and every link drops. `run_once` awaits the relay socket and may
+    only be awaited on the loop that owns it, hence the hand-off.
+
+    Returns (total, adopted), or None when no pass could run - sync disabled, no relay connected, the
+    relay went away mid-pass, or it did not answer in time. Every one of those is a skip rather than a
+    failure: a reset must not fail because GP happened to be unreachable."""
+    loop = _loop
```
Xilous, line 134: two problems with this guard.

_loop is set in run_forever and never cleared. lifespan task exits or is cancelled while the loop object is still open but no longer running -> run_coroutine_threadsafe accepts the coroutine -> nothing ever runs it -> .result(90) blocks the reset request for the full timeout, then logs a false "pass did not finish" warning. clear _loop in a finally on run_forever, and check loop.is_running() here.

enabled() is not checked. the docstring says a disabled sync returns None; that only holds because main.py gates the task creation on enabled(). the kill switch and the loop handle are separate facts - check the switch explicitly.

backend/main.py
```diff
@@ -347,24 +338,42 @@ def reset_data(request: Request):
...
+
+    with engine.connect() as conn:
+        links_restored, links_dropped = reset_preservation.relink_buyer_projects(conn, snap.buyer_project_pairs)
         conn.commit()
 
+    # The frontend alerts `message` verbatim, so it carries the summary; the fields below are for
+    # anyone reading the response itself.
+    parts = [f"{count} {table}" for table, count in preserved.items() if count]
```
Xilous: raw table names in text the frontend alerts verbatim: "Preserved 1 relay_installs, 2 warehouses". map the table name to a human label and agree in number, same as #405.

backend/app/services/reset_preservation.py
```diff
@@ -0,0 +1,158 @@
...
+    for model in PRESERVED_MODELS:
+        table = model.__table__
+        if not inspector.has_table(table.name):
+            continue
+        stmt = snapshot_statement(model, [c["name"] for c in inspector.get_columns(table.name)])
+        if stmt is None:
+            continue
+        snap.rows[table.name] = [dict(r) for r in conn.execute(stmt).mappings()]
+
+    if inspector.has_table(buyer_assignment_projects.name) and inspector.has_table(Project.__table__.name):
```
Xilous: this has_table check is independent of whether buyer_assignments was snapshotted. buyer_assignments snapshot skipped (table absent, or snapshot_statement returned None) while buyer_assignment_projects is present -> pairs still collected -> relink_buyer_projects inserts FK-violating rows -> 500 after the schema is already rebuilt and committed.

gate on BuyerAssignment.__table__.name in snap.rows, and filter the relink to buyer ids that actually came back.

backend/app/services/reset_preservation.py
```diff
@@ -0,0 +1,158 @@
...
+        ).select_from(buyer_assignment_projects.join(projects, projects.c.id == buyer_assignment_projects.c.project_id))
+        snap.buyer_project_pairs = [dict(r) for r in conn.execute(pairs).mappings()]
+
+    return snap
+
+
+def restore(conn: Connection, snap: ResetSnapshot) -> dict[str, int]:
+    """Insert the snapshotted rows back into the rebuilt schema, original UUIDs intact.
+
+    Runs before anything can lazily seed a default row - `PODocumentSettings` in particular is read
```
Xilous: "runs before anything can lazily seed a default row" does not hold. the app is serving traffic throughout and command.upgrade has already committed by the time this runs -> a concurrent GraphQL read reaching po_document_settings_repository.get_settings seeds a defaults row in the window before the restore. the restore then adds a second one, and nothing in the schema enforces the single-row invariant.

delete the table's contents before restoring.

backend/alembic/versions/073_release_stranded_in_po_hardware.py
```diff
@@ -0,0 +1,82 @@
...
+revision = "073"
+down_revision = "072"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Nothing to repair, and on a fresh database the statements below could not run anyway - see the
+    # module docstring for why the enum label is off limits inside this transaction.
+    if op.get_bind().execute(sa.text("SELECT 1 FROM hardware_items LIMIT 1")).first() is None:
```
Xilous, line 55: the guard does not cover the case the docstring describes. the unsafe-enum failure fires when 026's ALTER TYPE ... ADD VALUE runs in the same transaction as this UPDATE - a database sitting below revision 026 that has hardware_items rows. that database has rows -> guard passes -> deploy hard-fails on "unsafe use of new value AVAILABLE". empty-table is a proxy for "026 already committed", not the same condition. every live environment is far past 026, practical risk low. the docstring claims a guarantee the code does not give.

backend/app/services/gp_job_sync.py
```diff
@@ -102,11 +114,39 @@ async def run_once() -> tuple[int, int]:
...
+    if loop is None or not relay_gateway.connected:
+        return None
+    try:
+        return asyncio.run_coroutine_threadsafe(run_once(), loop).result(timeout=timeout)
+    except (RelayUnavailableError, RelayTimeoutError) as e:
+        logger.info("gp job sync: relay unavailable during reset (%s); projects not re-adopted", e.message)
+    except TimeoutError:
```
Xilous, line 141: in 3.11 asyncio.TimeoutError is TimeoutError, so this branch also catches a timeout raised inside run_once and logs it as "pass did not finish within 90s during reset" when the wait never actually expired.

check future.done() before choosing the message, or keep the future in a local and branch on it.

backend/tests/test_reset_preservation.py
```diff
@@ -0,0 +1,109 @@
...
+
+    assert json_columns, "expected at least one JSON column among the preserved models"
+
+
+def test_snapshot_statement_targets_the_models_own_table():
+    """Sanity check that the derived SELECT reads the table it claims to - a wrong FROM would preserve
+    the wrong rows silently."""
+    for model in PRESERVED_MODELS:
+        stmt = snapshot_statement(model, _all_columns(model))
+        assert sa_inspect(model).local_table.name in str(stmt)
```
Xilous: this assertion cannot fail. the compiled SQL prefixes every selected column with its own table name -> local_table.name in str(stmt) is true for any single-table select, including a wrong one.

assert on the FROM clause instead:

```suggestion
        assert stmt.get_final_froms() == [sa_inspect(model).local_table]
```

backend/app/services/reset_preservation.py
```diff
@@ -0,0 +1,158 @@
...
+    rows: dict[str, list[dict]] = field(default_factory=dict)
+    # (buyer_assignment_id, job_number) pairs - see the module docstring for why these are not rows.
+    buyer_project_pairs: list[dict] = field(default_factory=list)
+
+    @property
+    def counts(self) -> dict[str, int]:
+        return {table: len(rows) for table, rows in self.rows.items()}
+
+
+def preserved_columns(model, live_column_names) -> list[str]:
```
Xilous: model and live_column_names are unannotated in a module that annotates everything else, and snapshot_statement below has no return annotation. want

```suggestion
def preserved_columns(model: type[Base], live_column_names: Iterable[str]) -> list[str]:
```

and `-> Select | None` on snapshot_statement.

backend/app/services/reset_preservation.py
```diff
@@ -0,0 +1,158 @@
...
+    database: a table that isn't there yet is simply not preserved, which is exactly the state a reset
+    exists to clear."""
+    inspector = sa_inspect(conn)
+    snap = ResetSnapshot()
+
+    for model in PRESERVED_MODELS:
+        table = model.__table__
+        if not inspector.has_table(table.name):
+            continue
+        stmt = snapshot_statement(model, [c["name"] for c in inspector.get_columns(table.name)])
```
Xilous, line 126: has_table per model
get_columns per model
two more has_table calls for the pairs

-> 12 round trips to managed Postgres before the drop. inspector.get_table_names() once gives the same answer for every membership test.

Xilous (commented):
12 findings, 1 blocking.

blocking: Warehouse in PRESERVED_MODELS collides with 033's ('Warden','WRD')/('VP','VP') seed on uq_warehouses_name -> restore raises -> the whole restore transaction rolls back, including relay_installs -> reset 500s with the schema already rebuilt. deterministic on the first reset on any environment past 033.

fixes applied locally for 1, 2, 4, 5, 6, 7, 10, 11, 12. left as reported: 3 (buyer links lost for good when the sync cannot run), 8 (073's empty-table guard is not the condition its docstring claims), 9 (TimeoutError branch mislabels an inner timeout).

Issue #415, opened 2026-07-29
SECURITY: all four user.py resolvers are unauthenticated - updateUserRoles allows self-granting Admin/Manager


Xilous:
every resolver in `backend/app/schemas/user.py` is unauthenticated. no `require_admin`, no `require_user`, no global auth middleware to compensate. only app middleware is CORS with `allow_origins=["*"]`.

four resolvers are open:
`Query.users` -> full Clerk roster: id, email, roles, gpBuyerId
`Mutation.updateUserRoles` -> set any user's roles, incl granting `Admin/Manager`
`Mutation.updateUserName`
`Mutation.updateUserGpBuyerId`

`updateUserRoles` is the serious one. an unauthenticated caller can self-grant `Admin/Manager`, the gate on 18 other resolvers:
relay install provisioning
relay install deletion
adopt windows
buyer assignments
`createGpJob`
`createGpBuyer`
`gpEmployees`

unauthenticated POST to a PR environment `/graphql` with no Authorization header returned all 10 users with emails, roles, GP buyer ids. escalation mutation deliberately not exercised - Clerk keys are inherited from production across environments, so calling it would modify real user metadata.

pre-existing, not from #412. `origin/master` copy of `user.py` has the same zero auth calls. dates to #216 / #240.

add `require_admin(info)` to all four, matching the 18 resolvers elsewhere that already do it.
`UserQueries.users` needs the `info: strawberry.Info` param added; three mutations take it too.
check frontend still passes its Clerk token on these calls once enforcing.
sweep for any other resolver with no gate - this was found by accident while correcting an unrelated comment, which is not a reliable way to find the rest.

PR #417, opened 2026-07-29
merge the warehouse deliveries view into the receiving page (#416)

### LINKED ISSUE #416: merge the warehouse deliveries view into the receiving page
deliveries page (`/app/warehouse/deliveries`, `DeliveriesView.tsx`) stops being a standalone page, becomes a section of the receiving page. it holds an "Upcoming Deliveries" accordion and an "Outstanding Items" back-order grid.

`expectedDeliveries` (deliveries "Upcoming" view) and `openPOs` (receiving "POs Awaiting Receipt" table) select the identical PO population - not soft-deleted, same three statuses:

GP_REGISTERED
VENDOR_CONFIRMED
PARTIALLY_RECEIVED

rendering both on one page would show the same PO list twice. the accordion's per-line ordered/received detail is subsumed by the back-ordered grid. the merge keeps only the back-ordered grid and carries the urgency signal onto the awaiting-receipt table.

frontend

- `ReceivingPage.tsx` gains a "Back-Ordered Items" section between "POs Awaiting Receipt" and "Recent Activity", using the `OutstandingView` grid from `DeliveriesView` (`backOrderedItems` query with urgency chips). no project landing step - the section queries all projects (`projectId: null`).
- add a Project column to the back-ordered grid. null project renders "Stock PO".
- POs Awaiting Receipt gains the urgency chip on its Expected Delivery column, replacing the plain "(overdue)" text. chip shows one of:

  "Nd overdue"
  "Today"
  "Tomorrow"
  "In Nd"

- delete `DeliveriesView.tsx`. in `modules/warehouse/index.tsx` replace the `deliveries` route with a redirect to `receiving`.
- in `Sidebar.tsx` remove the Deliveries entry from the Warehouse group.
- in `WarehouseLanding.tsx` remove the Deliveries card, move its `backOrderedCount` metric onto the Receiving card, replacing "received (7d)".
- in `graphql/warehouse.ts` delete `GET_EXPECTED_DELIVERIES`.

backend

- remove the `expectedDeliveries` resolver (`app/schemas/warehouse.py`), `get_expected_deliveries` (`app/repositories/warehouse/receiving.py`), and their re-exports in `app/repositories/warehouse/__init__.py`. nothing else consumes them and no tests reference them.
- keep `backOrderedItems`. add nullable `projectName` to `BackOrderedItem` (`app/schemas/types.py`), outer-join Project in `get_back_ordered_items`.
- `warehouseDashboard.backOrderedCount` stays.

simulated user testing

- https://frontend-production-34fc.up.railway.app/app/warehouse -> no Deliveries card, Receiving card shows the back-ordered metric
- sidebar Warehouse group -> no Deliveries entry
- /app/warehouse/receiving -> three sections appear in order: POs Awaiting Receipt, Back-Ordered Items, Recent Activity. back-ordered grid shows line-level rows with a Project column, outstanding quantities, urgency chips. awaiting-receipt table shows urgency chips in its Expected Delivery column.
- /app/warehouse/deliveries -> redirected to /app/warehouse/receiving
- no console errors, no expectedDeliveries network request anywhere in the warehouse module

Xilous:
closes #416.

Deliveries page folded into Receiving page.

expectedDeliveries and openPOs select the identical PO population - GP_REGISTERED, VENDOR_CONFIRMED, PARTIALLY_RECEIVED, not soft-deleted. rendering both on one page would show the same PO list twice, and the accordion's per-line ordered/received detail is the back-order grid's rows grouped by PO. so the Upcoming Deliveries accordion is dropped, not moved, and the one signal it carried that the table lacked - the urgency chip - moves onto POs Awaiting Receipt.

frontend
- ReceivingPage gains a Back-Ordered Items section between POs Awaiting Receipt and Recent Activity. no project landing step, queries all projects.
- grid gains a Project column. PO with no project renders "Stock PO".
- grid gains Ordered and Received beside Outstanding - the accordion was the only place that breakdown showed, and "2 of 10" and "2 of 3" are different problems. both were already being fetched and discarded.
- unitCost dropped from the query, not displayed. line economics belong to the PO module.
- POs Awaiting Receipt shows the urgency chip ("3d overdue" / "Today" / "Tomorrow" / "In 5d") in Expected Delivery, replacing the plain "(overdue)" text.
- DeliveriesView.tsx deleted. `deliveries` route -> redirect to `receiving`.
- sidebar Deliveries entry removed.
- landing Deliveries card removed, its backOrderedCount metric moves onto the Receiving card.
- GET_EXPECTED_DELIVERIES deleted.

backend
- `expectedDeliveries` resolver, `get_expected_deliveries`, and their re-exports removed. nothing else consumed them, no tests referenced them.
- `BackOrderedItem` gains a nullable `projectName`. `get_back_ordered_items` outer-joins Project, falls back description -> project_id.

refetch, and this part is beyond the issue text: a receive only ever refetched the warehouse inventory summaries, leaving the Receiving page's own reads stale. survivable while the back-order grid lived on another route, not survivable with the grid sitting directly under the button that invalidates it.
- added RECEIVE_REFETCH_QUERIES
  - inventory summaries
  - GetOpenPOs
  - GetBackOrderedItems
  - GetRecentReceiveRecords
- added backOrderedItems to the GP outbox drain eviction list, so a queued receipt that posts later clears the lines it filled.
- refetching by name only touches mounted queries, so this costs nothing elsewhere.

date parsing, found while testing the chip on the deployed PR environment: receiving page ran expectedDeliveryDate through `parseServerDate` -> date-only string left to `new Date` -> `new Date("2026-07-29")` is UTC midnight -> renders and compares as the previous day for any viewer behind UTC. every expected delivery printed a day early, and the urgency chip labelled a PO due today "1d overdue".

`parseServerDate` unchanged - its contract is deliberate and pinned by its own test, callers parse calendar dates into local components themselves - so this is fixed at the call site. `parseServerDay` added to `utils/serverDate.ts`, receiving page uses it for both the printed date and the chip. two PO screens keep their inline copies.

`testing/CLAUDE.md` updated:
- Deliveries page is gone
- relay-gated PR database never populates either grid on the receiving page
- fetch-stub technique that drives the real components instead

two real defects in this PR's own new code.

urgency chip used Math.ceil over the span between two local midnights. a DST boundary makes that span 23 or 25 hours, ceil rounds the error up into a whole day. the code comment claimed ceil was chosen for exactly this and had it backwards.
- tomorrow across a fall-back read "In 2d"
- a delivery exactly seven days out lost its chip
- yesterday across spring-forward evaluated to -0, not less than zero, so an already-late PO chipped "Today"

Math.round is correct for all four DST spans, verified under node with TZ=America/Toronto.

parseServerDay tests could not fail. CI runners are UTC, and on UTC parseServerDay and a bare new Date are the same function, so all three passed against the #238 bug as happily as against the fix. they now force America/Toronto, assert the platform parse names Jul 31 where ours names Aug 1, and carry a guard that fails if the forced zone stops taking effect. reverting the fix fails 2 of 7 under TZ=UTC, previously 0.

also fixed
- parseServerDate docstring claimed date-only strings keep local parsing; they do not, which is why parseServerDay had to exist
- back-order rows keyed on array position, now key on poLineItemId, newly exposed on BackOrderedItem. the refetches this PR adds re-run the query's ORDER BY, so position keys made the grid rebuild every row instead of diffing
- openPOs joins backOrderedItems in the outbox-drain eviction list. evicting one without the other left the awaiting-receipt table showing pre-receipt pending counts directly above a grid that had updated
- section header counted lines while the landing card counts units; header now reads "5 lines, 30 units"
- unit_cost removed from BackOrderedItem, receivedLast7Days removed from the dashboard query. this PR removed the last reader of each

not taken - the reworded Receiving caption flagged as dead code. DestinationCard renders metric or caption, and the metric is only supplied when the dashboard rollup resolved, so the caption is what shows when it failed - the path the file already comments on.

deferred - backOrderedItems duplicates the population openPOs already returns to the same page, and GET_OPEN_POS carries every column the grid needs. deriving the grid from that cache instead of a second unbounded query is a design change beyond this merge, and the resolver predates it.

merge conflict with master resolved in app/schemas/warehouse.py. #415 added require_user gates to the two resolvers this branch was changing. expected_deliveries stays deleted, back_ordered_items takes master's gated signature on top of this branch's changes. unauthenticated backOrderedItems probe against the PR backend returns "Authentication required".


frontend/src/modules/warehouse/ReceivingPage.tsx
```diff
@@ -63,22 +67,110 @@ interface RecentReceiveRecord {
...
+  dateStr: string | null,
+): { label: string; color: 'error' | 'warning' | 'info' } | null {
+  if (!dateStr) return null;
+  const today = new Date();
+  today.setHours(0, 0, 0, 0);
+  const date = parseServerDay(dateStr);
+  date.setHours(0, 0, 0, 0);
+  // Ceil, not round: a DST boundary makes the span 23 or 25 hours, and only ceil still calls that
+  // one day.
+  const days = Math.ceil((date.getTime() - today.getTime()) / (1000 * 60 * 60 * 24));
```
Xilous: `Math.ceil` is wrong across DST in both directions, and the comment above it asserts the opposite. verified in America/Toronto with node.

- today 2026-11-01, expected 2026-11-02 -> span 25h -> ceil 2 -> chip reads "In 2d" for tomorrow
- today 2026-03-09, expected 2026-03-08 -> span -23h -> ceil -0 -> `days < 0` false, `days === 0` true -> PO due yesterday chips "Today"
- today 2026-10-28, expected 2026-11-04 -> span 169h -> ceil 8 -> `days <= 7` fails, chip disappears entirely

`Math.round` is correct for all four DST spans (23h, 25h, -23h, -25h all land on +-1 day). the deleted DeliveriesView had the same ceil, so this carried the defect over rather than fixing it.

```suggestion
  // Round, not ceil: a DST boundary makes the span 23 or 25 hours, and only round calls both of
  // those one day.
  const days = Math.round((date.getTime() - today.getTime()) / (1000 * 60 * 60 * 24));
```

frontend/src/utils/__tests__/serverDate.test.ts
```diff
@@ -21,3 +21,25 @@ describe('parseServerDate', () => {
...
+    expect([d.getHours(), d.getMinutes()]).toEqual([0, 0]);
+  });
+
+  it('lands on local midnight, which is what makes a day-difference come out whole', () => {
+    // The whole point of #238: the UTC parse of this string is Jul 31 20:00 in Toronto, so a raw
+    // `new Date` renders and compares as the wrong day for half the world. Asserted against a
+    // locally-constructed date rather than the UTC one so the expectation holds in a UTC CI runner
+    // too, where the two happen to coincide.
+    expect(parseServerDay('2026-08-01').getTime()).toBe(new Date(2026, 7, 1).getTime());
+  });
```
Xilous, line 57: all three parseServerDay tests pass against the exact bug they exist to pin, when TZ is UTC. no TZ is set anywhere:
vitest.config.ts
src/test/setup.ts
package.json test scripts
.github/workflows/ci.yml

GitHub runners are UTC, and CI is the only place the suite runs.

substitute `new Date(value)` back into parseServerDay, under TZ=UTC:
- test 1 - getFullYear/getMonth/getDate/getHours read 2026/7/15/0 -> passes
- test 2 - `new Date('2026-08-01')` and `new Date(2026, 7, 1)` are the same instant -> passes
- test 3 - falls through either way -> passes

the comment says the assertion "holds in a UTC CI runner too, where the two happen to coincide". coinciding is what makes it vacuous.

pin the zone (set process.env.TZ before the module loads, or run the describe twice under two zones), or assert against a fixed UTC instant that differs from local midnight.

frontend/src/modules/warehouse/WarehouseLanding.tsx
```diff
@@ -45,19 +44,15 @@ function buildDestinations(d: WarehouseDashboard | undefined): Destination[] {
...
       label: 'Receiving',
       path: '/app/warehouse/receiving',
       icon: <Download size={18} strokeWidth={1.75} />,
-      caption: 'Book POs into inventory',
-      metric: d ? { value: d.receivedLast7Days, noun: 'received (7d)', attention: false } : undefined,
+      caption: 'Book POs into inventory, and what is still owed',
+      // The back-order count rather than receipts-in-7-days: a card carries one number, and this is
+      // the one that says whether anything needs chasing. What was received lately is history, and
+      // the page itself still lists it.
+      metric: d ? { value: d.backOrderedCount, noun: 'back-ordered', attention: false } : undefined,
```
Xilous, line 56: receivedLast7Days is now rendered nowhere. DashboardCards draws only totalItemCount and totalValue, and this card was its only other consumer.

still selected by GET_WAREHOUSE_DASHBOARD (graphql/warehouse.ts:364)
still computed as a SUM over ReceiveLineItem on every dashboard load (repositories/warehouse/progress.py:79-86)

the comment says "the page itself still lists it". the Receiving page lists the last 10 receive records, not a 7-day unit total, so the number is gone rather than moved.

either surface it somewhere, or drop it from the query, the type and the rollup.

frontend/src/graphql/refetch.ts
```diff
@@ -226,6 +239,9 @@ export const GP_OUTBOX_DRAINED_STALE_ROOT_FIELDS = [
   'purchaseOrder',
   'poReceivingDetails',
   'recentReceiveRecords',
+  // A receipt that finally posts is a receipt that closes back-ordered lines, and the section
+  // reading them is on the page the user is most likely sitting on when the drain lands (#416).
+  'backOrderedItems',
```
Xilous, line 246: openPOs belongs in this list too. a drained receipt moves received_quantity and can move status to CLOSED, which drops the PO out of openPOs, and openPOs feeds POs Awaiting Receipt - the section immediately above the one this line was added for, on the same page.

as written, a drain empties the back-order rows the receipt closed while the PO sits in the table above with its old pending qty and status.

RECEIVE_REFETCH_QUERIES already names GetOpenPOs. the drain is the same receive landing later.

backend/app/repositories/warehouse/receiving.py
```diff
@@ -538,28 +539,25 @@ def get_po_receiving_details(session: Session, po_id: uuid.UUID) -> tuple[POMode
...
+            POLineItemModel,
+            POModel.po_number,
+            VendorModel.name,
+            POModel.expected_delivery_date,
+            ProjectModel.description,
+            ProjectModel.project_id,
+        )
         .join(POModel, POLineItemModel.po_id == POModel.id)
         .outerjoin(VendorModel, POModel.vendor_id == VendorModel.id)
+        .outerjoin(ProjectModel, POModel.project_id == ProjectModel.id)
```
Xilous, line 560: this selects the population get_open_pos already returns. both filter deleted_at IS NULL plus (po_repository.py:492-501)
GP_REGISTERED
VENDOR_CONFIRMED
PARTIALLY_RECEIVED

GET_OPEN_POS already ships lineItems{orderedQuantity receivedQuantity}, vendor, projectId and expectedDeliveryDate - every column of the new grid. the project name this outer join was added for is what projectMap already resolves for the PO table on the same page.

net effect: the receiving page runs a second unbounded cross-project query, and the backend grew a join, for rows already in the Apollo cache.

derive backOrderRows from openPOsData, flatMap the lines where ordered > received.

frontend/src/modules/warehouse/ReceiveModal.tsx
```diff
@@ -494,7 +494,7 @@ export default function ReceiveModal({ open, onClose, poIds }: ReceiveModalProps
         return next;
       });
       try {
-        await client.refetchQueries({ include: WAREHOUSE_REFETCH_QUERIES });
+        await client.refetchQueries({ include: RECEIVE_REFETCH_QUERIES });
```
Xilous, line 497: this await sits ahead of setSucceeded(true) and the toast, so the receive confirmation waits on 11 queries instead of 8. the three added are the receiving page's own reads, GetBackOrderedItems being the unbounded cross-project one. on Railway's network hop that lands between the user pressing confirm and seeing the receipt.

nothing in the success path reads the refetched data. drop the await, or move it below setSucceeded and showToast.

frontend/src/utils/serverDate.ts
```diff
@@ -12,3 +12,17 @@ export function parseServerDate(value: string): Date {
...
+ * `new Date("2026-08-01")` is UTC midnight per spec, which prints and compares as July 31 anywhere
+ * behind UTC, so a calendar date has to be rebuilt from its parts to mean the day it names. Use this
+ * for any date-ONLY field - an expected delivery, a required-by - and `parseServerDate` for instants.
+ *
+ * Anything that is not a bare YYYY-MM-DD falls through, so a caller holding a mix of the two can
+ * route everything here.
+ */
+export function parseServerDay(value: string): Date {
+  const m = /^(\d{4})-(\d{2})-(\d{2})$/.exec(value);
+  return m ? new Date(Number(m[1]), Number(m[2]) - 1, Number(m[3])) : parseServerDate(value);
```
Xilous, line 28: PODetailModal.tsx:63 and POGenerateDialog.tsx:63 are byte-identical to this line, regex included. both should call the helper - three copies of one date parse is how one of them drifts.

PODetailModal's wrapper also carries an isNaN guard that the new ReceivingPage formatDate does not. pick the intended behaviour and keep it in one place.

frontend/src/utils/serverDate.ts
```diff
@@ -12,3 +12,17 @@ export function parseServerDate(value: string): Date {
   const zoneless = /T\d{2}:\d{2}/.test(value) && !/(Z|[+-]\d{2}:?\d{2})$/.test(value);
```
Xilous, line 13: the docstring above this line still says date-only strings "must keep local parsing (#238)". `new Date("2026-07-29")` is UTC midnight per spec, the opposite of that, and is the entire reason parseServerDay had to be added below it. that sentence is what sends the next caller here with a calendar date and reproduces #238.

serverDate.test.ts:17 pins `parseServerDate('2026-08-15')` against `new Date('2026-08-15')`, true by construction, and asserts nothing about the contract it claims to pin.

frontend/src/modules/warehouse/ReceivingPage.tsx
```diff
@@ -286,6 +401,41 @@ export default function ReceivingPage() {
...
+        component="div"
+        sx={{
+          ...microLabelSx,
+          pb: 0.75,
+          mb: 1.5,
+          borderBottom: '2px solid',
+          borderColor: 'text.primary',
+        }}
+      >
+        Back-Ordered Items{backOrderRows.length > 0 ? ` (${backOrderRows.length})` : ''}
```
Xilous: this header counts lines; the landing card the metric just moved onto counts units. back_ordered_count is SUM(ordered - received) over the same rows (progress.py:89-100), so the Receiving card reads "412 back-ordered" while this header reads "(37)" for the same data. the Deliveries card had the same mismatch.

this PR drops testing/CLAUDE.md's "The two numbers matching their destination pages is the thing to assert" rather than the discrepancy.

either sum outstandingQuantity for the header, or count lines in the rollup.

frontend/src/modules/warehouse/WarehouseLanding.tsx
```diff
@@ -45,19 +44,15 @@ function buildDestinations(d: WarehouseDashboard | undefined): Destination[] {
...
-      caption: 'Upcoming POs and outstanding lines',
-      metric: d ? { value: d.backOrderedCount, noun: 'back-ordered', attention: false } : undefined,
-    },
     {
       label: 'Receiving',
       path: '/app/warehouse/receiving',
       icon: <Download size={18} strokeWidth={1.75} />,
-      caption: 'Book POs into inventory',
-      metric: d ? { value: d.receivedLast7Days, noun: 'received (7d)', attention: false } : undefined,
+      caption: 'Book POs into inventory, and what is still owed',
```
Xilous, line 52: this caption is unreachable. DestinationCard renders `dest.metric ? <metric> : <caption>` and this card always has a metric, so the reworded text never reaches the screen. the same was true of the old wording, so the edit has no visible effect either way.

frontend/src/modules/warehouse/ReceivingPage.tsx
```diff
@@ -203,6 +304,19 @@ export default function ReceivingPage() {
     [openPOsData, projectMap],
   );
 
+  const backOrderRows = useMemo(
+    () =>
+      (backOrderData?.backOrderedItems ?? []).map((item, i) => ({
+        ...item,
+        // No field on the row is unique on its own - the same product code is routinely back-ordered
+        // on more than one PO - so the index is what keeps the grid keys apart.
+        id: `${item.poNumber}-${item.productCode}-${i}`,
```
Xilous: the line does have a unique key - POLineItem.id, already in hand at receiving.py:573 as row[0].id, just not on the BackOrderedItem type.

keying on the array index makes the ordering (expected_delivery_date, hardware_category) decide row identity, so every refetch this PR just wired up hands DataGrid a fresh id set for unchanged rows and it rebuilds instead of diffing.

expose `id` on BackOrderedItem, and drop the comment above.

backend/app/repositories/warehouse/receiving.py
```diff
@@ -576,6 +574,9 @@ def get_back_ordered_items(session: Session, project_id: uuid.UUID | None = None
             "po_number": row[1],
             "vendor_name": row[2],
             "expected_delivery_date": row[3],
+            # Same fallback the rest of the app shows a project by: the description if it has one,
+            # otherwise the TITAN project number.
+            "project_name": (row[4] or row[5]) or None,
```
Xilous: the outer `or None` is a no-op. `(a or b) or None` and `a or b or None` are the same expression, both falling to None when description and project_id are both empty.

```suggestion
            "project_name": row[4] or row[5] or None,
```

backend/app/schemas/warehouse.py
```diff
@@ -430,6 +422,7 @@ def back_ordered_items(self, project_id: strawberry.ID | None = None) -> list[Ba
                     po_number=item["po_number"],
                     vendor_name=item["vendor_name"],
                     expected_delivery_date=item["expected_delivery_date"],
+                    project_name=item["project_name"],
```
Xilous, line 453: unit_cost has no consumer now that GET_BACK_ORDERED_ITEMS stopped selecting it, the only query reading the field. the float() conversion at line 421 still runs per row and the field stays on the type.

if line economics belong to the PO module, drop it from BackOrderedItem (types.py:1350) and from this constructor.

Issue #240, opened 2026-07-14
Admin -> User Management: allow name change control


Xilous:
change user's display name

Issue #216, opened 2026-07-09
Greg Sutton - PO Creation feedback


Xilous:
manual direct po creation
- select buyer first, buyers are assigned to specific projects, which they can create POs for.

cost codes are also designated to specific buyers

PO Request creation - user needs to be able to write down a "Preferred Delivery Date" on the PO Request

PO request getting processed by PO User to go into GP Registered,
	- PO user does not set the site/shop classification, remove it from there
	- site/shop should by PM, therefore, it should be during PO request creation within hardware schedule import module (Start a Task)

vendor dictates the currency of the PO being created, the tax schedule is set by GP (i think) right after creation?
- if the PO's currency is USD, then the GP "tax schedule ID" needs to be null
- so if it is right after the PO is created, GP sets the tax schedule id to USD.
- the requirement is that GP auto assigns "BC HST 5%" company tax schedule on the PO, not sure if this happens during PO creation or right after, but it needs to not be ensured to be not the case. all POs that with currencies set to USD needs to have no tax schedule on the PO.

and for the actual "tax detail", these are the ones that can be possibly used, but dictated by the user 

"BC HST P"
"ON HST - P"
"PST 7%"

these following values also needs to be set on a PO for it to be "GP Registered".
Trade Discount
Freight
Miscellaneous

PR #420, opened 2026-07-29
fix(security): gate every unauthenticated GraphQL resolver (#415)

### LINKED ISSUE #415: SECURITY: all four user.py resolvers are unauthenticated - updateUserRoles allows self-granting Admin/Manager

Xilous:
Closes #415.

every resolver in backend/app/schemas/user.py reachable with no session. get_context in app/auth.py only stashes the request; the only middleware is CORS with allow_origins=["*"]. verification is opt-in per resolver.

updateUserRoles sets a user's roles outright including Admin/Manager, the gate on
relay install provisioning
relay install deletion
adopt windows
buyer assignments
createGpJob
createGpBuyer
gpEmployees

an unauthenticated caller could grant themselves that role. users returned the full Clerk roster - id, email, roles, gpBuyerId.

pre-existing, dating to #216 / #240.

sweep found 87 more ungated resolvers across 11 further schema files. the ones with real teeth:
deleteWarehouse, createWarehouse, updateWarehouse - unauthenticated warehouse CRUD
overrideInventoryQuantity - sets an inventory row's quantity outright
mergeLocations - rewrites the location of every inventory row, opening item and stock item at a location
createVendor, updateVendor, deleteVendor
poDocumentDownloadUrl - mints a presigned S3 URL for a supplier PO document
every PO, project, shipping, stock and warehouse list query, plus dashboards, notifications and the whole import path

gates assigned from the frontend consumers, grepped. require_admin where every frontend caller lives in the admin module, require_user otherwise.

require_admin on 13 resolvers -
the four user.py ones
openingHardwareStatus
adminStats
the three vendor writes
locationDuplicates
overrideInventoryQuantity
mergeLocations
the three warehouse writes

require_user on the other 78, the cheaper gate: JWT verify only, no Clerk Backend API round-trip. vendors, warehouses and buyerAssignments are read by non-admin screens (VendorSelect on the PO screens, the warehouse picker in the receive/transfer/return dialogs, the PO dialog), so they take require_user even though their writes are admin.

161 resolvers now call a gate as the first statement of their body. enrollRelayInstall is the only exception, called by the relay during one-time setup before it holds any credential and authenticated by the enrollment token - relay_repository.enroll_install matches by hash, rejects when expired, rejects on reuse via the enrolled_at guard.

two test layers.
- test_resolver_gate_completeness.py is new. AST-parses every @strawberry.field / @strawberry.mutation under app/schemas/ and fails on any that calls none of the gates. an ungated resolver is an explicit entry on _EXEMPT with a reason. no database, runs locally and in CI. carries a guard-the-guard assertion so an AST change that matches nothing cannot pass vacuously.
- test_resolver_auth_gates.py gains rows for all 13 admin-gated resolvers, existing monkeypatched-sentinel pin pattern. these prove the gate runs, which the source scan cannot.

no frontend changes. frontend/src/apollo.ts already attaches the Clerk session token to every GraphQL request via SetContextLink, and the app sits behind `<SignedIn>` / `<RedirectToSignIn>`. public GraphQL SDL byte-identical to master, built on both branches and diffed; strawberry.Info is excluded from the schema by annotation, so no query document changes.

verified locally - pytest 347 passed / 493 skipped (skips are the DB-requiring tests, no local Postgres), ruff check and ruff format clean, vitest 358 passed across 36 files, SDL diff empty.

out of scope, worth separate issues: tightening CORS allow_origins=["*"], and caching the Clerk role lookup per-request for pages that hit several require_admin resolvers at once. /admin/reset-data was already double-gated by TESTING_ENABLED plus require_admin_request.

PR environments inherit production Clerk keys, so any tooling that POSTs to /graphql without a token starts getting auth errors.

### DIFF
--- a/backend/app/auth.py
+++ b/backend/app/auth.py
@@ -5,6 +5,13 @@ The backend has no global auth middleware; verification is opt-in per resolver v
 so ungated queries pay no JWT/JWKS/Clerk-API cost. Roles live in Clerk publicMetadata
 (not in the default session token), so we verify the token for identity (``sub``) and
 then look up roles through the Clerk Backend API.
+
+Opt-in means a resolver that forgets to ask is silently public, which is how every resolver in
+``app/schemas/user.py`` - ``updateUserRoles`` among them - shipped reachable with no session at all
+(#415). Since that sweep the opt-in is enforced from outside: ``tests/test_resolver_gate_completeness``
+walks every ``@strawberry.field`` / ``@strawberry.mutation`` under ``app/schemas/`` and fails on any
+that calls none of the gates below, so leaving one open is now an explicit entry on that test's
+exemption list rather than an omission nobody notices.
 """
 
 import time
--- a/backend/app/schemas/buyer.py
+++ b/backend/app/schemas/buyer.py
@@ -4,7 +4,7 @@ import uuid
 
 import strawberry
 
-from app.auth import require_admin
+from app.auth import require_admin, require_user
 from app.database import SessionLocal
 from app.repositories import buyer_repository
 
@@ -15,9 +15,11 @@ from .types import BuyerAssignment
 @strawberry.type
 class BuyerQueries:
     @strawberry.field
-    def buyer_assignments(self) -> list[BuyerAssignment]:
+    def buyer_assignments(self, info: strawberry.Info) -> list[BuyerAssignment]:
         """Issue #216: per-buyer project + cost-code authorization. The PO dialog reads this to
-        filter its options; the create/register mutations re-enforce it server-side."""
+        filter its options; the create/register mutations re-enforce it server-side. require_user,
+        not admin (#415), because that dialog is a PO-user screen."""
+        require_user(info)
         with SessionLocal() as session:
             return [buyer_assignment_to_type(a) for a in buyer_repository.list_assignments(session)]
 
--- a/backend/app/schemas/dashboard.py
+++ b/backend/app/schemas/dashboard.py
@@ -2,6 +2,7 @@
 
 import strawberry
 
+from app.auth import require_admin, require_user
 from app.database import SessionLocal
 from app.repositories import dashboard_repository, user_repository
 
@@ -11,7 +12,8 @@ from .types import AdminStats, HomeDashboardStats, ShopAssemblyStats
 @strawberry.type
 class DashboardQueries:
     @strawberry.field
-    def home_dashboard_stats(self) -> HomeDashboardStats:
+    def home_dashboard_stats(self, info: strawberry.Info) -> HomeDashboardStats:
+        require_user(info)
         with SessionLocal() as session:
             d = dashboard_repository.get_home_dashboard_stats(session)
             return HomeDashboardStats(
@@ -22,7 +24,8 @@ class DashboardQueries:
             )
 
     @strawberry.field
-    def shop_assembly_stats(self) -> ShopAssemblyStats:
+    def shop_assembly_stats(self, info: strawberry.Info) -> ShopAssemblyStats:
+        require_user(info)
         with SessionLocal() as session:
             d = dashboard_repository.get_shop_assembly_stats(session)
             return ShopAssemblyStats(
@@ -30,7 +33,10 @@ class DashboardQueries:
             )
 
     @strawberry.field
-    def admin_stats(self) -> AdminStats:
+    def admin_stats(self, info: strawberry.Info) -> AdminStats:
+        """Admin-gated (#415): only the admin landing page reads it, and it enumerates Clerk users to
+        count them."""
+        require_admin(info)
         users = user_repository.list_users()
         with SessionLocal() as session:
             d = dashboard_repository.get_admin_stats(session, user_count=len(users))
--- a/backend/app/schemas/vendor.py
+++ b/backend/app/schemas/vendor.py
@@ -4,6 +4,7 @@ import uuid
 
 import strawberry
 
+from app.auth import require_admin, require_user
 from app.database import SessionLocal
 from app.repositories import vendor_repository
 
@@ -15,12 +16,16 @@ from .types import Vendor
 @strawberry.type
 class VendorQueries:
     @strawberry.field
-    def vendors(self) -> list[Vendor]:
+    def vendors(self, info: strawberry.Info) -> list[Vendor]:
+        """require_user, not admin (#415): VendorSelect renders this list on the PO screens, so it is
+        not an admin-only read even though only an admin may write it."""
+        require_user(info)
         with SessionLocal() as session:
             return [vendor_to_type(v) for v in vendor_repository.list_vendors(session)]
 
     @strawberry.field
-    def vendor(self, id: strawberry.ID) -> Vendor | None:
+    def vendor(self, info: strawberry.Info, id: strawberry.ID) -> Vendor | None:
+        require_user(info)
         with SessionLocal() as session:
             v = vendor_repository.find_vendor(session, uuid.UUID(str(id)))
             return vendor_to_type(v) if v is not None else None
@@ -29,7 +34,8 @@ class VendorQueries:
 @strawberry.type
 class VendorMutations:
     @strawberry.mutation
-    def create_vendor(self, input: CreateVendorInput) -> Vendor:
+    def create_vendor(self, info: strawberry.Info, input: CreateVendorInput) -> Vendor:
+        require_admin(info)
         with SessionLocal() as session:
             vendor = vendor_repository.create_vendor(
                 session,
@@ -44,7 +50,8 @@ class VendorMutations:
             return vendor_to_type(vendor)
 
     @strawberry.mutation
-    def update_vendor(self, id: strawberry.ID, input: UpdateVendorInput) -> Vendor:
+    def update_vendor(self, info: strawberry.Info, id: strawberry.ID, input: UpdateVendorInput) -> Vendor:
+        require_admin(info)
         with SessionLocal() as session:
             vendor = vendor_repository.update_vendor(
                 session,
@@ -60,7 +67,8 @@ class VendorMutations:
             return vendor_to_type(vendor)
 
     @strawberry.mutation
-    def delete_vendor(self, id: strawberry.ID) -> bool:
+    def delete_vendor(self, info: strawberry.Info, id: strawberry.ID) -> bool:
+        require_admin(info)
         with SessionLocal() as session:
             vendor_repository.delete_vendor(session, uuid.UUID(str(id)))
             session.commit()
--- a/backend/app/schemas/vendor.py
+++ b/backend/app/schemas/vendor.py
@@ -35,7 +35,12 @@ class VendorQueries:
 class VendorMutations:
     @strawberry.mutation
     def create_vendor(self, info: strawberry.Info, input: CreateVendorInput) -> Vendor:
-        require_admin(info)
+        """require_user, not admin. `VendorSelect` offers "+ Create new vendor" to every caller and
+        renders `VendorEditDialog` with `vendor={null}`, so a PO user adding a supplier mid-order goes
+        through here. Editing or deleting an existing vendor is still admin - only the admin Vendors
+        page reaches those. The dialog living under `modules/admin/` says nothing about who can open
+        it, which is exactly the trap this comment exists to stop."""
+        require_user(info)
         with SessionLocal() as session:
             vendor = vendor_repository.create_vendor(
                 session,
--- a/backend/app/schemas/warehouse.py
+++ b/backend/app/schemas/warehouse.py
@@ -1027,9 +1027,14 @@ class WarehouseMutations:
     def override_inventory_quantity(
         self, info: strawberry.Info, input: OverrideInventoryQuantityInput
     ) -> InventoryLocation:
-        """Admin-gated (#415): sets an inventory row's quantity outright rather than by a delta, from
-        the admin Inventory Correction modal. Nothing downstream re-checks the caller."""
-        require_admin(info)
+        """require_user, not admin. The Correction button on the warehouse Hardware Items tab is
+        rendered for everyone ("available to all users, not just admins", HardwareItemsTab.tsx), and it
+        opens `InventoryCorrectionModal`, which is what calls this. Admin-gating it would break the
+        warehouse's own count-correction workflow.
+
+        Gating it would also buy nothing while `adjustInventoryQuantity` next door stays open to any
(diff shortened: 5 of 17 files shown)


backend/app/schemas/vendor.py
```diff
@@ -29,7 +34,8 @@ def vendor(self, id: strawberry.ID) -> Vendor | None:
 @strawberry.type
 class VendorMutations:
     @strawberry.mutation
-    def create_vendor(self, input: CreateVendorInput) -> Vendor:
+    def create_vendor(self, info: strawberry.Info, input: CreateVendorInput) -> Vendor:
+        require_admin(info)
```
Xilous: createVendor is admin-gated, the create path is not admin-only in the UI.

components/VendorSelect.tsx renders VendorEditDialog behind a "+ Create new vendor..." option. VendorSelect is mounted on:
modules/po/GpPurchaseOrderDialog.tsx:837
modules/po/PODetailModal.tsx:610
modules/import/PurchaseOrdersStep.tsx:167

all PO User / Hardware Schedule Import surfaces, none with an isAdmin check. a PO user who picks "Create new vendor" and saves gets "Admin/Manager role required" where it used to work.

the docstring added on `vendors` two lines up covers the read half ("VendorSelect renders this list on the PO screens, so it is not an admin-only read even though only an admin may write it"). the same component does the write.

backend/app/schemas/warehouse.py
```diff
@@ -965,7 +1024,12 @@ def adjust_inventory_quantity(
             return inventory_location_to_type(result)
 
     @strawberry.mutation
-    def override_inventory_quantity(self, input: OverrideInventoryQuantityInput) -> InventoryLocation:
+    def override_inventory_quantity(
+        self, info: strawberry.Info, input: OverrideInventoryQuantityInput
+    ) -> InventoryLocation:
+        """Admin-gated (#415): sets an inventory row's quantity outright rather than by a delta, from
+        the admin Inventory Correction modal. Nothing downstream re-checks the caller."""
+        require_admin(info)
```
Xilous: overrideInventoryQuantity is admin-gated, modules/warehouse/HardwareItemsTab.tsx opens InventoryCorrectionModal from an ungated Actions column at line 432. line 440 states "Correction (quantity override + location fixes) is available to all users, not just admins."

modules/warehouse/OpeningItemsTab.tsx:270 wraps its Correction button in `isAdmin &&`, HardwareItemsTab does not. a Warehouse Staff user clicking Correction on the Hardware Items tab hits FORBIDDEN on confirm.

gate the button, or leave this at require_user.

backend/app/schemas/warehouse.py
```diff
@@ -954,8 +1012,9 @@ def cancel_pull_request(self, info: strawberry.Info, input: CancelPullRequestInp
     # Admin Corrections
     @strawberry.mutation
     def adjust_inventory_quantity(
-        self, inventory_location_id: strawberry.ID, adjustment: int, reason: str
+        self, info: strawberry.Info, inventory_location_id: strawberry.ID, adjustment: int, reason: str
     ) -> InventoryLocation:
+        require_user(info)
```
Xilous, line 1017: adjustInventoryQuantity stays require_user, override_inventory_quantity 15 lines below becomes require_admin, both write the same column.

app/repositories/warehouse/inventory.py:357 accepts any delta that lands on a non-negative quantity -> non-admin computes the delta -> any target quantity, the admin gate on the override bypassed through its sibling.

adjustInventoryQuantity is reachable from
modules/warehouse/SpotCheckModal.tsx
modules/warehouse/LocationActionDialog.tsx

app/repositories/warehouse/inventory.py:381 hardcodes performed_by="Admin/Manager" on the audit row, so the trail claims an admin did what a Warehouse Staff user did.

testing/CLAUDE.md
```diff
@@ -19,6 +19,30 @@ This is a tester's knowledge journal for UC Nexus. It documents how the app work
   - Tokens are one-time use; fetch a fresh one each session. Works on any runtime with the same Clerk dev instance.
 - **Test XML file**: `testing/fixtures/contracterp-74.xml` - TITAN hardware schedule export, use for Import wizard testing (upload via `upload_file`)
 
+### Every resolver needs a token now (#415)
+
+Until #415 most resolvers were reachable with no `Authorization` header at all, so an injected
+`fetch('/graphql', ...)` helper that forgot the token still returned data and nothing looked wrong.
+That is over: every resolver in `app/schemas/` calls `require_user` / `require_admin` / `require_role`
+as its first statement, enforced by `backend/tests/test_resolver_gate_completeness.py`. The only
+exception is `enrollRelayInstall`, which carries its own enrollment-token auth.
```
Xilous, line 28: "The only exception is enrollRelayInstall" holds only for app/schemas/.

backend/main.py:381 GET /testing/clerk-sign-in takes an arbitrary `email` query param, is gated on nothing but TESTING_ENABLED, and returns a Clerk sign-in token for that account. on any PR environment or test deployment, an unauthenticated path to a full admin session that nullifies all 92 gates this PR adds.

/admin/reset-data directly above it at main.py:292 got TESTING_ENABLED plus require_admin_request, this one got neither.

test_resolver_gate_completeness.py only walks app/schemas/, nothing catches it.

backend/tests/test_resolver_gate_completeness.py
```diff
@@ -0,0 +1,98 @@
...
+            f = node.func
+            name = f.id if isinstance(f, ast.Name) else (f.attr if isinstance(f, ast.Attribute) else None)
+            if name in _GATES:
+                return True
+    return False
+
+
+def _all_resolvers():
+    rows = []
+    for path in sorted(_SCHEMAS_DIR.glob("*.py")):
```
Xilous: _SCHEMAS_DIR.glob("*.py") is non-recursive, and the sweep only looks at app/schemas/.

the new auth.py docstring claims the test "walks every ``@strawberry.field`` / ``@strawberry.mutation`` under ``app/schemas/``", which reads as recursive.

a resolver added in a future app/schemas/<subpackage>/*.py, or any Strawberry type registered from outside app/schemas/, is silently unchecked.

use rglob("*.py").

backend/tests/test_resolver_gate_completeness.py
```diff
@@ -0,0 +1,98 @@
...
+    return rows
+
+
+_RESOLVERS = _all_resolvers()
+
+
+def test_the_sweep_found_resolvers_at_all():
+    """Guard the guard: an AST change that silently matches nothing would make this file pass by
+    finding no resolvers to check."""
+    assert len(_RESOLVERS) > 100
```
Xilous: 161 resolvers match today, `> 100` leaves room for a whole module to fall out of the AST match and still pass.
warehouse.py 46
po.py 19
stock.py 17

raise the floor, or assert that every schema module with resolvers contributes at least one row.

backend/tests/test_resolver_gate_completeness.py
```diff
@@ -0,0 +1,98 @@
...
+"""
+
+import ast
+from pathlib import Path
+
+import pytest
+
+_SCHEMAS_DIR = Path(__file__).resolve().parent.parent / "app" / "schemas"
+
+_GATES = {"require_user", "require_admin", "require_role", "require_admin_request"}
```
Xilous: require_admin_request does not belong in _GATES. it takes a FastAPI Request, not a strawberry Info (auth.py:142), so a resolver calling require_admin_request(info) satisfies this test and then AttributeErrors on info.headers at runtime.

no resolver callers, only main.py's plain HTTP routes, which this test never scans.

drop it, leaving the three gates a resolver can call.

backend/app/auth.py
```diff
@@ -5,6 +5,13 @@
 so ungated queries pay no JWT/JWKS/Clerk-API cost. Roles live in Clerk publicMetadata
 (not in the default session token), so we verify the token for identity (``sub``) and
 then look up roles through the Clerk Backend API.
+
+Opt-in means a resolver that forgets to ask is silently public, which is how every resolver in
+``app/schemas/user.py`` - ``updateUserRoles`` among them - shipped reachable with no session at all
+(#415). Since that sweep the opt-in is enforced from outside: ``tests/test_resolver_gate_completeness``
+walks every ``@strawberry.field`` / ``@strawberry.mutation`` under ``app/schemas/`` and fails on any
+that calls none of the gates below, so leaving one open is now an explicit entry on that test's
```
Xilous, line 13: enforcement here is 92 hand-copied require_user(info) lines plus an AST test policing them.

main.py:35 already has ErrorHandlerExtension, a SchemaExtension whose resolve() hook sees every field. a deny-by-default check there - root fields only via info.path.prev is None, with a named allowlist - makes "forgot the gate" unrepresentable rather than merely detectable, and collapses JWT verification from once per root field to once per request.

the per-resolver call then only where the gate is stronger than the default.

backend/app/schemas/dashboard.py
```diff
@@ -22,15 +24,19 @@ def home_dashboard_stats(self) -> HomeDashboardStats:
...
             return ShopAssemblyStats(
                 active_pull_request_count=d["active_pull_request_count"],
             )
 
     @strawberry.field
-    def admin_stats(self) -> AdminStats:
+    def admin_stats(self, info: strawberry.Info) -> AdminStats:
+        """Admin-gated (#415): only the admin landing page reads it, and it enumerates Clerk users to
+        count them."""
+        require_admin(info)
```
Xilous, line 39: admin_stats costs two Clerk Backend API round trips on every admin landing load - require_admin calls user_repository.get_user_roles(user_id), then list_users() paginates the same API 100 users at a time. user.users has the same double hit.

list_users() already returns roles per user via _user_summary, so the caller's roles can come out of that payload.

backend/tests/test_resolver_auth_gates.py
```diff
@@ -219,6 +237,79 @@ def _id() -> str:
...
+        "updateUserName",
+        user_module,
+        lambda: UserMutations().update_user_name(FakeInfo(), "user_1", "First", "Last"),
+    ),
+    (
+        "updateUserGpBuyerId",
+        user_module,
+        lambda: UserMutations().update_user_gp_buyer_id(FakeInfo(), "user_1", "donr"),
+    ),
+    # #415 sweep: the rest of the admin-only surface, all of it previously ungated. The warehouse and
```
Xilous, line 277: _USER_GATED was not extended, so none of the roughly 80 new require_user resolvers gets a pin test - only the AST scan, which by its own docstring cannot prove the gate runs first.

the ones with teeth are the writes:
confirmShipment
destockInventory
transferInventory
allocateStockToProject
reclassifyStockItem
completePullRequest

moving require_user below the SessionLocal block in any of those keeps both this file and the completeness test green.

backend/app/schemas/warehouse.py
```diff
@@ -868,11 +923,14 @@ def set_pull_item_fetched(
...
     @strawberry.mutation
-    def complete_pull_request(self, id: strawberry.ID, completed_by: str | None = None) -> PullRequest:
+    def complete_pull_request(
+        self, info: strawberry.Info, id: strawberry.ID, completed_by: str | None = None
+    ) -> PullRequest:
         """Close the whole pull in one go. Since #343 this reads as "stage everything still
         outstanding and finish": any opening not yet confirmed individually is flipped to PULLED and
         stamped here. `completedBy` is optional and additive - it only names the actor on those
         stamps, so existing callers are unaffected."""
+        require_user(info)
```
Xilous, line 933: require_user(info) returns {user_id} and all 92 call sites throw it away.

completed_by here, and performed_by / picked_by / received_by elsewhere, stay client-supplied strings. the resolver knows who the caller is and still stamps whoever the browser named.

resolve_display_name(user_id) exists for this, #199 used it for received_by.

testing/CLAUDE.md
```diff
@@ -19,6 +19,30 @@ This is a tester's knowledge journal for UC Nexus. It documents how the app work
...
+
+Consequences when driving the app by script:
+
+- Always mint a token first - `await window.Clerk.session.getToken({skipCache: true})` - and send it
+  as `Authorization: Bearer <token>`. A helper without one now gets
```
Xilous, line 33: frontend/src/apollo.ts:17 reads window.Clerk.session.getToken() and, when that is null or throws, sends no Authorization header at all. errorLink only console.errors the result.

before this PR a token gap degraded a couple of admin panels. it now blanks every page with "Authentication required".

useAuth().getToken from @clerk/clerk-react plus a retry on UNAUTHENTICATED.

backend/app/schemas/buyer.py
```diff
@@ -15,9 +15,11 @@
 @strawberry.type
 class BuyerQueries:
     @strawberry.field
-    def buyer_assignments(self) -> list[BuyerAssignment]:
+    def buyer_assignments(self, info: strawberry.Info) -> list[BuyerAssignment]:
         """Issue #216: per-buyer project + cost-code authorization. The PO dialog reads this to
-        filter its options; the create/register mutations re-enforce it server-side."""
+        filter its options; the create/register mutations re-enforce it server-side. require_user,
+        not admin (#415), because that dialog is a PO-user screen."""
+        require_user(info)
```
Xilous, line 22: buyerAssignments returns buyer_repository.list_assignments(session) - every buyer's authorized projects and cost codes, not the caller's. the PO dialog only needs the caller's row.

scope the query to the caller, admins see all.

PR #421, opened 2026-07-29
feat(relay): multi-backend channels + seeded PR-environment credential (#414)

### LINKED ISSUE #414: PR environments can't test relay-dependent paths: relay dials one backend, PR env has no credential
relay-dependent paths can't be exercised in Railway PR environments. GP-touching changes verifiable only after merge, which defeats the point of having PR environments.

blocker 1 - relay dials exactly one backend
`ChannelCfg.backend_url` (`relay/src/ucnexus_relay/config.py`) is a single string, default `wss://backend-production-7866.up.railway.app/relay-link`. `channel.run_forever` opens one connection to it. no list, no second channel.

blocker 2 - PR environment has no credential
`relay_installs` is a Postgres table. `backend/entrypoint.sh` -> `alembic upgrade head` -> uvicorn, no seed step. empty table -> `authenticate_secret` matches nothing -> `/relay-link` closes 4403. `provision_install` mints enrollment tokens per backend, so every PR environment needs its own provision + enroll.

gap 3 - relay code in a PR is not deployed anywhere
the relay is on-prem; Railway never deploys it in any environment, production included. so this is not a build-packaging gap. a PR adding an op (#409 added `list_buyers_detailed` and `create_buyer` to `_OPS`) still answers `unknown_op` -> `RELAY_OP_UNSUPPORTED` against the workstation's installed build. true e2e for a relay-touching PR needs a packaged build from the branch, since the standing rule is to test through the packaged exe rather than a from-source dev relay.

agreed approach - multi-channel relay + seeded credential
- `[channel] backend_url` accepts a list, one `run_forever` task per URL
- production's connection is never dropped to test a PR
- backend auto-seeds a trusted relay install on non-production environments from an env var set only in the PR environment template, removing the manual provision + enroll per PR
- must refuse to seed when the environment looks like production

open question - which GP company a PR environment writes to. suggested a separate GP database cloned from production.

read-only findings from the GP server
`SY01500` holds
UCSH, UCA, UCI, UHOLD, KEYMA, UBC (production)
TUCSH, TUCI, TUCA, TUBC (test)
TWO (Fabrikam sample)
TEM (Template Company)
- GP company creation is a GP Utilities operation against `TEM`, not a raw restore
- eConnect and WennSoft procs are per-company-database objects (`taCreateBuyer`, `taPoHdr`, `wsiJCJobMaster` live in the company DB), any new company needs them installed
- dev login has `dbcreator=0` and `sysadmin=0`, so this is an IT/DBA request, not something Nexus or the relay can perform
- per-PR clones are not realistic. one shared PR-testing company is, but does not isolate PRs from each other, relocates the shared sandbox away from TUBC. still worth something, because TUBC is where manual testing happens

proposal to keep this unblocked - gate writes per environment, non-production channel defaults to reads only. covers the common case of verifying a dropdown against live GP with no mutation risk. `list_*` served, these refused:
`create_po`
`create_receipt`
`create_job`
`create_buyer`

company choice stays pure config. if a PR-testing company is created later, flip writes on for it, no further code.

relay `allowed_companies` guardrail (TUBC/TUCSH only) already blocks any PR environment from reaching a production GP company.

related
#409 - first change to hit this
#315 - op-set parity gating
#353 PR B - adopt window

Xilous:
Closes #414.

relay dialled exactly one backend, and a PR environment's fresh Postgres had an empty `relay_installs` so the handshake closed 4403. GP-touching changes were only verifiable after merge.

multi-backend channels

`[channel] backend_url` accepts a list or a bare string. `run_forever` is now a supervisor spawning one independent reconnecting channel per URL. per-channel backoff and failure classification. same enrolled secret authenticates on all channels. production's connection is never dropped to test a PR. bare string = one channel, existing config.toml unchanged.

```toml
[channel]
backend_url = [
  "wss://backend-production-7866.up.railway.app/relay-link",
  "wss://backend-pr-414.up.railway.app/relay-link",
]
```

company pin on non-production channels

- full read and write access, pinned to TUBC
- `config.NON_PRIMARY_ALLOWED_COMPANIES = ["TUBC"]`, baked, not operator-editable
- job for any other company returns `company_not_allowed_on_channel` before any handler runs
- production channel identified by matching `config.PRODUCTION_BACKEND_URL`, not list position; match tolerates trailing slash and case, lookalike host never matches
- existing relay `allowed_companies` and backend `relay_call` company check unchanged

seeded credential

new `backend/app/services/relay_seed.py`, called from lifespan startup ahead of the workers. reads `RELAY_SEED_SECRET_HASH`, upserts a relay install carrying that hash, pinned to TUBC.

`relay_installs` stores only the SHA-256 digest, `authenticate_secret` matches on that digest alone -> copy production's digest into a PR environment and the workstation relay authenticates there with the secret it already holds. no re-enrollment.

- refuses when `RAILWAY_ENVIRONMENT_NAME` is production. Railway duplicates variables into PR environments, so the var will be visible in production and the guard cannot depend on its absence
- validates 64 hex chars, normalises case
- idempotent on the hash, repeated boots converge on one row
- never raises, backend still boots

admin ui

`RelayInstallInfo` gains `secretHash`. Admin -> Relay Installs gets a Seed hash column with a copy button.

other

- setup wizard preserves a hand-added `[channel]` block
- desktop app status panel shows one row per channel once more than one is configured
- non-production channel logs a repeated failure once, then at DEBUG

not covered: a PR that adds a relay op still answers `unknown_op` against the workstation's installed build; needs a branch-packaged relay, gap 3 on the issue. PR environments share TUBC with manual testing, no isolation between them, as accepted on the issue.

rollout

1. set `RELAY_SEED_SECRET_HASH` on the Railway environment PR deploys are cloned from, copied from Admin -> Relay Installs.
2. get this relay build onto the workstation through the normal packaged-build flow.
3. for each PR needing relay coverage, add the PR backend's wss URL to config.toml, restart the relay, remove it when the PR closes.

relay 333 pass, frontend 360 pass, backend seed tests run in CI (no local Postgres).


relay/src/ucnexus_relay/config.py
```diff
@@ -90,10 +90,42 @@ class LoggingCfg(BaseModel):
...
+
+def _normalize_url(url: str) -> str:
+    return (url or "").strip().rstrip("/").lower()
+
+
+def is_primary_backend_url(url: str) -> bool:
+    """Whether this channel is the production backend. Tolerates a trailing slash and case so a
+    cosmetic difference in config.toml cannot silently demote production to a restricted channel; a
+    genuinely different host never matches."""
+    return _normalize_url(url) == _normalize_url(PRODUCTION_BACKEND_URL)
```
Xilous, line 113: production channel's unrestricted status hinges on an exact string match against a hardcoded URL. README step 3 has the operator hand-retype that URL into a TOML list.

any deviation returns False:
- transposed digit in `7866`
- `https://` instead of `wss://`
- a regenerated Railway domain
- copying `config.example.toml`'s `backend-production-xxxx` placeholder

`channel_allowed_companies` then pins the production channel to `["TUBC"]`, and every production job for UBC or UCSH is refused with `company_not_allowed_on_channel` before it reaches GP. the only signal is `restricted_to` inside an INFO log line.

keep production implicit - config.toml carries only the extra test URLs, `backend_urls` always prepends `PRODUCTION_BACKEND_URL`. or log a WARNING from `run_forever` when no configured URL is primary.

backend/app/services/relay_seed.py
```diff
@@ -0,0 +1,129 @@
...
+# The GP sandbox every non-production environment is pinned to (#414). Not configurable: a variable
+# here would be one typo away from pointing a PR backend at a live GP company, and the whole reason
+# seeding is acceptable at all is that the blast radius is a sandbox.
+SEED_COMPANY = "TUBC"
+
+_HEX64 = re.compile(r"^[0-9a-f]{64}$")
+
+
+def _is_production(environment_name: str) -> bool:
+    return environment_name.strip().lower() == "production"
```
Xilous, line 64: guard fails open. `RELAY_SEED_SECRET_HASH` lives on production so PR environments inherit it, leaving `RAILWAY_ENVIRONMENT_NAME == "production"` as the only thing keeping production from seeding.

seeding proceeds in production if you:
- rename the base environment
- duplicate it to a staging environment
- lose the injected variable

it inserts a second `relay_installs` row carrying the same `secret_hash` as the real enrolled install but with `company = "TUBC"`. `authenticate_secret` selects with `.first()` and no ORDER BY, so the handshake can bind the seeded row -> `relay_gateway.company` becomes TUBC -> every production GP call fails with "connected relay is enrolled for TUBC, not UCSH".

fail closed - seed only when the environment name positively identifies a PR or preview environment.

backend/app/services/relay_seed.py
```diff
@@ -0,0 +1,129 @@
...
+        return None
+
+    existing = session.scalars(select(RelayInstall).where(RelayInstall.secret_hash == secret_hash)).first()
+    if existing is not None:
+        if existing.company != SEED_COMPANY:
+            logger.warning(
+                "seeded relay install had the wrong company; repointing to the sandbox",
+                extra={"install_id": str(existing.id), "was": existing.company, "now": SEED_COMPANY},
+            )
+            existing.company = SEED_COMPANY
```
Xilous: silently rewrites a genuinely-enrolled install. wherever the variable is set and the DB already holds the real production row, that install is repointed from UBC/UCSH to TUBC:
- a staging environment restored from a prod dump
- a PR env whose DATABASE_URL points somewhere shared
- the production guard above not firing

the original company survives only in a log line, and the row is the live credential.

gate the repair on a row this service created (`label.startswith("seed:")`) and refuse to touch an enrolled non-seed row.

relay/src/ucnexus_relay/channel.py
```diff
@@ -467,3 +521,36 @@ async def run_forever(stop_event: asyncio.Event | None = None) -> None:
...
+        logger.info("channel disabled (no [channel] backend_url configured)")
+        return
+
+    logger.info("starting backend channels", extra={"urls": urls})
+    tasks = [asyncio.create_task(_run_channel(url, stop_event), name=f"channel:{url}") for url in urls]
+    try:
+        # Each _run_channel loops until stop_event; gather returns only on cancellation or a bug. A
+        # channel that somehow raises must not silently take the others down with it, so the exception
+        # is returned rather than propagated, and logged.
+        for url, result in zip(urls, await asyncio.gather(*tasks, return_exceptions=True), strict=True):
```
Xilous: this log line is unreachable. `asyncio.gather(..., return_exceptions=True)` returns only once every awaitable has finished, and `_run_channel` loops until `stop_event`, so one channel raising leaves gather still awaiting the survivors. on shutdown gather raises CancelledError and the `for` loop is skipped entirely.

a channel that dies is invisible - no log line, no restart, just a stale "disconnected" on /health.

attach a done callback per task, or wrap each `_run_channel` in a logging shim.

relay/src/ucnexus_relay/channel.py
```diff
@@ -434,14 +485,11 @@ async def run_forever(stop_event: asyncio.Event | None = None) -> None:
         settings = get_settings()
```
Xilous: the per-attempt config re-read sits outside the try, so any config.toml problem kills this channel permanently rather than being retried. hand-editing `[channel] backend_url` while the relay runs is the documented rollout step, and every channel re-reads on every reconnect attempt.

these raise `tomllib.TOMLDecodeError` or a pydantic `ValidationError` out of `_run_channel`, production's included:
- a save caught mid-write
- a missing quote
- a bracket typo

fixing the file does not self-heal it, only a serve restart does. with the gather issue below, it fails silently.

move `cache_clear` and `get_settings` inside the try and treat a bad config as a retryable failure category.

relay/src/ucnexus_relay/channel.py
```diff
@@ -280,15 +310,32 @@ def _run_create_receipt(company: str, payload: dict) -> dict:
 }
 
 
-def _dispatch(op: str, company: str, payload: dict) -> dict:
+def _dispatch(op: str, company: str, payload: dict, allowed_companies: list[str] | None = None) -> dict:
```
Xilous, line 332: `allowed_companies` defaults to None, which means unrestricted, on both `_dispatch` and `_handle_job`. the docstring calls this gate "the sole thing keeping that safe", so unrestricted is the wrong default: any future call path that forgets to thread it through gets full production-company access.

no test covers the wiring - the tests pass `channel_allowed_companies(PR_URL)` into `_dispatch` by hand, so a regression in `_run_once` dropping the argument would pass CI.

make the parameter required, or default it to the restricted list and have production pass None explicitly.

relay/src/ucnexus_relay/config.py
```diff
@@ -90,10 +90,42 @@ class LoggingCfg(BaseModel):
...
+# THE production backend. Identity - not list position - is what makes a channel the primary one, so
+# reordering backend_url can never accidentally hand a test backend unrestricted company access.
+PRODUCTION_BACKEND_URL = "wss://backend-production-7866.up.railway.app/relay-link"
+
+# What a NON-PRIMARY channel (a Railway PR environment, a local dev backend) may target. Reads AND
+# writes are served on those channels - that is the whole point, a PR that touches GP has to be
+# verifiable before it merges - but only against the sandbox company, so the worst a test backend can
+# do is write to TUBC. Baked deliberately: an operator-editable value here would be one typo away from
+# pointing a PR backend at a live GP company, which is the only thing making this safe (#414).
+NON_PRIMARY_ALLOWED_COMPANIES = ["TUBC"]
```
Xilous, line 102: a test channel is pinned to TUBC here, but the handler it reaches calls `ops.check_company_allowed`, which requires the company to be in `[gp] allowed_companies` from config.toml. a workstation configured for production work (`allowed_companies = ["UBC", "UCSH"]`) refuses every TUBC job with `company_not_allowed`, so every relay call from a PR environment errors out.

neither the README rollout steps nor testing/CLAUDE.md mention adding TUBC to `allowed_companies`, and nothing checks the two lists intersect at startup.

relay/src/ucnexus_relay/channel.py
```diff
@@ -28,37 +40,55 @@
...
+    and neither should start reporting on a throwaway PR-environment channel. `channels` carries the
+    per-URL detail for the app to render when more than one is configured. With no primary configured
+    (a dev checkout pointed at localhost) the first channel stands in, so the app is never blank."""
+    primary_url = next((url for url in _STATES if is_primary_backend_url(url)), None)
+    if primary_url is None:
+        primary_url = next(iter(_STATES), None)
+    snapshot = dict(_STATES.get(primary_url, _UNKNOWN_STATE) if primary_url else _UNKNOWN_STATE)
     snapshot["jobs_in_flight"] = _INFLIGHT
     snapshot["last_job_finished_ago"] = None if _LAST_JOB_AT is None else time.monotonic() - _LAST_JOB_AT
+    snapshot["channels"] = [
```
Xilous: `_STATES` is now iterated here, and the relay's `/health` is a sync `def` (main.py:89), so Starlette runs it on a threadpool worker while the event loop is adding keys from `_mark_connected` and `_mark_disconnected`. a snapshot overlapping a channel's first report raises `RuntimeError: dictionary changed size during iteration` -> /health 500s -> `ui.relay_health` returns `{"running": False}` -> app shows "not running" and `update_poller.is_busy` defers the update.

the old `_STATE` had fixed keys and was only copied, so there was nothing to iterate.

pre-seed `_STATES` with every configured URL in `run_forever` so the dict never changes size.

backend/app/services/relay_seed.py
```diff
@@ -0,0 +1,129 @@
...
+
+    if not _HEX64.match(secret_hash):
+        logger.error(
+            "RELAY_SEED_SECRET_HASH is not a SHA-256 hex digest (expected 64 hex chars); skipping "
+            "relay credential seeding. Copy the value from Admin -> Relay Installs.",
+            extra={"length": len(secret_hash)},
+        )
+        return None
+
+    existing = session.scalars(select(RelayInstall).where(RelayInstall.secret_hash == secret_hash)).first()
```
Xilous, line 117: read-then-insert with no unique constraint on `relay_installs.secret_hash` and no ON CONFLICT. two containers booting at once (Railway overlaps deployments) can both find nothing and both insert, leaving two rows for the same digest.

rotating `RELAY_SEED_SECRET_HASH` after the workstation re-enrolls adds a new row and leaves the previous `seed:<env>` row in place, so the superseded secret keeps authenticating in that environment.

retire prior `seed:` rows for the same label, and consider a unique index on `secret_hash`.

relay/src/ucnexus_relay/ui.py
```diff
@@ -60,7 +81,11 @@ def config_summary(config_path: str | Path | None = None) -> dict:
         "allowed_companies": gp.get("allowed_companies") or [],
         "sql_server": sql.get("server"),
         "odbc_driver": sql.get("driver"),
-        "backend_url": channel.get("backend_url") or "",
+        # Singular stays the PRIMARY (production) URL, not merely the first configured one: it is what
+        # the enroll wizard derives its GraphQL endpoint from, and enrolling against a PR environment
+        # would rotate the workstation's secret away from production.
+        "backend_url": _primary_backend_url(backend_urls),
+        "backend_urls": backend_urls,
```
Xilous: nothing reads `backend_urls` - the status JS reads `s.channel.channels`. those per-channel rows only render when `chans.length > 1`, and `_STATES` fills in lazily, so with serve stopped (`channels: []`) or before the second channel's first attempt the panel shows only the production URL. the operator who just hand-added a PR URL gets no confirmation it was picked up.

render `c.backend_urls` in the panel or drop the key.

frontend/src/modules/admin/__tests__/RelayInstallsPage.test.tsx
```diff
@@ -189,3 +190,34 @@ it('disables Remove on the install that is currently connected', async () => {
   // The adopt action stays available - it is the recovery path, not a destructive one.
   expect(screen.getByRole('button', { name: /adopt next connection/i })).not.toBeDisabled();
 });
+
+it('copies the full seed hash for the PR-environment variable', async () => {
+  // #414: RELAY_SEED_SECRET_HASH is what lets a Railway PR environment accept this relay without a
+  // provision + enroll cycle. The grid truncates it to stay readable, so the copy must carry the whole
+  // digest - a truncated one silently never matches on the handshake.
+  const copied: string[] = [];
+  Object.defineProperty(navigator, 'clipboard', {
```
Xilous, line 202: `Object.defineProperty(navigator, 'clipboard', ...)` is never restored, so the stub leaks into every test that runs after this one in the file.

use `vi.spyOn(navigator.clipboard, 'writeText')` with an `afterEach` restore, or save and reinstate the original descriptor.

relay/tests/conftest.py
```diff
@@ -14,6 +14,22 @@
 
 import pytest
 
+from ucnexus_relay import channel
```
Xilous: importing `channel` at conftest module scope pulls `pyodbc` and `websockets` into every relay test session, for a fixture five tests use. if either fails to import on a runner, the whole suite errors at collection instead of just the channel modules.

import inside the fixture body.

relay/src/ucnexus_relay/ui.py
```diff
@@ -20,9 +20,23 @@
...
+    KNOWN_COMPANIES,
+    SqlCfg,
+    is_primary_backend_url,
+)
+
+
+def _primary_backend_url(urls: list[str]) -> str:
+    """The production URL among the configured channels, falling back to the first one so a dev
+    checkout pointed only at localhost still shows (and can enroll against) something."""
+    return next((u for u in urls if is_primary_backend_url(u)), urls[0] if urls else "")
```
Xilous: duplicates `channel_state_snapshot`'s primary-selection rule (channel.py:74-76) - find the primary, else fall back to the first. two copies drift, one over a list, one over `_STATES` keys.

a `primary_url(urls)` helper in config.py next to `is_primary_backend_url` would serve both.

relay/src/ucnexus_relay/channel.py
```diff
@@ -280,15 +310,32 @@ def _run_create_receipt(company: str, payload: dict) -> dict:
...
     if handler is None:
         return {"ok": False, "error": errors.error_body("unknown_op", f"unknown op {op!r}")}
     if not company:
         return {"ok": False, "error": errors.error_body("missing_company", "company is required")}
+    if allowed_companies is not None and company not in allowed_companies:
+        return {
+            "ok": False,
+            "error": errors.error_body(
+                "company_not_allowed_on_channel",
+                f"{company} is not reachable from a non-production backend; only {allowed_companies} is",
```
Xilous: `{allowed_companies}` interpolates a Python list, so the message reads `only ['TUBC'] is`. `relay_gateway.relay_call` surfaces `error["message"]` verbatim as the GraphQL error the user sees.

join it: `', '.join(allowed_companies)`.

Issue #424, opened 2026-07-29
OPS: set TESTING_ENABLED=false on the production Railway backend


Xilous:
ops action, no code change. do before the code fixes in #422.

TESTING_ENABLED is true on the production Railway backend service. probe GET https://backend-production-7866.up.railway.app/testing/clerk-sign-in?email=<nonexistent> -> 404 No user found, not 403 Testing is not enabled.

two endpoints gated on that one variable, together a full production data loss path from an unauthenticated start.
- GET /testing/clerk-sign-in (backend/main.py:381) - TESTING_ENABLED only, no auth. mints a real Clerk sign-in token for any email, including Admin/Manager accounts.
- POST /admin/reset-data (backend/main.py:292) - TESTING_ENABLED plus require_admin_request. drops and rebuilds the entire public schema.

unauthenticated caller mints admin token with the first -> satisfies the second. chain not exercised, read off both routes plus the one TESTING_ENABLED probe.

set TESTING_ENABLED=false on the production backend service in Railway. closes both endpoints. reversible. restarts the service, stops the Clerk sign-in helper working against production.

keep it true on PR environments - that is where end-to-end testing belongs and the sign-in helper is needed there.

longer term, `/testing/clerk-sign-in` should not ship in a production-capable build at all.

Xilous:
done. TESTING_ENABLED=false set on production backend service via Railway MCP, deployment 45c1c309 SUCCESS.

GET /testing/clerk-sign-in?email=<nonexistent address> -> 403 {"error":"Testing is not enabled"} (was 404 No user found)

both endpoints gated on the variable are now closed on production.

Railway PR envs clone their variables from production at creation (#431). PR envs created from now on inherit TESTING_ENABLED=false - set TESTING_ENABLED=true directly on that env's backend service for the Clerk sign-in helper. existing PR envs keep what they already have.

endpoint authentication tracked in #422.

PR #433, opened 2026-07-30
fix(security): require a credential on /testing/clerk-sign-in (#422)

### LINKED ISSUE #422: SECURITY (production): /testing/clerk-sign-in mints an admin Clerk session for any email, unauthenticated - chains with /admin/reset-data to full data loss
`GET /testing/clerk-sign-in` (`backend/main.py:381`) mints a Clerk sign-in token for an arbitrary email, unauthenticated. only gate is `TESTING_ENABLED`.

```python
@app.get("/testing/clerk-sign-in")
def get_clerk_sign_in_token(email: str = "jayp@ucsh.com"):
    if not TESTING_ENABLED:
        return JSONResponse(status_code=403, ...)
    # looks the email up in Clerk, POSTs /v1/sign_in_tokens, returns {token, url, user_id}
```

`email` is a plain query param. no `require_admin_request`, no shared secret, no allowlist of impersonable accounts.

live on the shared backend, not just PR environments. verified without minting a real token, by asking for an address that does not exist:

```
$ curl -s -w "%{http_code}" "https://backend-production-7866.up.railway.app/testing/clerk-sign-in?email=definitely-no-such-user-probe@example.invalid"
404 {"error":"No user found with email definitely-no-such-user-probe@example.invalid"}
```

404 not 403 -> `TESTING_ENABLED` true there, endpoint serving. a real admin address returns a working sign-in token.

anyone who knows the URL can sign in as any UC Nexus user, Admin/Manager included. defeats the #415 resolver gate sweep - every `require_admin` resolver is reachable after minting a token here:
updateUserRoles
deleteRelayInstall
mergeLocations
deleteWarehouse

PR environments inherit production Clerk keys, so a token minted against a PR environment is a real Clerk session.

`POST /admin/reset-data` (`main.py:292`) takes both `TESTING_ENABLED` and `require_admin_request(request)`. this route got only the first.

fixes, most preferred first
- add `require_admin_request(request)`, so an existing admin session is needed to mint another one
- and/or require a shared secret header known only to CI and local test runs
- and/or restrict `email` to an allowlist of dedicated test accounts, so real staff accounts can never be impersonated
- decide whether `TESTING_ENABLED` should be true on the shared backend at all

`backend/tests/test_resolver_gate_completeness.py` (#415) walks `app/schemas/` only, misses plain FastAPI routes in `main.py`. extend the sweep to cover them.

found during code review of #420 (issue #415).

Xilous:
Closes #422.

GET /testing/clerk-sign-in had only the TESTING_ENABLED environment gate - no auth. every environment shares the production Clerk instance, so it minted a real Clerk session for any staff email, Admin/Manager included. #424 flipped TESTING_ENABLED off on production as the ops stopgap; this adds the auth that should have been there regardless.

the route now requires, on top of TESTING_ENABLED:
- an existing Admin/Manager session (require_admin_request, same gate as /admin/reset-data), or
- the shared testing secret in an X-Testing-Secret header, verified against new TESTING_SIGN_IN_SECRET_HASH env var (SHA-256 hex digest). bootstrap path for a fresh PR environment where no session exists yet. same verifier-not-credential shape as RELAY_SEED_SECRET_HASH (#414): Railway stores only a digest. constant-time compare (hmac.compare_digest). blank hash disables the secret path - fails closed.

TESTING_ENABLED stays first, so a production deployment refuses outright rather than leaking whether the credential would have been good enough - same ordering as the reset route.

the #415 gate-completeness sweep walked app/schemas/ only, which is why this route escaped it. it now also collects every @app.<method> route in main.py and asserts each calls require_admin_request or sits on a route exemption list with a reason (/health, /relay-link). no ordering assertion for routes - environment-gate-first is correct there.

new test_testing_sign_in_auth.py: refusal when TESTING_ENABLED false, no credential, wrong secret, garbage bearer, blank-hash fail-closed, success path with correct secret (Clerk Backend API faked).

testing/CLAUDE.md + backend/.env.example updated: testers on a PR environment set both TESTING_ENABLED=true and a TESTING_SIGN_IN_SECRET_HASH they know the preimage of (new PR envs inherit TESTING_ENABLED=false from production since #424); the sign-in fetch carries X-Testing-Secret. .env.example includes a generator one-liner.

ruff check + format clean. 331 tests pass locally across the sweep/auth-route files; full backend suite is CI-only.


Xilous:
verified live on the pr-433 environment (TESTING_ENABLED=true + TESTING_SIGN_IN_SECRET_HASH set on its backend, redeployed):

no credential -> 401 UNAUTHENTICATED
wrong X-Testing-Secret -> 401 UNAUTHENTICATED
garbage bearer -> 401 Malformed authentication token
correct X-Testing-Secret -> auth passed, 404 No user found for a nonexistent probe address (no real token minted)

before the variables were set the env returned 403 Testing is not enabled, confirming the #424 inheritance note: post-#424 PR environments inherit TESTING_ENABLED=false and need both variables set to test.

PR #445, opened 2026-07-30
feat(gp): create customer addresses from the create-job dialog (#444)

### LINKED ISSUE #444: Create GP customer addresses from the create-job dialog
job address and bill-to pickers in the create-job dialog are dropdowns; wsiJCJobMaster validates the address code against RM00102 for that customer. a site address missing from GP blocks job create until someone adds it in GP (Cards > Sales > Addresses).

eConnect owns the write, dbo.taCreateCustomerAddress. verified live in
TUBC
TUCSH
UBC
UCSH

- proc plus Pre/Post hooks present
- DYNGRP has EXECUTE
- effective EXECUTE confirmed for the relay's SSPI login
- 37 params matching the reference XSD
- no OnlyValidate param, no dry run; duplicate guard is our own pre-check
- UpdateIfExists pinned to 0 - Nexus creates, never overwrites what accounting maintains

decisions
- add-address affordance on both address pickers (job address, bill-to), each scoped to its own customer
- user types the ADRSCODE, uppercased, max 15 chars, duplicate-checked against that customer's existing codes
- createGpCustomerAddress gated ADMIN_ROLE, same as createGpJob

site fields only
code (required)
address 1 (required)
address 2
city (required)
province/state
postal code
country

relay
- econnect.py customer_address_exists, read-only RM00102 probe
- econnect.py create_customer_address, EXEC dbo.taCreateCustomerAddress with UpdateIfExists=0 hardcoded
- econnect.py get_customer_address reads back the stored row, response reports what GP stored
- models.py CreateCustomerAddressRequest validates the live column widths (code 15, address lines 60, city 35, state 29, zip 10, country 60), over-length rejected rather than truncated
- ops.py create_customer_address_op mirrors create_job_op minus the dry run: pre-check (address_code_already_exists, also makes a retry after a lost reply safe) -> real call -> read back -> caller commits
- channel.py _OPS["create_customer_address"]; non-prod channels already pinned to TUBC with writes allowed (#414)

backend
- CreateGpCustomerAddressInput
- create_gp_customer_address mutation on RelayMutations returning GpCustomerAddress
- ROOT_FIELD_POLICY["createGpCustomerAddress"] = ADMIN_ROLE
- no Postgres write, no actor recording

frontend, CreateGpJobDialog
- "+ Add new address" MenuItem at the bottom of each address select, opens a nested dialog bound to that picker's customer
- on success refetch that picker's gpCustomerAddresses query, auto-select the new code, toast
- GP rejections surface through the existing GpErrorAlert path
- duplicate code shows the error without closing the dialog

this hands admins the ability to create Receivables master data accounting currently owns in GP. create-only, no overwrite path. accounting may want a say before this ships.

Xilous:
Closes #444.

"+ Add new address" row on the job address and bill-to address selects in the create-job dialog, opening a nested dialog that creates the code under that picker's customer.

relay
- create_customer_address duplicate pre-check (address_code_already_exists) -> EXEC dbo.taCreateCustomerAddress -> read stored row back from RM00102
- UpdateIfExists=0 is a literal in the SQL, not a parameter - Nexus creates addresses, it never overwrites master data accounting maintains in GP
- CreateCustomerAddressRequest trims, uppercases the code, rejects over-length values against GP's live column widths (15/15/60/60/35/29/10/60)
- response carries GP's stored row, same shape list_customer_addresses serves the picker

backend
- createGpCustomerAddress mutation on RelayMutations, returns GpCustomerAddress
- ROOT_FIELD_POLICY entry ADMIN_ROLE, same as createGpJob
- no company argument, resolves the connected relay's company the way createGpJob does
- nothing persisted in Nexus, the address lives only in GP

frontend
- "+ Add new address" MenuItem pinned last in both address selects, behind a sentinel value intercepted in onChange, never stored as a selection
- nested AddCustomerAddressDialog scoped to that picker's customer; on success refetch that picker's own gpCustomerAddresses -> select the new code
- GP rejections and duplicates surface inside the nested dialog through GpErrorAlert, typed input left intact

35 new relay tests, 9 new backend tests, 4 new dialog tests. Full suites pass locally - relay 410, backend 900 passed with two pre-existing test_manufacturer_match failures unrelated to this change, frontend 394.

taCreateCustomerAddress has no OnlyValidate parameter, unlike wsiJCJobMaster, so there is no dry-run pass. The pre-check is the entire duplicate guard, which is why UpdateIfExists staying pinned to 0 matters.


frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -340,6 +359,25 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
...
+   * code that is not in that list yet leaves the field blank until the read lands. `catch` keeps a
+   * failed re-read from stranding the selection: the address exists in GP either way.
+   */
+  const handleAddressCreated = useCallback(
+    (addressCode: string) => {
+      const forBill = addAddressFor === 'bill';
+      const refetch = forBill && billScopedToOwnCustomer ? refetchBillAddresses : refetchAddresses;
+      const select = forBill ? setBilltoAddressCode : setJobAddressCode;
+      void refetch()
+        .catch(() => undefined)
```
Xilous: `.catch(() => undefined)` turns a failed refetch into a resolved value, so `select(addressCode)` still runs. the code is not among the select's MenuItem options at that point; a MUI TextField select holding an out-of-range value renders blank, `requiredComplete` counts `jobAddressCode !== ''` as satisfied, and Create Job stays enabled over a required field the user sees as empty. the failed refetch also sets the query's error state, so the read-error banner can appear right after the success toast. the likeliest trigger is the relay dropping right after the create commits. the comment above this line says the catch keeps a failed re-read from stranding the selection; it produces the stranded state. GpBuyerSelect solved this with a held-value fallback (RegisterGpBuyerDialog.tsx). either only select when the refetch succeeded, or splice the mutation's returned row (addressCode/address1/city/state) into the options.

frontend/src/modules/import/AddCustomerAddressDialog.tsx
```diff
@@ -0,0 +1,231 @@
...
+      // else would leave the picker holding a code the job proc goes on to reject.
+      const created = response.data?.createGpCustomerAddress?.addressCode ?? addressCode.trim();
+      showToast(`Address ${created} added to customer ${customer.customerNumber}.`, 'success');
+      onCreated(created);
+      reset();
+      onClose();
+    } catch (err) {
+      // GP's own words - most often a code the customer already has. The dialog stays open with the
+      // typed address intact, so fixing it is one edit rather than re-keying the whole thing.
+      setGpError(extractGpError(err));
```
Xilous: on `address_code_already_exists` the catch only sets the error alert; nothing refetches the address list. both address queries are `fetchPolicy: 'cache-first'`, so the address exists in GP, the cached list lacks it, and the create is permanently refused. the job picker can escape via Refresh GP data after cancelling; a bill-to picker scoped to its own customer has no in-dialog escape (refreshReads never touches that query), only a page reload. RegisterGpBuyerDialog.tsx refetches the buyer list on `buyer_already_exists`; do the same here - when `gp.relay?.error === 'address_code_already_exists'`, refetch the picker's list.

frontend/src/modules/import/AddCustomerAddressDialog.tsx
```diff
@@ -0,0 +1,231 @@
...
+            />
+          </Stack>
+        </Stack>
+      </DialogContent>
+      <DialogActions>
+        <Button onClick={handleClose} disabled={loading}>
+          Cancel
+        </Button>
+        <Button variant="contained" onClick={handleSubmit} disabled={loading || !requiredComplete}>
+          {loading ? 'Adding…' : 'Add address'}
```
Xilous, line 279: the nested dialog gates only on `loading`, not on the relay. the parent applies `disabled = !relayConnected || loading` to every GP-touching control with a "The GP relay is not connected" banner, and useRelayStatus polls every 10s, so the relay can drop while this dialog is open mid-entry. the parent behind the modal disables itself; these fields and the Add address button stay live, and the submit lands on RelayUnavailableError rendered as a red GP error. pass `relayConnected` down, fold it into `disabled`, and add the parent's plain-language banner.

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -249,7 +261,11 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
   // customer and rejects - the form could not express a valid pairing at all.
   const billCustomerNumber = billCustomer?.customerNumber ?? null;
   const billScopedToOwnCustomer = billCustomerNumber !== null && billCustomerNumber !== customer?.customerNumber;
-  const { data: billAddressesData, loading: billAddressesLoading } = useQuery<{
+  const {
+    data: billAddressesData,
+    loading: billAddressesLoading,
+    refetch: refetchBillAddresses,
```
Xilous, line 299: refreshReads never calls the `refetchBillAddresses` this PR just destructured. it refetches only:
customers
divisions
tax schedules
employees
the job customer's addresses
when billScopedToOwnCustomer is true, the bill query is a separate cache entry (`cache-first`, distinct customer variable) that nothing re-reads, so the bill-to picker keeps serving the pre-refresh list after a refresh. one line fixes it, `if (billScopedToOwnCustomer) void refetchBillAddresses();` plus the dep entry.

backend/app/schemas/relay.py
```diff
@@ -300,6 +300,67 @@ async def create_gp_buyer(
...
+        this can never overwrite an address accounting maintains in GP.
+        """
+
+        company = relay_gateway.company
+        if not company:
+            raise RelayUnavailableError(
+                "The GP relay is not connected, so this address cannot be created in GP. Start the relay and try again."
+            )
+
+        payload = {
```
Xilous, line 372: the mutation forwards every field raw, with no trim, blank check, or width bound. a relay-side rejection surfaces as `errors.error_body("invalid_payload", str(e))`, and str() of a pydantic v2 ValidationError is the multi-line dump with the input repr and the errors.pydantic.dev URL, rendered verbatim by GpErrorAlert with no `field` anchor. the dialog's maxLength attributes shield UI users; direct admin API callers and any future consumer are exposed. mirror create_gp_buyer fifty lines up - strip and bound against the same widths, raising field-scoped ValidationErrors.

relay/src/ucnexus_relay/ops.py
```diff
@@ -279,6 +279,50 @@ def create_job_op(conn, *, company: str, request: models.CreateJobRequest) -> mo
...
+
+    It also makes a retry after an ambiguous failure safe: if the first attempt committed and the reply
+    was lost, pressing Create again reads as "that customer already has this address code" rather than
+    as a raw eConnect state for something that already worked.
+
+    The read-back guards the err=0-but-nothing-landed case create_po_line has a known mode for, and
+    answers with GP's stored row - which is exactly what list_customer_addresses will serve the picker
+    on its next refetch, so answering with anything else could hand the dialog an address that differs
+    from the one everything else is about to see."""
+    if econnect.customer_address_exists(conn, request.customer_number, request.address_code):
```
Xilous, line 318: the duplicate pre-check has a check-then-act window - a plain SELECT COUNT under READ COMMITTED holds no lock past the statement, and the backend dispatches relay calls concurrently, so two simultaneous creates of the same code (or a race against someone saving the same code in GP's own window) both pass. the loser reaches taCreateCustomerAddress, whose duplicate refusal comes back as a raw econnect_error state; nothing maps that proc state to `address_code_already_exists`. the read-back never compares the stored row to the request; it proves a row with that code exists, not that it is this call's row. the docstrings here and in econnect.py call the pre-check "the entire duplicate guard", true only single-threaded. either map the proc's duplicate state to address_code_already_exists or correct the claim.

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -583,6 +634,11 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
                   {addressLabel(a)}
                 </MenuItem>
               ))}
+              {billPickerCustomer && (
```
Xilous, line 744: the "+ Add new address" row in the bill-to select gates only on `billPickerCustomer`, not on the addresses read having succeeded, and the bill-scoped useQuery never destructures `error`, so a failed bill read feeds nothing into readError. bill read fails -> picker enabled -> only row is "+ Add new address" -> no banner. the user re-creates an address the customer already has, and the pre-check refuses it against an empty list on screen. bind the bill query's `error` into readError; consider gating the add row on the read having succeeded.

backend/app/schemas/relay.py
```diff
@@ -300,6 +300,67 @@ async def create_gp_buyer(
...
+            # GP said no - a customer that does not exist, a code already on this customer, a value GP
+            # will not take. The relay words those better than a generic failure would, and the detail
+            # body carries the proc + error state the dialog's error alert renders, which is what
+            # validation_error_from_relay preserves.
+            raise validation_error_from_relay(e) from e
+
+        # GP's stored row, read back from RM00102 by the relay. That row is what gpCustomerAddresses
+        # will serve on its next refetch, so answering with anything else could hand the dialog an
+        # address that differs from the one it is about to see in the picker.
+        address = (result or {}).get("address") or {}
```
Xilous: `(result or {}).get("address") or {}` and the `or input.address_code` fallback are unreachable - CreateCustomerAddressResponse requires `address`, and create_customer_address_op raises when the read-back is None. a reply lacking the key would return a well-formed success echoing the caller's raw un-uppercased code with address1/city/state all null; the dialog selects it while the refetched list holds GP's uppercased version, stranding a blank select. every sibling resolver in this file subscripts bare (`result["addresses"]`, `result["buyers"]`). use `address = result["address"]`.

relay/src/ucnexus_relay/channel.py
```diff
@@ -305,6 +305,28 @@ def _run_create_job(company: str, payload: dict) -> dict:
...
 
+def _run_create_customer_address(company: str, payload: dict) -> dict:
+    """Add an address code to a GP customer (#444) - the write half of _run_list_customer_addresses.
+
+    The customer arrives under the key `customer`, as it does on the read op: the two halves of the same
+    picker should not disagree about what the customer key is called. `customer_number` is accepted too,
+    since that is the name the model and the proc parameter use."""
+    ops.check_company_allowed(company)
+    fields = dict(payload)
+    customer = fields.pop("customer", None)
```
Xilous, line 325: the customer/customer_number aliasing is silently lossy. a payload carrying both keys has `customer` overwrite `customer_number` with no error. `customer: ""` passes the `is not None` check and ends as a pydantic invalid_payload dump; _run_list_customer_addresses answers RelayOpError("missing_customer", "customer is required") for the same mistake. no other handler in this file rewrites keys, the only real caller always sends `customer`, and nothing tests this handler. either send customer_number from the backend and make this the standard one-liner `Model(company=company, **payload)`, or declare the alias on the pydantic field (AliasChoices).

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -548,7 +586,12 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
               select
               label="Job address"
               value={jobAddressCode}
-              onChange={(e) => setJobAddressCode(e.target.value)}
+              // #444: the add row arrives as a value like any option, so it is intercepted here and
+              // never stored - it opens the nested dialog and leaves the current selection alone.
+              onChange={(e) => {
+                if (e.target.value === ADD_ADDRESS) setAddAddressFor('job');
```
Xilous: the add-row affordance is four copy-paste sites:
this onChange interceptor and its bill twin
the two MenuItem blocks
this file's own EmployeeField docstring: "a fix applied to a copy-pasted twin is a fix that silently misses one of them". only the job picker has the test asserting the sentinel is never stored. the `'__add__'` sentinel is excluded only by these two interceptors; requiredComplete and the submit payload pass it through, the relay's CreateJobRequest accepts it (7 chars, non-blank), and the only rejector is GP's proc. extract an AddressSelect owning the option list, the add row, and the interception.

frontend/src/graphql/import.ts
```diff
@@ -149,6 +149,20 @@ export const GET_GP_CUSTOMER_ADDRESSES = gql`
   }
 `;
 
+// Issue #444: an address the job needs but GP does not hold yet. No company argument - the backend
+// uses the connected relay's own company, the same one the reads above are keyed on. The selection
+// is the address code, so the returned row is exactly what the picker needs to offer it immediately.
+export const CREATE_GP_CUSTOMER_ADDRESS = gql`
+  mutation CreateGpCustomerAddress($input: CreateGpCustomerAddressInput!) {
+    createGpCustomerAddress(input: $input) {
```
Xilous, line 159: the mutation selects address1/city/state that nothing consumes - the hook generic in AddCustomerAddressDialog types the result as `{ addressCode: string }`, only addressCode is read, and no cache write puts the row into any list. the comment says the returned row is "exactly what the picker needs to offer it immediately"; the picker offers nothing until the separate refetch lands. either use the returned row (splice it into the picker's options or write it into the cache), or trim the selection and the comment to addressCode.

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -267,6 +283,9 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
     () => (billScopedToOwnCustomer ? (billAddressesData?.gpCustomerAddresses ?? []) : addresses),
     [billScopedToOwnCustomer, billAddressesData, addresses],
   );
+  // #444: and so does an address created FROM that picker - same rule, same reason.
+  const billPickerCustomer = billScopedToOwnCustomer ? billCustomer : customer;
```
Xilous, line 319: the mapping from which picker asked to {customer, refetch, setter} is spelled three ways that must agree by inspection:
billPickerCustomer here
the addAddressCustomer nested ternary below it
handleAddressCreated's `forBill && billScopedToOwnCustomer` branches
all four (addAddressFor, billScopedToOwnCustomer) combinations agree today. the null case already diverges between spellings (addAddressCustomer maps null to null; forBill folds null into the job branch), reset() clears 17 pieces of state and omits setAddAddressFor(null), and the bill-picker-with-no-bill-customer combination has no test. capture {customer, refetch, select} once when the add row is clicked.

relay/src/ucnexus_relay/models.py
```diff
@@ -371,3 +371,88 @@ class CreateJobResponse(BaseModel):
...
+    "address1": 60,
+    "address2": 60,
+    "city": 35,
+    "state": 29,
+    "zip_code": 10,
+    "country": 60,
+}
+
+
+class CreateCustomerAddressRequest(BaseModel):
```
Xilous, line 398: CreateCustomerAddressRequest sets no model_config, so pydantic v2 defaults to extra="ignore"; any key the model does not declare is silently dropped when the channel constructs the request, and econnect.create_customer_address's unknown-field guard is unreachable from the channel path (ops.py passes model_dump, which only emits the eight declared keys). the hello-frame handshake is op-name parity only and the relay deploys separately from the backend, so a newer backend sending a new optional key to an older relay has it vanish with success reported. create_job carries the identical unreachable guard; hardening for all the relay request models, not a regression here. extra="forbid" turns the skew into a loud invalid_payload.

frontend/src/modules/import/AddCustomerAddressDialog.tsx
```diff
@@ -0,0 +1,231 @@
...
+  ]);
+
+  const customerLabel = customer.customerName
+    ? `${customer.customerNumber} - ${customer.customerName}`
+    : customer.customerNumber;
+
+  return (
+    <Dialog open={open} onClose={loading ? undefined : handleClose} maxWidth="sm" fullWidth>
+      <DialogTitle>Add a Customer Address</DialogTitle>
+      <DialogContent>
```
Xilous, line 172: no signal that this write is irreversible from Nexus. the relay pins UpdateIfExists to 0 and _OPS carries no update or delete op for an address. a typo'd code lands permanently in RM00102, appears in that customer's picker for every future job, and only GP can clean it up. RegisterGpBuyerDialog opens with an info Alert for its equivalent write; add the same Alert here before the first keystroke.

frontend/src/modules/import/__tests__/CreateGpJobDialog.test.tsx
```diff
@@ -466,3 +466,224 @@ test('a successful submit sends only the optional fields that were filled in', a
...
+          state: null,
+          zipCode: null,
+          country: null,
+        },
+      },
+    },
+    result: {
+      errors: [
+        new GraphQLError('Customer ELL100 already has an address code MAIN', {
+          extensions: { code: 'address_code_already_exists' },
```
Xilous: the mock's shape is one the backend never produces. the real duplicate path emits extensions.code = "VALIDATION_ERROR" with the relay body under extensions.relayError = {error: "address_code_already_exists", message, context}. extractGpError reads ext.code and ext.relayError separately; under the real shape GpErrorAlert renders the code and relay detail rows, under this mock only the message. the test passes on message text alone, and a regression that drops extensions.relayError would not be caught. mock the real shape.

Issue #427, opened 2026-07-30
Audit trail records the client-supplied actor, not the authenticated caller


Xilous:
`require_user(info)` returns `{user_id}`, call sites throw it away. the field recording who did it comes from the client.

signed-in user can attribute their own action to somebody else. `completePullRequest(id, completedBy: "Someone Else")` stamps that name on every opening the pull staged. same shape for
`performed_by` on inventory corrections
`picked_by` on pick mutations

`resolve_display_name(user_id)` already exists at `app/auth.py:133` (#199, so a receive records the Clerk-authenticated caller rather than the relay's Windows account), not applied to the rest.

`adjust_inventory_quantity` (`app/repositories/warehouse/inventory.py:357`) takes no `performed_by` at all, hardcodes `performed_by="Admin/Manager"` (inventory.py:381). every quantity adjustment is attributed to an admin no matter who made it, and there is no parameter to pass the truth through even if the resolver wanted to.

`performed_by` is read by
`auditLog`
location history panel
recent-activity feed

none of it answers "who did this". it answers "who did the client say did this", and for inventory adjustments not even that.

pre-existing, not introduced by #415. #415 sharpened it: every resolver now authenticates its caller, so the true identity is sitting in the return value at the moment the wrong one gets written.

take the actor from the gate, not the input.
- feed `require_user`'s return value into `resolve_display_name(user_id)` at call sites that record an actor
- give `adjust_inventory_quantity` a `performed_by` parameter like its siblings already have
- drop client-supplied `completedBy` / `performedBy` / `pickedBy` from the schema, or keep only as a display hint never written to an audit row
- test in the shape of `test_resolver_gate_completeness.py`: a resolver that writes an audit row must not take its actor from user input

found in code review of #420 (issue #415).

PR #449, opened 2026-07-31
feat(gp): provision the new job's cost codes from GP's master and division setup

### LINKED ISSUE #448: create-project provisions the job's cost codes from GP's cost-code master and division account setup
create-project (Start a Task) -> relay create_job -> wsiJCJobMaster writes only the JC00102 job-master row, assigns no cost codes.

all four relay-created TUBC test jobs have zero JC00701 rows:
NEXUS-380-T1
NEXUS-380-T2
NEXUS-392-T1
NEXUS-444-T1

GP-UI-created jobs 80001-80005 carry 4-9 codes, every account index resolving.

code-less job -> empty register-PO cost-code dropdown + fails #439 health rule one (at least one active cost code) -> next gp_job_sync stamps gp_setup_ok = false -> project quarantined, blocking
import
start a task
PO registration
receiving
shipping

until this ships, every Nexus-created project quarantines at the next sync. accounting adding codes in GP lifts it.

create_job provisions the job's cost codes from GP's own master and division setup. #439's rules stay strict and unchanged.

read-only discovery against TUBC 2026-07-30

- wsiJCJobDetailMSTR is the per-job cost-code write proc (JC00701), EXECUTE + VIEW DEFINITION granted, same validate-then-create convention create_job already uses for wsiJCJobMaster.

wsiJCJobDetailMSTR parameters of note
@I_vWSJobNumber char(17)
@I_vWSCostCodeNumber1..4 char(7)
@I_vCostCodeDescription char(31)
@I_vCostElement smallint
@I_vWSAccountIndex1 int
@I_vWSInactive
@I_vUpdateIfExists
@I_vOnlyValidate
@I_vReturnErrorText
@O_iErrorState / @oErrString OUTPUT

- JC40202 is the company cost-code master, 36 rows of code segments + alias + description + Cost_Element. ACTINDX is 0 on every row - the master does not hold the account mapping. it contains rows nobody should provision blindly:
550-000-5 "VOID: DO NOT USE"
250-000-2 "Test cost code"
370-000-3 "test door"
- UBC's healthy job 23093 carries 46 codes vs this 36-row master. the relay reads the master live at create time, never caches or hardcodes.
- JC40302 maps (Divisions, Cost_Element) -> ACTINDX. VANCOUVER is the only configured division, last modified by mira 2024-09-12:
1/2/3/9 -> 96 (4000-100-00 COGS)
4 -> 105 (4042-100-00 UCA Warden)
5 -> 101 (4010-100-00 UCI-Labour)
6 -> 102 (4020-100-00 outside contractor)
7 -> 138 (5120-100-00 travel)
8 -> 107 (4060-100-00 Freight-In)
- matches production UBC's healthy 23093 account-for-account (46/46 resolve). JC40303/JC40307 hold the same values as 35-slot arrays, JC40302 is the cleanest read.
- account index comes from the JC40302 lookup at create time - never Nexus config, never user input, never hardcoded (same as #427, #430).
- eConnect-only holds, everything through the wsi proc, no direct SQL writes.

relay
- new read op list_cost_code_master joins JC40202 through Cost_Element to JC40302 for the division (default VANCOUVER), feeds the dialog picker. it returns:
cost_code
description
cost_element
account_index
mapped (whether the element has a JC40302 row)
- CreateJobRequest gains cost_codes (selected codes, code + element as returned by the master read). after wsiJCJobMaster succeeds, resolve each code's account from JC40302, then wsiJCJobDetailMSTR per code - OnlyValidate=1 for ALL codes first, then the real calls with UpdateIfExists=1. one transaction, like the /po orchestration.
- an element with no JC40302 row for the division is refused, error naming code and element. do not silently pass index 0 - index 0 is legal (eConnect defaults at posting, #439 passes it) but stamped division accounts are what accounting maintains; index 0 for a code is accounting's call to express in JC40302.
- read back the JC00701 row count, fail loudly on a mismatch.
- verify whether JC00102 carries the job's division or the request must. fall back to the configured default, clear error when a division has no JC40302 rows at all.

backend
- the mutation fronting relay create_job (added with #392's create-project dialog, locate with Auggie) passes cost_codes through.
- expose list_cost_code_master like the other relay reads.
- new root fields go in ROOT_FIELD_POLICY (backend/app/auth_policy.py), test_resolver_gate_completeness fails CI if forgotten.

frontend
- the create-project dialog (#444 extended it, locate with Auggie) gains a cost-code step - master list, mapped codes pre-checked, creator unticks, unmapped codes disabled with the reason.
- after create, the register-PO dropdown shows exactly the selected codes.

one supervised write test in TUBC before building, needs jayp's explicit approval:
1. throwaway job in TUBC
2. one code via hand-run wsiJCJobDetailMSTR (OnlyValidate then real)
3. compare the JC00701 row field-for-field against a UI-created row on 80001-80005, diff JC00102 before/after for rollups
4. companion state the UI writes but the proc does not becomes part of the op design

relay unit tests (fake-conn pattern, tests/test_job_setup_health.py / tests/test_create_job.py)
validate-all-then-write-all ordering
UpdateIfExists on real calls only
account index from the JC40302 read, never the request
refusal for an unmapped element
read-back guard
master-read join shape + mapped flag

backend tests are CI-only (no local Postgres), read the CI conclusion.

simulated user testing (Chrome DevTools MCP on the Railway PR environment; TESTING_ENABLED + TESTING_SIGN_IN_SECRET_HASH set by hand; the NEW relay build installed as the packaged exe, never from-source, with the PR backend URL in extra_backend_urls and the relay restarted; PR channels pin to TUBC)
1. PR frontend URL, testing sign-in
2. Start a Task -> create project, unique job number (NEXUS-<issue>-T1), name
3. cost-code step lists the TUBC master, mapped codes pre-checked, unmapped disabled. uncheck 550-000-5 VOID and the test rows, create
4. creation succeeds. after the next sync stamp the project does NOT carry the GP Setup badge (admin projects grid)
5. register-PO dialog for the new project - dropdown lists exactly the selected codes, none extra, not empty
6. read-only TUBC probe - JC00701 has one row per selected code, WS_Account_Index_1 equal to the JC40302 value for its element (pyodbc, 10.0.0.246,1435, trusted connection, via cd relay; poetry run python)
7. negative - a second project with zero codes selected (if the dialog allows) IS quarantined after sync with the banner

production UBC 23093 is fully healthy (46/46 codes resolve; PO0000070's 4612 was TUBC's stale snapshot, cloned pre-provisioning). UBC's 62 broken jobs (1504 dangling rows) are 2014-2023 legacy with zero PO history ever, all 6352 job-cost PO lines in UBC sit on healthy jobs. #430 and #439 are correct as shipped. #446 filtered the dropdown to account-usable codes with the same GL00105-or-index-0 rule as the create_po guard. full prose version of this plan with the complete probe record is NEXUS_GP_COST_CODE_PROVISIONING_PLAN.md in jayp's Downloads.

Xilous:
closes #448.

Nexus-created job (wsiJCJobMaster) lands with zero JC00701 cost codes -> register-PO cost-code dropdown empty -> #439 health rule one quarantines the project at the next gp_job_sync stamp. relay-created TUBC test jobs in that state:
NEXUS-380-T1/T2
NEXUS-392-T1
NEXUS-444-T1

create flow now provisions the job's cost codes in the same GP transaction as the job. #439 rules unchanged.

every written value comes from GP config, never Nexus or the user (#427/#430 provenance). the request names WHICH codes only. alias, description, profit type, transaction type, GL account index read at create time from:
JC40202 cost-code master
JC40302 (division, cost element) -> account index

relay
- new read op list_cost_code_master - JC40202 joined division-scoped to JC40302 and GL00105
- mapped = JC40302 row exists and its index is 0 or in GL00105, same rule as account_index_exists and #446's dropdown filter. the GL00105 join defends against a stale sandbox where the division mapping itself dangles
- unmapped codes returned, not filtered
- create_job - CreateJobRequest gains cost_codes (code + element only, duplicates rejected)
- after the job lands: refuse cost_code_not_in_master / cost_code_unmapped before any write -> OnlyValidate=1 for ALL codes -> real wsiJCJobDetailMSTR calls with UpdateIfExists=1 -> JC00701 count read-back, fails loudly on a silent no-op
- one transaction, job + codes commit together or neither
- wsiJCJobDetailMSTR call shape (9 params, others defaulted) validated live in TUBC under jayp's supervision 2026-07-30. proc-created row field-for-field identical to a GP-UI-created row, JC00102 carries no rollups at provisioning time (details in #448's comment)
- review fix - cost_codes leaked into the wsiJCJobMaster kwargs via model_dump and tripped create_job's unknown-field guard. now excluded, regression test pins that the job proc never receives the selection

backend
- CreateGpJobInput gains costCodes (CostCodeSelectionInput, costCode + costElement). empty list = unprovisioned job
- create_gp_job passes them through stripped
- new gpCostCodeMaster(company, division) -> GpCostCodeMasterEntry (costCode, description, costElement, mapped). account index not exposed to the client
- ROOT_FIELD_POLICY entry ADMIN_ROLE

frontend
- CreateGpJobDialog gains a Cost codes section, the division's master as a checkbox list, mapped codes pre-checked, unmapped disabled with "No GP account for cost element N in this division"
- selection is derived, state holds only deselections. division change re-defaults, an unmapped code can never enter the payload
- zero selection stays submittable with a warning that the job cannot buy until codes are added in GP
- a failed master read degrades locally (warning inside the section), not folded into the relay-down gate

relay 454 passed (32 new), ruff clean. backend ruff + format clean, suite verified in CI (local run 925 passed, 2 pre-existing environmental failures from missing local rapidfuzz, untouched by this change). frontend 399 passed (3 new), 0 lint errors, build clean.

deployed relays pick up the relay half on their next build/install, same build train as #446.

### DIFF
--- a/backend/app/schemas/project.py
+++ b/backend/app/schemas/project.py
@@ -105,6 +105,11 @@ class ProjectMutations:
         nothing to retry from the still-open dialog. Queuing would buy latency tolerance nobody needs
         and add an ambiguous-write class that does not otherwise exist here.
 
+        The selected cost codes are provisioned in the SAME GP transaction as the job (#448), because
+        the proc alone leaves JC00701 empty - a job with no cost codes has an empty register-PO
+        dropdown and is quarantined by the #425 setup check on the next sync stamp. Splitting them
+        into a second call would make that broken state reachable through a partial failure.
+
         Admin-only. Creating a job writes to the accounting system of record.
         """
 
@@ -133,6 +138,12 @@ class ProjectMutations:
                 input.scheduled_completion_date.isoformat() if input.scheduled_completion_date else None
             ),
             "bid_due_date": input.bid_due_date.isoformat() if input.bid_due_date else None,
+            # Only the code number and its element travel (#448). Everything else the JC00701 row
+            # needs - alias, description, and the GL account index above all - the relay reads out of
+            # GP's own master inside the same transaction, so nothing here can dictate an account.
+            "cost_codes": [
+                {"cost_code": c.cost_code.strip(), "cost_element": c.cost_element} for c in input.cost_codes
+            ],
         }
 
         try:
--- a/backend/app/schemas/project.py
+++ b/backend/app/schemas/project.py
@@ -110,6 +110,11 @@ class ProjectMutations:
         dropdown and is quarantined by the #425 setup check on the next sync stamp. Splitting them
         into a second call would make that broken state reachable through a partial failure.
 
+        `cost_codes_provisioned` carries how many of them GP actually kept, so the dialog can be honest
+        about a selection that did not land. On the adopt path it is zero and means it: an already
+        existing job is left exactly as GP has it, the picked codes are NOT applied to it, and
+        created=false plus the zero count is how the client is told that.
+
         Admin-only. Creating a job writes to the accounting system of record.
         """
 
@@ -157,7 +162,9 @@ class ProjectMutations:
                 # amount of retrying can clear.
                 logger.info("create_gp_job: %s already in GP; adopting instead", input.job_number)
                 project = await _adopt_existing(input.job_number)
-                return CreateGpJobResult(project=project, created=False)
+                # Nothing was provisioned: the create never ran, and this path deliberately does not
+                # go on to write cost codes onto a job somebody else's setup already owns.
+                return CreateGpJobResult(project=project, created=False, cost_codes_provisioned=0)
             # GP said no - a closed fiscal period, an address code that isn't on the customer, a
             # division without accounts. The proc words those better than we could, so the message is
             # passed through to the dialog rather than replaced with a generic failure.
@@ -167,6 +174,11 @@ class ProjectMutations:
         # echoed back (see ops.create_job_op).
         job_number = str((result or {}).get("job_number") or input.job_number).strip()
         job_name = str((result or {}).get("job_name") or input.job_name).strip() or None
+        # The relay's verified read-back of JC00701, not len(payload["cost_codes"]). A relay older than
+        # #448 ignores the unknown cost_codes key entirely and answers without this field, so it reads
+        # as 0 - which is exactly what happened in GP, and how a silently dropped selection surfaces to
+        # the dialog instead of being reported as a provisioned job.
+        cost_codes_provisioned = int((result or {}).get("cost_codes_provisioned") or 0)
 
         def _persist() -> Project:
             with SessionLocal() as session:
@@ -177,14 +189,22 @@ class ProjectMutations:
 
         try:
             # Off the event loop: the /relay-link read loop runs on it and must not block on Postgres.
-            return CreateGpJobResult(project=await asyncio.to_thread(_persist), created=True)
+            return CreateGpJobResult(
+                project=await asyncio.to_thread(_persist),
+                created=True,
+                cost_codes_provisioned=cost_codes_provisioned,
+            )
         except ConflictError:
             # The sync adopted this job between GP committing and us persisting. Benign race, same as
             # the one _persist_missing swallows from the other side - the row we wanted exists.
             # GP still created the job on this call, so this is a real creation - only the Nexus row
             # was written by someone else first.
             logger.info("create_gp_job: %s was adopted by the sync first", job_number)
-            return CreateGpJobResult(project=await asyncio.to_thread(_load_project, job_number), created=True)
+            return CreateGpJobResult(
+                project=await asyncio.to_thread(_load_project, job_number),
+                created=True,
+                cost_codes_provisioned=cost_codes_provisioned,
+            )
         except Exception:
             # The job EXISTS in GP at this point - that call already committed. Losing the Nexus row is
             # recoverable rather than fatal: the sync adopts it on its next pass, and retrying the
--- a/frontend/src/modules/import/CreateGpJobDialog.tsx
+++ b/frontend/src/modules/import/CreateGpJobDialog.tsx
@@ -376,6 +376,10 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
     () => new Set(selectedCostCodeEntries.map(costCodeKey)),
     [selectedCostCodeEntries],
   );
+  // The denominator of the counter. Mapped rows, not every row: an unmapped one can never be checked,
+  // so counting it made an untouched form - which has everything it is allowed to have selected -
+  // read as though the user had already deselected something.
+  const mappedCostCodeCount = useMemo(() => costCodeMaster.filter((c) => c.mapped).length, [costCodeMaster]);
 
   const toggleCostCode = useCallback((key: string) => {
     setDeselectedCostCodes((prev) => {
@@ -415,6 +419,11 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
   ].some((e) => isRelayOpUnsupported(e));
   // The optional section degrades on its own instead: pickers become free-text entry.
   const employeesUnavailable = Boolean(employeesError);
+  // #448: deliberately NOT folded into readsUnsupported. A relay too old for list_cost_code_master
+  // still creates jobs, and a job with no cost codes is a legal create - blocking the form would take
+  // away something #380 allowed. What it changes is the wording: the raw "does not support
+  // list_cost_code_master" says nothing a user can act on, whereas "update the relay" does.
+  const costCodeMasterUnsupported = isRelayOpUnsupported(costCodeMasterError);
 
   const refreshReads = useCallback(() => {
     void refetchCustomers();
@@ -440,15 +449,21 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
     billScopedToOwnCustomer,
   ]);
 
-  const [createGpJob, { loading }] = useMutation<{ createGpJob: { created: boolean; project: Pick<Project, 'id'> } }>(
-    CREATE_GP_JOB,
-    { refetchQueries: [{ query: GET_PROJECTS }] },
-  );
+  const [createGpJob, { loading }] = useMutation<{
+    createGpJob: { created: boolean; costCodesProvisioned: number; project: Pick<Project, 'id'> };
+  }>(CREATE_GP_JOB, { refetchQueries: [{ query: GET_PROJECTS }] });
+
+  // #448: the selection is derived from a master that is not there yet while the read is in flight, so
+  // a submit fired in that window sends costCodes: [] and creates exactly the bare, quarantined job
+  // this feature exists to prevent - and it looks like a normal success. Folded into requiredComplete
+  // rather than into `disabled` so it also stops handleSubmit, which re-checks the same flag.
+  const costCodesPending = division !== '' && costCodeMasterLoading;
 
   const requiredComplete =
     jobNumber.trim() !== '' &&
     jobName.trim() !== '' &&
     division !== '' &&
+    !costCodesPending &&
     customer !== null &&
     addressCodeOrBlank(jobAddressCode) !== '' &&
     addressCodeOrBlank(billtoAddressCode) !== '' &&
@@ -628,12 +643,32 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
       // `=== true`, not `!== false`: an absent or partial payload must not be read as a creation,
       // which is the very claim this branch exists to stop making.
       const created = response.data?.createGpJob?.created === true;
-      showToast(
-        created
-          ? `Job ${jobNumber.trim()} created in GP.`
-          : `Job ${jobNumber.trim()} already existed in GP and is now a project.`,
-        'success',
-      );
+      // #448: and the same rule one level down. GP's read-back count is what says the selection
+      // actually landed - a relay older than #448 drops the codes without a word and still answers a
+      // perfectly ordinary success, so a job that reports "created" can be the bare, quarantined one
+      // this feature exists to prevent. Both silent cases are told rather than dressed as a success:
+      // nobody would otherwise find out until the register-PO dropdown came up empty.
+      const provisioned = response.data?.createGpJob?.costCodesProvisioned ?? 0;
+      const asked = selectedCostCodeEntries.length;
+      const job = jobNumber.trim();
+      if (!created) {
+        // The job is somebody else's setup, left exactly as GP has it - the picked codes were not
+        // added to it, so the selection the user made here has gone nowhere.
+        showToast(
+          asked === 0
+            ? `Job ${job} already existed in GP and is now a project.`
+            : `Job ${job} already existed in GP and is now a project. The cost codes selected here were not applied to it.`,
+          asked === 0 ? 'success' : 'warning',
+        );
+      } else if (asked > 0 && provisioned === 0) {
+        showToast(
+          `Job ${job} was created in GP, but its cost codes were not provisioned - the relay on that workstation is ` +
+            `too old to add them. Add them in GP, and update the relay before creating another job.`,
+          'warning',
+        );
+      } else {
+        showToast(`Job ${job} created in GP.`, 'success');
+      }
       handleClose();
     } catch (err) {
       // GP's own words - a closed fiscal period, an address code not on the customer, a division with
@@ -674,7 +709,9 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
       : costCodeMasterLoading
         ? 'Reading the cost-code master from GP…'
         : costCodeMaster.length === 0 && !costCodeMasterError
-          ? 'GP reports no cost codes for this division.'
+          ? // The master is company-wide - the division only decides which of its codes have a GL
+            // account - so an empty one is not a fact about the division that was picked.
+            "GP's cost-code master is empty for this company, so there is nothing to provision."
           : null;
 
   return (
@@ -880,17 +917,18 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
           <Stack spacing={0.75}>
             <Stack direction="row" spacing={2} alignItems="baseline" justifyContent="space-between">
               <Typography sx={microLabelSx}>Cost codes</Typography>
-              {costCodeMaster.length > 0 && (
+              {mappedCostCodeCount > 0 && (
                 <Typography variant="caption" color="text.secondary" sx={tabularSx}>
-                  {`${selectedCostCodeEntries.length} of ${costCodeMaster.length} selected`}
+                  {`${selectedCostCodeEntries.length} of ${mappedCostCodeCount} selected`}
                 </Typography>
               )}
             </Stack>
 
             {costCodeMasterError && (
               <Alert severity="warning">
-                Could not read the cost codes from GP. The job can still be created, but it will have none until they
-                are added in GP. {costCodeMasterError.message}
+                {costCodeMasterUnsupported
+                  ? 'The connected relay is too old to offer cost codes. Update the relay on that workstation to pick them here. The job can still be created, but it will have none until they are added in GP.'
+                  : `Could not read the cost codes from GP. The job can still be created, but it will have none until they are added in GP. ${costCodeMasterError.message}`}
               </Alert>
             )}
 
--- a/frontend/src/modules/import/CreateGpJobDialog.tsx
+++ b/frontend/src/modules/import/CreateGpJobDialog.tsx
@@ -6,7 +6,10 @@ import {
   DialogActions,
   TextField,
   Autocomplete,
+  Box,
   Button,
+  Checkbox,
+  FormControlLabel,
   Stack,
   Alert,
   MenuItem,
@@ -19,6 +22,7 @@ import { RefreshCw } from 'lucide-react';
 import { useApolloClient, useMutation, useQuery } from '@apollo/client/react';
 import {
   CREATE_GP_JOB,
+  GET_GP_COST_CODE_MASTER,
   GET_GP_CUSTOMERS,
   GET_GP_CUSTOMER_ADDRESSES,
   GET_GP_DIVISIONS,
@@ -32,7 +36,7 @@ import { extractGpError, isRelayOpUnsupported, type GpError } from '../../graphq
 import RelayStatusChip from '../../relay/RelayStatusChip';
 import { useRelayStatus } from '../../relay/useRelayStatus';
 import type { Project } from '../../types/project';
-import { monoSx, microLabelSx } from '../../theme';
+import { monoSx, microLabelSx, tabularSx } from '../../theme';
 import AddCustomerAddressDialog, { type CreatedGpCustomerAddress } from './AddCustomerAddressDialog';
 
 interface GpCustomerOption {
@@ -56,6 +60,25 @@ interface GpEmployeeOption {
   employeeId: string;
   firstName: string | null;
   lastName: string | null;
+}
+
+/**
+ * #448: one row of the company's cost-code master, as it applies to the chosen division.
+ *
+ * `mapped` is not a property of the code on its own - it says the division has a GL account for this
+ * code's cost element. An unmapped code provisioned onto the job is exactly what the GP-setup health
+ * check quarantines the project for, so those rows are offered as facts and never as choices.
+ */
+interface GpCostCodeMasterEntry {
+  costCode: string; // two-segment number 'cc1-cc2', e.g. '310-000'
+  description: string | null;
+  costElement: number; // GP Cost_Element; the trailing digit of the register-PO cost code
+  mapped: boolean;
+}
+
+/** GP's own 'phase-step-element' spelling, and the row identity the selection is keyed on. */
+function costCodeKey(c: GpCostCodeMasterEntry): string {
+  return `${c.costCode}-${c.costElement}`;
 }
 
 function employeeLabel(e: GpEmployeeOption): string {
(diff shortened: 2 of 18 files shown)


frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -530,6 +609,9 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
             billtoAddressCode: addressCodeOrBlank(billtoAddressCode),
             taxScheduleId,
             createdDate,
+            // #448: what the job is provisioned with. An empty list is a legal create - it just makes
+            // a job that cannot take a PO until accounting adds cost codes in GP.
+            costCodes: selectedCostCodeEntries.map((c) => ({ costCode: c.costCode, costElement: c.costElement })),
```
Xilous, line 629: Create Job button is never gated on the cost-code master read being in flight. `requiredComplete` (line 448) and `disabled={disabled || !requiredComplete}` (line 1083) don't see `costCodeMasterLoading`.

fill every other field first, pick Division last -> button enables on the same tick gpCostCodeMaster fires -> `costCodeMaster` is `[]` while in flight -> `selectedCostCodeEntries` is `[]` -> click sends `costCodes: []` -> relay skips `_provision_cost_codes` (`if request.cost_codes:`) -> job lands in GP with zero JC00701 rows, toast reports success - the exact unprovisioned, quarantined job this PR exists to prevent.

no warning fires in that window - the "With no cost codes the job is blocked..." caption is gated on `costCodeMaster.length > 0`. window reopens on every division change. test suite works around the race - `fillRequired()` awaits `screen.findByText('2 of 3 selected')`.

gate the submit (or the button) on the master read having settled whenever a division is picked.

relay/src/ucnexus_relay/econnect.py
```diff
@@ -1106,6 +1106,164 @@ def create_job(conn, *, only_validate: bool = False, **fields) -> None:
...
+    when the element has no JC40302 row at all) and mapped. Unmapped codes are returned rather than
+    filtered out, so the picker can show them greyed with a reason instead of silently offering a
+    shorter list than GP does."""
+    rows = conn.cursor().execute(
+        "SELECT RTRIM(m.Cost_Code_Number_1) AS cc1, RTRIM(m.Cost_Code_Number_2) AS cc2, "
+        "RTRIM(m.Cost_Code_Alias) AS alias, RTRIM(m.Cost_Code_Description) AS descr, "
+        "m.Cost_Element AS elem, m.Profit_Type_Number AS ptype, m.Type_of_Transaction AS ttype, "
+        "a.ACTINDX AS account_index, g.ACTINDX AS resolved "
+        "FROM dbo.JC40202 m "
+        "LEFT JOIN dbo.JC40302 a ON RTRIM(a.Divisions) = ? AND a.Cost_Element = m.Cost_Element "
```
Xilous, line 1182: nothing enforces one master row per (cost_code, cost_element). duplicates arrive two ways:
- JC40302 LEFT JOIN fans out if that table ever holds more than one row per (Divisions, Cost_Element)
- SELECT drops Cost_Code_Number_3/_4, so two JC40202 rows differing only in those segments collapse to the same identity

either way `list_cost_code_master` emits two dicts with the same ("cost_code", "cost_element") key, and three downstream layers assume uniqueness:
- picker renders duplicate React keys from `costCodeKey`, one deselection entry drives both checkboxes
- `master = {(row["cost_code"], row["cost_element"]): row ...}` (ops.py:266) keeps the last row, so the provisioned GL account depends on SQL row order
- both duplicates ride the payload, `CreateJobRequest.normalize` rejects with "selected more than once" - raw pydantic invalid_payload dump on every create for that division, no way to deselect just one from the UI

`create_job_cost_code` also never sends @I_vWSCostCodeNumber3/4. assert/dedupe the cardinality here, or carry cc3/cc4 through the identity.

relay/src/ucnexus_relay/ops.py
```diff
@@ -235,6 +235,96 @@ def create_po_op(conn, *, company: str, request: models.CreatePoRequest) -> mode
...
+                cost_element=selection.cost_element,
+                division=request.division,
+            )
+        chosen.append(row)
+
+    for only_validate in (True, False):
+        for row in chosen:
+            # split_cost_code is what the PO path splits with, so the segments provisioned here and the
+            # segments a PO later looks the code up by cannot drift.
+            cc1, cc2, _cc3, _cc4, _element = econnect.split_cost_code(row["cost_code"])
```
Xilous: write path re-derives cc1/cc2 by re-splitting the joined string. `list_cost_code_master` throws the segments away into `f"{r.cc1}-{r.cc2}"` (econnect.py:1169); this line recovers them with `split_cost_code`, a three-segment 'phase-step-element' parser (`split("-")`, segs[0]/segs[1]).

a master segment containing '-' round-trips to a different code. cc1='210', cc2='200-A' -> '210-200-A' -> splits back to 210/200 -> `create_job_cost_code` writes a JC00701 row for a code the company master has never held, carrying 200-A's alias and account index. the COUNT read-back still sees exactly one row, nothing detects it.

carry `cost_code_number_1`/`cost_code_number_2` through the dict. the inline comment's claim the split "cannot drift" from the PO path doesn't hold - the PO path splits three-segment strings, this splits two-segment ones.

relay/src/ucnexus_relay/ops.py
```diff
@@ -235,6 +235,96 @@ def create_po_op(conn, *, company: str, request: models.CreatePoRequest) -> mode
...
+                f"account for division '{request.division}' in {company}: JC40302 maps no account index "
+                f"for that cost element, or the one it maps does not exist in the chart (GL00105). A job "
+                f"provisioned with it could register POs and never receive them.",
+                cost_code=selection.cost_code,
+                cost_element=selection.cost_element,
+                division=request.division,
+            )
+        chosen.append(row)
+
+    for only_validate in (True, False):
```
Xilous, line 327: validate-all-then-write-all loop issues 2N wsiJCJobDetailMSTR round-trips inside the single create_job relay call, against the fixed default timeout (`DEFAULT_TIMEOUT_SECONDS = 30.0`, relay_gateway.py:36; no override at project.py:150). picker defaults to select-all of the division master, so N is the whole master.

2N EXECs plus the master read plus the job proc's own two passes push past 30s -> backend raises RelayTimeoutError while the relay keeps going and commits -> dialog reports failure for a job that exists -> retry lands on the job_already_exists branch, adopting with created=false.

the dry-run pass buys nothing the transaction does not already guarantee - `_run_create_job` (channel.py:316-323) rolls back everything on a failure at code k, and @I_vUpdateIfExists=1 makes retries idempotent. dropping the validate pass halves the window; bounding it (or raising the timeout for this op) closes it.

relay/src/ucnexus_relay/ops.py
```diff
@@ -235,6 +235,96 @@ def create_po_op(conn, *, company: str, request: models.CreatePoRequest) -> mode
...
+                cost_code_number_2=cc2,
+                alias=row["alias"] or "",
+                description=row["description"] or "",
+                cost_element=row["cost_element"],
+                account_index=row["account_index"],
+                profit_type_number=row["profit_type_number"],
+                type_of_transaction=row["type_of_transaction"],
+            )
+
+    landed = conn.cursor().execute(
```
Xilous: read-back counts every active JC00701 row on the job and demands exact equality with `len(chosen)`, with no baseline count before the writes. if wsiJCJobMaster, a WennSoft trigger, or template seeding ever writes a row, every create in that company fails permanently ("reported success but job X has N active cost codes, expected M") and rolls back a job GP built correctly. any of these survives that instead:
- before/after delta
- `>=`
- check the specific keys written (econnect already has the per-row mechanism, `cost_code_on_job`)

also the only raw GP-table SQL in ops.py, and it hardcodes 'WS_Inactive = 0' instead of econnect's `_ACTIVE_COST_CODE` constant.

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -780,6 +874,90 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
...
+              {costCodeMaster.length > 0 && (
+                <Typography variant="caption" color="text.secondary" sx={tabularSx}>
+                  {`${selectedCostCodeEntries.length} of ${costCodeMaster.length} selected`}
+                </Typography>
+              )}
+            </Stack>
+
+            {costCodeMasterError && (
+              <Alert severity="warning">
+                Could not read the cost codes from GP. The job can still be created, but it will have none until they
```
Xilous: `costCodeMasterError` is excluded from both `readError` and `readsUnsupported` (lines 407-415), so `isRelayOpUnsupported` is never consulted for it. in the deploy-before-relay-update window (guaranteed for a brand-new op) the relay answers unknown_op and this Alert renders the raw relay string via `{costCodeMasterError.message}` instead of the "relay too old, update the relay on that workstation" wording every sibling read gets. Create button stays enabled and sends `costCodes: []` -> every job created in that window is unprovisioned and quarantined on the next sync while the toast says created.

branch on `isRelayOpUnsupported(costCodeMasterError)` for the message; consider whether the warning-severity Alert is enough signal.

relay/src/ucnexus_relay/ops.py
```diff
@@ -272,10 +369,15 @@ def create_job_op(conn, *, company: str, request: models.CreateJobRequest) -> mo
             proc="wsiJCJobMaster",
         )
 
+    provisioned = 0
+    if request.cost_codes:
```
Xilous: `cost_code_not_in_master` and `cost_code_unmapped` fire only after `econnect.create_job(only_validate=False)` has written the job and `get_job` has confirmed it. both are answered by a read-only master SELECT that could run before the dry run. a stale picker selection rejects the create after the JC00102 row exists -> channel's rollback has to undo a completed job create, contradicting step 2's own docstring ("Running it first means a rejected create fails having written nothing") and leaving anything non-transactional the WennSoft proc did unrecovered. move the master read and the two refusal checks ahead of the job proc.

relay/src/ucnexus_relay/econnect.py
```diff
@@ -1106,6 +1106,164 @@ def create_job(conn, *, only_validate: bool = False, **fields) -> None:
...
+    out = []
+    for r in rows:
+        account_index = None if r.account_index is None else int(r.account_index)
+        out.append(
+            {
+                "cost_code": f"{r.cc1}-{r.cc2}",
+                "alias": r.alias or None,
+                "description": r.descr or None,
+                "cost_element": int(r.elem),
+                "profit_type_number": int(r.ptype),
```
Xilous: `int(r.ptype)` and `int(r.ttype)` are the only columns in this loop with no NULL guard; every neighbour has one (`account_index = None if r.account_index is None else int(...)`, `r.alias or None`, `r.descr or None`). one JC40202 row with NULL Profit_Type_Number or Type_of_Transaction raises TypeError -> none of `_dispatch`'s typed handlers catch it -> last-resort internal_error -> picker fails for the entire company with a Python type error in the Alert, and the same read inside `_provision_cost_codes` aborts a create mid-transaction.

backend/app/schemas/project.py
```diff
@@ -133,6 +138,12 @@ async def create_gp_job(self, info: strawberry.Info, input: CreateGpJobInput) ->
                 input.scheduled_completion_date.isoformat() if input.scheduled_completion_date else None
             ),
             "bid_due_date": input.bid_due_date.isoformat() if input.bid_due_date else None,
+            # Only the code number and its element travel (#448). Everything else the JC00701 row
+            # needs - alias, description, and the GL account index above all - the relay reads out of
+            # GP's own master inside the same transaction, so nothing here can dictate an account.
+            "cost_codes": [
```
Xilous, line 149: job_already_exists adopt fallback silently discards the cost-code selection. likeliest existing job here is the sync-adopted one created directly in GP with no cost codes, and this dialog is the only screen in Nexus that offers cost codes. admin ticks the division master -> submits -> created=false, project still quarantined, register-PO dropdown empty, nothing in the reply or toast says the selection was dropped. even if provisioning onto an adopted job stays out of scope, the response should say the codes were not applied.

relay/src/ucnexus_relay/models.py
```diff
@@ -364,13 +400,25 @@ def normalize(self):
...
         return self
 
 
 class CreateJobResponse(BaseModel):
     job_number: str
     job_name: str
     company: str
+    # Issue #448: how many JC00701 cost codes were written onto the new job, verified by a read-back
+    # rather than counted off the request. 0 for a caller that sent none.
+    cost_codes_provisioned: int = 0
```
Xilous, line 422: `cost_codes_provisioned` is returned on every reply and consumed by nothing - backend reads only job_number and job_name (project.py:168), CreateGpJobResult has no field for it. meanwhile a pre-#448 relay accepts create_job, silently drops the unknown `cost_codes` key (CreateJobRequest declares no extra='forbid', pydantic's default is ignore), creates a bare job, and the dialog toasts success; the op-set handshake covers only the new read op, not the new field on the existing create_job op. have the backend compare `result.get("cost_codes_provisioned")` against the count it sent - that catches exactly this. otherwise the field is dead weight pinned into the wire shape by test_channel.py.

relay/src/ucnexus_relay/econnect.py
```diff
@@ -1106,6 +1106,164 @@ def create_job(conn, *, only_validate: bool = False, **fields) -> None:
...
+    use), alias, description, cost_element, profit_type_number, type_of_transaction, account_index (None
+    when the element has no JC40302 row at all) and mapped. Unmapped codes are returned rather than
+    filtered out, so the picker can show them greyed with a reason instead of silently offering a
+    shorter list than GP does."""
+    rows = conn.cursor().execute(
+        "SELECT RTRIM(m.Cost_Code_Number_1) AS cc1, RTRIM(m.Cost_Code_Number_2) AS cc2, "
+        "RTRIM(m.Cost_Code_Alias) AS alias, RTRIM(m.Cost_Code_Description) AS descr, "
+        "m.Cost_Element AS elem, m.Profit_Type_Number AS ptype, m.Type_of_Transaction AS ttype, "
+        "a.ACTINDX AS account_index, g.ACTINDX AS resolved "
+        "FROM dbo.JC40202 m "
```
Xilous, line 1181: no inactive/retired filter on JC40202, while every sibling cost-code read filters on WS_Inactive (list_cost_codes line 722, job_setup_health via _ACTIVE_COST_CODE). if JC40202 carries such a flag, a retired code comes back mapped, arrives pre-checked (select-all default), and is written as an ACTIVE JC00701 row - codes GP's own Job Cost Code Maintenance window would no longer offer. one live probe of JC40202's columns settles it: add the filter, or record in the docstring that no such flag exists.

relay/src/ucnexus_relay/econnect.py
```diff
@@ -1106,6 +1106,164 @@ def create_job(conn, *, only_validate: bool = False, **fields) -> None:
...
+                "cost_code": f"{r.cc1}-{r.cc2}",
+                "alias": r.alias or None,
+                "description": r.descr or None,
+                "cost_element": int(r.elem),
+                "profit_type_number": int(r.ptype),
+                "type_of_transaction": int(r.ttype),
+                "account_index": account_index,
+                # No JC40302 row -> nothing to provision with. A row whose index is non-zero but absent
+                # from GL00105 is the stale-sandbox case above and is just as unusable.
+                "mapped": account_index is not None and (account_index == 0 or r.resolved is not None),
```
Xilous, line 1222: `mapped` is the fourth independent spelling of the "index 0, or present in GL00105" rule, tied together only by docstrings:
- account_index_exists (Python, line 853)
- list_cost_codes (SQL, line 723)
- job_setup_health (negated SQL, lines 786/796)
- this expression

tighten the rule for the create_po guard and the picker keeps offering codes the guard refuses. `_ACTIVE_COST_CODE` one page up is the in-file precedent - a shared predicate (SQL fragment plus the existing Python helper) consumed by all four sites.

relay/src/ucnexus_relay/econnect.py
```diff
@@ -1106,6 +1106,164 @@ def create_job(conn, *, only_validate: bool = False, **fields) -> None:
...
+    cost_element: int,
+    account_index: int,
+    profit_type_number: int,
+    type_of_transaction: int,
+) -> None:
+    """Put ONE cost code on a GP job through the WennSoft cost-code proc wsiJCJobDetailMSTR - the
+    JC00701 write, and the missing half of create_job above (see this section's header for why a job
+    with no cost codes is a job nobody can raise a PO against).
+
+    Nine inputs are sent; the proc's remaining ones are left defaulted, the same treatment create_job
```
Xilous: docstring claims create_job's treatment ("an unset optional is absent from the EXEC, not an explicit NULL"), but this EXEC is a fixed statement that always binds all twelve parameters, and the caller substitutes '' for a missing alias/description (`alias=row["alias"] or ""`, ops.py:306). a blank master alias is written as an explicit blank. with @I_vUpdateIfExists=1, a retry after a lost reply overwrites a hand-set alias/description on the existing row with ''. either build the EXEC dynamically like create_job does, or correct the docstring and accept the overwrite semantics deliberately.

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -583,6 +666,17 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
...
 
+  // #448: why there is no list to tick, when there isn't one. A failed read is handled separately -
+  // it is the one case where the section has to say the job will go out unprovisioned.
+  const costCodesNote =
+    division === ''
+      ? 'Pick a division first - whether a code has a GP account depends on it.'
+      : costCodeMasterLoading
+        ? 'Reading the cost-code master from GP…'
+        : costCodeMaster.length === 0 && !costCodeMasterError
+          ? 'GP reports no cost codes for this division.'
```
Xilous: "GP reports no cost codes for this division." is not what an empty list means. the master read has no WHERE clause on JC40202 - the division only drives the LEFT-JOINed account - so the row count is identical for every division; empty means the company has no cost-code master at all. the actual division-specific empty state (no usable account for any element) renders as a full list of disabled checkboxes and never reaches this branch.

frontend/src/modules/import/CreateGpJobDialog.tsx
```diff
@@ -780,6 +874,90 @@ export default function CreateGpJobDialog({ open, onClose }: CreateGpJobDialogPr
...
 
+          {/* #448: a job created with no cost codes has an empty register-PO dropdown and is
+              quarantined by the GP-setup check on the next sync, so they are provisioned here rather
+              than left for accounting to notice. */}
+          <Stack spacing={0.75}>
+            <Stack direction="row" spacing={2} alignItems="baseline" justifyContent="space-between">
+              <Typography sx={microLabelSx}>Cost codes</Typography>
+              {costCodeMaster.length > 0 && (
+                <Typography variant="caption" color="text.secondary" sx={tabularSx}>
+                  {`${selectedCostCodeEntries.length} of ${costCodeMaster.length} selected`}
```
Xilous: counter's denominator includes unmapped rows that can never be checked, so an untouched form already reads as partially deselected - this PR's own test fixture shows "2 of 3 selected" on a fresh form, indistinguishable from someone having unchecked one. on a division with several unmapped elements the counter can never reach parity. use the mapped count as the denominator and note the unmapped separately.

Xilous:
all 15 review findings addressed in 47fb08a, except two answered below.

1 (submit while master read in flight) - fixed. costCodesPending = division chosen + read loading, folded into requiredComplete, blocks the button and handleSubmit's own guard.

2 (duplicate (code, element) identity / cc3-cc4 dropped) - hardened. master read filters RTRIM(cc3) = '' AND RTRIM(cc4) = '' in SQL and dedupes on (cost_code, cost_element) keeping the first row. live probe 2026-07-30 found zero non-blank cc3/cc4 rows and zero duplicate pairs in JC40202 and JC40302.

3 (re-splitting 'cc1-cc2' loses fidelity) - fixed. master read returns the raw segments (cost_code_number_1/_2), write binds them directly, split_cost_code re-split deleted. test fixture whose second segment contains '-' pins that the segments win.

4 (2N round-trips vs the 30s timeout; dry-run duplicates the rollback) - answered, not changed. full 36-row master is 72 EXECs at LAN millisecond latency, roughly 2s worst case - an order of magnitude inside the timeout. validate pass stays. retry's adopt path now warns (finding 9).

5 (COUNT read-back has no baseline) - fixed. replaced with per-code verification through cost_code_on_job, the exact lookup /po runs.

6 (unknown_op renders raw) - fixed. costCodeMasterError goes through isRelayOpUnsupported, section says to update the relay. still not folded into the hard relay gate.

7 (refusals after the job's real write) - fixed. selection resolved and refused (cost_code_not_in_master / cost_code_unmapped) BEFORE the wsiJCJobMaster dry run. refusal tests assert create_job was never called.

8 (int(None) kills the read) - fixed. NULL Profit_Type_Number / Type_of_Transaction degrade to 0 (proc default). live probe found zero NULL rows.

9 (adopt path silently drops the selection) - fixed. created=false with a non-empty selection shows a warning toast saying the selected codes were not applied to the existing job. backend returns cost_codes_provisioned 0 on that path and documents it.

10 (cost_codes_provisioned consumed by nothing) - fixed. CREATE_GP_JOB selects costCodesProvisioned. created=true with a selection and 0 provisioned shows a warning naming the too-old relay. toast does not say "retry" - a retry hits the job_already_exists adopt path, which provisions nothing by design.

11 (no inactive filter on JC40202) - probed live 2026-07-30. JC40202 has no inactive or retired flag at all (full column list checked). docstring records the absence is deliberate.

12 (fourth spelling of the usable-account rule) - deferred with a cross-reference comment next to _ACTIVE_COST_CODE naming all four sites.

13 (docstring claims the create_job omitted-optional treatment) - fixed. docstring now says all nine inputs are always bound and blanks arrive as ''. the proc's remaining ~120 inputs stay defaulted. UpdateIfExists clobber unreachable through this op, job_exists refuses an existing job, docstring says that too.

14 (empty-master copy blames the division) - fixed. now says the company's cost-code master is empty.

15 (counter denominator includes uncheckable rows) - fixed. denominator is the mapped count, counter hides when nothing is mapped.

relay 459 passed + ruff clean
backend 927 passed with the two pre-existing rapidfuzz-environmental failures + ruff/format clean
frontend 403 passed + 0 lint errors + clean build

PR #450, opened 2026-07-31
feat(warehouse,shipping): GP receipt numbers, receiving history, delivery request lifecycle

### LINKED ISSUE #447: URGENT FEATURES
Receive entities displayed against each PO

Each "receive" should have a "Receipt Number"

Flesh out shipping out process

some form of display that lists all the POs, expandable and shows all the "receives" against it

---

Flesh out shipping out

shipping out document that is generated: 

[20260730163232085.pdf](https://github.com/user-attachments/files/30566397/20260730163232085.pdf)

Xilous:
closes #447.

receiving

every receive stores the GP receipt number (RCT######) and batch number returned by the relay's create_receipt call. both were discarded before, so a Nexus receive and its GP receipt shared no identifier. columns nullable, pre-change rows show a dash. live path and outbox drain path share one persist function. receipt number shown in the receive success dialog and the Recent Activity grid.

new Receiving History view on the warehouse Receiving page (Receive/History toggle). lists every GP-registered PO including completed ones, with text search and a project filter. each row shows:
received/ordered totals
receive count
last-received date

list built from grouped aggregates; per-PO receives load on first expand via the existing poReceivingDetails query. rows expand to show each receive:
receipt number
date
receiver
line quantities

shipping

packing slips become Delivery Requests. packing-slip PDF deleted; the generated document is the Delivery Request form, replicating the paper layout. confirm-shipment dialog captures the whole paper form:
pickup date
delivery date
shipper email
shipper phone
pickup location (snapshotted from the primary warehouse address)
carrier/tag/BOL
weight
the eight deliver-to questionnaire answers
contractor contact
UCSH contact
sales order number

lifecycle SCHEDULED -> PICKED_UP -> DELIVERED. actor and timestamp stamped on each transition, strict one-way ladder, row-locked transitions. confirming the shipment is still the moment items are claimed (opening items go SHIPPED_OUT); the states only document the truck's journey, no inventory moves between them.

header editable only while SCHEDULED, full-replace semantics (a blank clears the field). slip number, project, items, shipped-by/at never editable.

shipments list reworked into expandable rows. each row shows:
status chip
dates
carrier

expansion holds:
items
Delivery Request reprint (regenerated from stored fields)
Edit
Mark Picked Up
Mark Delivered
Return

returns unchanged.

migration 076 adds receive receipt numbers. migration 077 adds delivery request + lifecycle; existing slips backfill as DELIVERED.


backend/app/repositories/warehouse/receiving.py
```diff
@@ -591,6 +602,94 @@ def get_back_ordered_items(session: Session, project_id: uuid.UUID | None = None
...
+                    POStatus.GP_REGISTERED,
+                    POStatus.VENDOR_CONFIRMED,
+                    POStatus.PARTIALLY_RECEIVED,
+                    POStatus.CLOSED,
+                ]
+            ),
+        )
+        # Most recently received first, then the never-received POs - which are the ones somebody is
+        # about to receive, so they belong at the top of what is left rather than buried.
+        .order_by(receive_totals.c.last_received_at.desc().nulls_first(), POModel.po_number.desc())
```
Xilous: the docstring two lines up says most recently received first, then the never-received POs. `.desc().nulls_first()` sorts the NULL last_received_at rows to the top: every never-received PO stacks above the entire receipt history, so the receive someone just booked is buried below all of them. no test pins the ordering. `nulls_last()` appears to be what was meant.

frontend/src/modules/shipping/deliveryRequest.ts
```diff
@@ -0,0 +1,291 @@
...
+    );
+  }
+  return lines;
+}
+
+/** The same block built from a stored shipment's items, for a Delivery Request reprinted later. */
+export function slipMaterialLines(items: PackingSlipItem[]): string[] {
+  const openingItems = items
+    .filter((i) => i.itemType === 'OPENING_ITEM')
+    .map((i) => ({ openingNumber: i.openingNumber ?? '', leaf: i.leaf ?? null }));
```
Xilous, line 265: slipMaterialLines maps opening items down to openingNumber and leaf, so a reprint from the Shipments list loses the building/floor/location suffix the confirm-time print carried. buildMaterialLines reads those off the cart items; PackingSlipItem stores no placement. ship 0019-EX Leaf 1 sitting in building A / floor 1 / Rm 101:
- success-screen PDF prints "(1) Unit of Opening 0019-EX Leaf 1 - A / 1 / Rm 101"
- reprint prints "(1) Unit of Opening 0019-EX Leaf 1"
separately, the success-screen PDF prints sentDetailsRef (the untrimmed client copy) and reprints print the stored fields, so the two documents can differ where the backend trimmed whitespace. printing valuesFromSlip(result) and slipMaterialLines(result.items) makes both paths read from one source.

frontend/src/modules/shipping/DeliveryRequestForm.tsx
```diff
@@ -67,34 +77,92 @@ function describeConfirmError(err: unknown): string {
...
   // this dialog kept offering the leaf that just left (#337).
   //
   // Not awaited. `awaitRefetchQueries` folds the refetches into the mutate promise
   // (QueryInfo.markMutationResult -> Promise.all), so one slow or failed refetch would reject a
-  // shipment that already committed: no success view, no packing-slip PDF, and a retry that can only
-  // fail on the duplicate slip number. The shipment is the thing that must be reported truthfully;
-  // the grid behind the modal can catch up on its own.
+  // shipment that already committed: no success view, no Delivery Request PDF, and a retry that can
+  // only fail on the duplicate slip number. The shipment is the thing that must be reported
+  // truthfully; the grid behind the modal can catch up on its own.
```
Xilous, line 168: nothing here touches packingSlips. SHIPPING_STALE_ROOT_FIELDS is
shipReadyItems
openingItems
SHIPPING_REFETCH_QUERIES is GetOpeningLeafStatus. the cart trigger and ShippingCart render outside <Routes> in shipping/index.tsx. user on /app/shipping/returns with ShipmentsList mounted and its cache-and-network read settled opens the cart and confirms -> the new SCHEDULED slip never appears in the mounted list until a route remount, which is exactly the window in which it is editable. the refetch.ts comment ("its next mount already goes to the network") covers the warehouse ShipmentsPage case, not this one. add packingSlips to the evicted root fields.

frontend/src/modules/warehouse/ReceiveModal.tsx
```diff
@@ -537,6 +564,7 @@ export default function ReceiveModal({ open, onClose, poIds }: ReceiveModalProps
 
     setReceivedCount(totalItemsToReceive);
     setQueuedCount(queued.length);
+    setPostedReceipts(receipts);
```
Xilous: receipts is pushed per committed PO inside the loop, but this setPostedReceipts sits below the `if (failureMessage)` early return, and the loop breaks on the first failure. receive PO-A and PO-B together, A commits and GP returns its receipt number, B fails -> number collected, never rendered. a mixed live+queued batch (relay drops between the two calls) loses it differently. setPostedReceipts runs, but the receipt block renders only under `succeeded && queuedCount === 0` and the amber queued alert names no numbers. mitigated by the new GP Receipt column in Recent Activity showing the number once the modal closes.

frontend/src/modules/shipping/deliveryRequest.ts
```diff
@@ -0,0 +1,291 @@
...
+  openingItems: MaterialOpeningItem[],
+  looseItems: MaterialLooseItem[],
+): string[] {
+  const lines = openingItems.map((item) => {
+    const leaf = item.leaf != null ? ` Leaf ${item.leaf}` : '';
+    const where = [item.building, item.floor, item.location].filter(Boolean).join(' / ');
+    return `(1) Unit of Opening ${item.openingNumber}${leaf}${where ? ` - ${where}` : ''}`;
+  });
+  for (const item of looseItems) {
+    lines.push(
```
Xilous, line 254: buildMaterialLines takes MaterialLooseItem.openingNumber and never reads it. loose lines print as "(4) Units of AD8406 - Locksets" with nothing naming the opening. the deleted PackingSlipDocument printed an Opening # column on every loose row. two loose lines of the same product for different openings are indistinguishable on the printed form. deliveryRequest tests all use a single opening, so the loss is unpinned.

frontend/src/modules/shipping/ShipmentsList.tsx
```diff
@@ -1,164 +1,222 @@
...
+  const [shown, setShown] = useState(PAGE);
+  const [expanded, setExpanded] = useState<Set<string>>(() => new Set());
   const [activeSlip, setActiveSlip] = useState<ReturnSlip | null>(null);
+  const [editing, setEditing] = useState<PackingSlip | null>(null);
+  const [lifecycle, setLifecycle] = useState<{ slip: PackingSlip; action: LifecycleAction } | null>(
+    null,
+  );
+  const [generatingFor, setGeneratingFor] = useState<string | null>(null);
 
   const { data, loading, refetch } = useQuery<{ packingSlips: PackingSlip[] }>(GET_PACKING_SLIPS, {
```
Xilous: `error` is never destructured and the render branches only on `loading && !data`. a failed packingSlips query falls through to the table body and prints "No shipments match this search." a backend error reads as an empty result with nothing on screen to retry from.

frontend/src/modules/warehouse/ReceivingHistory.tsx
```diff
@@ -0,0 +1,384 @@
...
+ * the moment it is complete. Reconciling a delivery against GP - "which receipt was this, and who
+ * booked it" - needs the finished ones, so this is the one surface where CLOSED POs are in scope.
+ */
+export default function ReceivingHistory({ projects, projectMap }: ReceivingHistoryProps) {
+  const [expandedIds, setExpandedIds] = useState<Set<string>>(new Set());
+  const [search, setSearch] = useState('');
+  const [projectFilter, setProjectFilter] = useState('');
+
+  // cache-and-network, because this view is unmounted while the user is receiving. A receive that
+  // just posted has to show up the moment they switch over here, and RECEIVE_REFETCH_QUERIES does
```
Xilous, line 275: this covers the live receive path (view unmounted) but not the outbox drain path (view can be mounted). GP_OUTBOX_DRAINED_STALE_ROOT_FIELDS evicts poReceivingDetails but not receivingHistoryPos. a queued receipt drains while the user sits on History -> the expanded row's receives repair (new receipt number appears) while the row above keeps its old
receive count
totals
last-received timestamp
until the user navigates away and back. add receivingHistoryPos to that eviction list.

backend/alembic/versions/077_shipment_delivery_request.py
```diff
@@ -0,0 +1,106 @@
...
+def upgrade() -> None:
+    shipment_status = postgresql.ENUM("SCHEDULED", "PICKED_UP", "DELIVERED", name=_STATUS_ENUM)
+    shipment_status.create(op.get_bind(), checkfirst=True)
+    op.add_column(
+        "packing_slips",
+        sa.Column(
+            "status",
+            postgresql.ENUM("SCHEDULED", "PICKED_UP", "DELIVERED", name=_STATUS_ENUM, create_type=False),
+            nullable=False,
+            server_default="DELIVERED",
```
Xilous, line 98: two consequences of the DELIVERED backfill need an explicit decision.
- Railway production carries real shipments now. a slip confirmed before the deploy whose truck has not arrived lands DELIVERED, the ladder is one-way and edits are refused past SCHEDULED, so nothing in the product can ever correct it. any future "was this actually delivered" consumer needs the unwritten rule status = DELIVERED AND delivered_at IS NOT NULL.
- the server-default safety net (an insert that omits status degrading to DELIVERED instead of failing) does not hold through the ORM. the model declares neither a default nor a server_default, so SQLAlchemy renders the unset column as an explicit NULL. verified on SQLAlchemy 2.0: INSERT INTO ps (id, status) VALUES (?, ?) with None, NOT NULL violation. every construction site in this PR names status, so nothing breaks today; the safety net the migration docstring describes is unreachable from Python.

frontend/src/modules/shipping/DeliveryRequestDocument.tsx
```diff
@@ -0,0 +1,341 @@
...
+export default function DeliveryRequestDocument({
+  packingSlipNumber,
+  projectName,
+  jobNumber,
+  date,
+  shipper,
+  materialLines,
+  values,
+}: DeliveryRequestDocumentProps) {
+  const divisionLines = (values.pickupLocation ?? '').split('\n').filter((l) => l.trim() !== '');
```
Xilous: the Division Address box is printed from values.pickupLocation, the same field the PICKUP LOCATION block prints in full below. the two only agree because the create form prefills pickup location from the primary warehouse. edit a SCHEDULED shipment's pickup location to a vendor yard or job-site trailer (a supported edit) -> the printed form states UC Hardware's division address as that third-party location, with the same text twice on one page.

backend/app/models/shipping.py
```diff
@@ -19,6 +48,52 @@ class PackingSlip(Base):
...
+    contractor_contact_phone: Mapped[str | None] = mapped_column(String, nullable=True)
+    ucsh_contact_name: Mapped[str | None] = mapped_column(String, nullable=True)
+    ucsh_contact_phone: Mapped[str | None] = mapped_column(String, nullable=True)
+    sales_order_number: Mapped[str | None] = mapped_column(String, nullable=True)
+
+    # --- Lifecycle stamps ------------------------------------------------------------------------
+    # Who confirmed each leg and when. The actor is the Clerk-authenticated caller (#427), never
+    # something the client named.
+    picked_up_at: Mapped[datetime | None] = mapped_column(nullable=True)
+    picked_up_by: Mapped[str | None] = mapped_column(String, nullable=True)
```
Xilous, line 93: picked_up_by and delivered_by are new actor columns. ACTOR_NAMES in backend/tests/test_resolver_actor_provenance.py (the #427 enforcement) contains picked_by and shipped_by and neither of these. resolver bodies are compliant today (both go current_user -> resolve_display_name), but a future regression that read the actor off `input` would pass the guard. add both names to ACTOR_NAMES.

frontend/src/modules/warehouse/ReceivingHistory.tsx
```diff
@@ -0,0 +1,384 @@
...
+ * run until somebody actually opens the PO - the list itself is deliberately scalars-only, and
+ * eagerly fetching every PO's receives would undo that. Apollo's cache answers a re-expand, so
+ * opening the same row twice is one round trip.
+ */
+function ReceivesPanel({ poId }: { poId: string }) {
+  const { data, loading, error } = useQuery<{
+    poReceivingDetails: { id: string; receiveRecords: ReceiveRecord[] };
+  }>(GET_PO_RECEIVING_DETAILS, { variables: { poId } });
+
+  if (loading) {
```
Xilous: bare `if (loading)` where the parent uses `loading && !data` (line 336). apollo.ts sets defaultOptions.watchQuery.fetchPolicy to cache-and-network and the Collapse uses unmountOnExit, so every expand -> collapse -> re-expand tears down the watch, refetches the full poReceivingDetails payload of
PO
vendor
every line item
every receive
and flashes a spinner over data the cache already holds. the docstring above ("Apollo's cache answers a re-expand, so opening the same row twice is one round trip") is wrong in both halves. matching the parent's guard makes a re-expand paint from cache while revalidating behind it.

frontend/src/modules/shipping/DeliveryRequestFields.tsx
```diff
@@ -0,0 +1,212 @@
...
+          <Stack direction={{ xs: 'column', sm: 'row' }} spacing={2}>
+            <TextField
+              label="Carrier / Tag / BOL"
+              value={details.carrierTagBol}
+              onChange={(e) => set('carrierTagBol')(e.target.value)}
+              disabled={disabled}
+              fullWidth
+            />
+            <TextField
+              label="Weight (lbs)"
```
Xilous, line 97: two weight validation gaps.
- the input is type="number", so an entry the browser rejects (an exponent character, or pasting "812.5 lbs") reads back as '', isWeightInvalid('') is false, and the full-replace update maps '' to null. editing a slip that had weight 812.5 and fat-fingering the box silently clears the stored weight with a success toast.
- nothing bounds the value against Numeric(10,2). a weight of 1e9 passes Number.isFinite, reaches confirm_shipment, and the commit dies on a Postgres numeric overflow, aborting the whole shipment with a raw DB error instead of a field message.

frontend/src/modules/shipping/EditShipmentDialog.tsx
```diff
@@ -0,0 +1,85 @@
...
+    }
+    updateShipment({
+      variables: { input: { id: slip.id, ...deliveryDetailsInput(details) } },
+    });
+  };
+
+  return (
+    <Modal
+      open
+      onClose={onClose}
```
Xilous: onClose goes straight to Modal and then to MUI Dialog, so backdrop click and Escape both dismiss with no unsaved-changes guard, and ShipmentsList mounts this conditionally, so dismissal discards every edit. retype
the address
the eight site answers
both contacts
then click slightly outside the dialog -> all gone silently. the backdrop also stays live while the mutation is in flight (Cancel is disabled, the backdrop is not). save-then-backdrop-click unmounts the dialog and the "updated" toast fires for a dialog the user believes they cancelled.

backend/app/repositories/warehouse/receiving.py
```diff
@@ -591,6 +602,94 @@ def get_back_ordered_items(session: Session, project_id: uuid.UUID | None = None
...
+    reached GP can have no receipts, and a cancelled one is not history anybody reconciles.
+
+    Three grouped aggregates, one row per PO, no relationship iteration. The obvious shape - load the
+    POs and walk `po.line_items` / `po.receive_records` per row - is the N+1 in the perf rules, and
+    this list is cross-project by default, so it is exactly the resolver that would find every PO in
+    the database one SELECT at a time. Each PO's individual receives are NOT returned: the row
+    expands on demand through the existing `poReceivingDetails` query, so a page of a hundred POs
+    costs a hundred rows rather than every receive line ever recorded.
+    """
+    line_totals = (
```
Xilous: both grouped subqueries carry no predicates, so every call aggregates all of po_line_items and all of receive_records before the outer join. the project/status/deleted filters cannot be pushed into them and the po_id indexes go unused; cost grows with total history, re-paid on every History toggle (frontend is cache-and-network). there is also no LIMIT, and ReceivingHistory renders every returned row with no page cap and unmemoized rows (ShipmentsList in this same PR got PAGE = 25). repeat the PO-scope predicates inside each subquery plus the same shown/PAGE tail the sibling uses.

backend/app/schemas/inputs.py
```diff
@@ -493,9 +493,74 @@ class ShipmentItemInput:
...
+    extra_frame_anchors: str | None = None
+    contractor_contact_name: str | None = None
+    contractor_contact_phone: str | None = None
+    ucsh_contact_name: str | None = None
+    ucsh_contact_phone: str | None = None
+    sales_order_number: str | None = None
+
+
+@strawberry.input
+class UpdateShipmentDetailsInput:
```
Xilous, line 529: the 20-field header is enumerated longhand roughly 17 times across the PR:
migration
model
DELIVERY_REQUEST_FIELDS
both inputs (this one restates ConfirmShipmentInput's block verbatim)
types.PackingSlip
the converter
both SDL blocks
PACKING_SLIP_FIELDS
five shapes in deliveryRequest.ts
two drift directions fail silently:
- a field missing from DELIVERY_REQUEST_FIELDS is accepted by GraphQL and never persisted or cleared
- a field missing from PACKING_SLIP_FIELDS saves but reads back undefined and prints blank
collapse most copies with a shared Strawberry input base for the two mutations, a converter built by iterating DELIVERY_REQUEST_FIELDS, and one frontend field list the five shapes derive from.

Issue #425, opened 2026-07-29
quarantine projects with broken GP job setup (dangling cost-code account indexes)


Xilous:
receiving PO0000070 (TUBC) failed, eConnect 4612 "Invalid Account Index" from taPopRcptLineInsert. the PO line was stamped INVINDX=1617 at registration by the WennSoft PO integration, copied from JC00701 (job 23093, cost code 210-200-2). 1617 does not exist in UBC/TUBC's chart, it is a UCSH index (4000-500-00, Cost of Goods Sold).

jobs replicate UCSH -> UBC via SQL-level copy run from client-side tooling (no SQL Agent job on the server). before ~2023 it copied JC00701 account indexes raw. production UBC has 1504 dangling JC00701 rows across 62 jobs (created 2014-2023). jobs copied 2023 onward carry UBC's own accounts, same account numbers with division segment 100 instead of 500:
1617 -> 96
1622 -> 101
1623 -> 102
1626 -> 105
1628 -> 107
1656 -> 138
TUBC is a clone of UBC and mirrors this. a job-cost PO against one of the 62 jobs registers fine but can never be received.

build a per-project GP setup health check plus quarantine.

the health check runs per job, per company, all read-only:
- job has at least one active cost code in JC00701
- every active cost code's WS_Account_Index_1 is 0 or exists in GL00105
fail either -> unhealthy. index 0 passes: eConnect defaults the account at receipt time, only a non-zero dangling index is the proven failure mode.

relay
- new read-only op job_setup_health returns per-job verdicts for all jobs in one query (JC00701 left join GL00105 grouped by job), with per-job detail of the broken cost codes and indexes
- create_po validates the chosen cost code's account index next to the existing cost_code_on_job guard, new error cost_code_account_invalid
- create_receipt pre-flights the PO lines' POP10110.INVINDX against GL00105, returns a job-setup error naming the account

backend
- gp_job_sync stamps the verdict on every project each pass into new columns:
  gp_setup_ok (nullable bool)
  gp_setup_detail (which cost codes are broken)
  gp_setup_checked_at
- columns exposed on the Project GraphQL type
- gp_setup_ok = false blocks every project-scoped mutation server-side:
  import
  start a task
  PO registration
  receiving
  shipping
- gp_setup_ok null (never checked) does not quarantine, a down relay must not freeze all of Nexus
- register_po_in_gp re-checks the job live at submit time, authoritative gate

frontend
[...]

PR #455, opened 2026-08-02
feat(warehouse,po): draft receives with manager approval, and a keep-or-ship decision

Xilous:
receiving splits into two actions. an approved receive asks where the hardware goes.

part A - drafted receives

counting a delivery writes a ReceiveDraft. Nexus-only, no GP write, no inventory.
a Warehouse Manager approves; approval runs the GP-first pipeline: relay create_receipt -> RCT###### -> persist -> inventory credit.
every receive is a draft, including a manager's own.

new Clerk role Warehouse Manager.
approveReceiveDraft and rejectReceiveDraft take that role OR Admin/Manager - first any-of requirement in ROOT_FIELD_POLICY. the table's value type is now `str | frozenset[str]`, enforce_root_field gained a frozenset branch. no implicit admin bypass anywhere; Admin/Manager is named in the set.

createReceive is removed from the schema. approveReceiveDraft IS that resolver, with a claim in front of it. an old tab during deploy skew gets a GraphQL validation error, nothing reaches GP.

a database lock cannot span the relay round trip, hence the claim:

- approve flips the draft to APPROVING in its own committed transaction before calling the relay.
- a second approver bounces off with a conflict naming the reviewer holding it.
- a retry carrying the same idempotency key resumes.
- a failure BEFORE the relay call releases the claim.
- a DISPATCHED relay failure does not release - GP may hold the receipt.
- the claim sums other APPROVING / APPROVED-unpersisted drafts on the same PO line and refuses before GP is touched. two deliveries against one PO is legitimate; approving both past what the line still owes is not.

ReceiveRecord.received_by is the DRAFT AUTHOR's name, captured at creation. the approver goes on the draft's review fields, from the Clerk token (#427).

part B - keep-or-ship decision

on approve + persist, the person who raised the PO is asked whether the shipment stays in the project's inventory or ships out now.

the row is minted inside create_receive, not the resolver, so both routes into that persist produce it:
- online approval
- outbox worker draining a receipt queued while the relay was down

SHIP_OUT records the choice and nothing else. only the schedule knows which opening and leaf a fungible quantity is owed to, so the frontend deep-links into Start a Task, scoped to the project, where identity is re-attached (docs/HARDWARE_IDENTITY_LIFECYCLE.md).

purchase_orders.created_by_user_id stamps the originator going forward, at both creation paths - finalize_import_session, createDraftPo - from the Clerk token. POs raised before it existed keep a NULL and the read falls back to matching the PO's GP buyer against the CALLER's own gpBuyerId. read time, not write time: resolving it in the persist would put a Clerk round trip inside the transaction of a GP receipt that has already posted.

stock POs (no project) get drafts but no decision.

frontend

ReceiveModal becomes draft creation. the relay chip, queued-outbox panel, GP receipt number and eConnect error detail move to the new ReceiveDraftReviewModal, where the posting happens.

the GP-job quarantine banner warns at draft time and hard-blocks at approval. a PO that is not in GP at all hard-blocks both.

new shared ReceiveLinesEditor + receiveLines.ts behind all three editors (count, review, author edit), extracted behavior-neutrally first.

new Receive Approvals page, manager-gated, Pending | Rejected, no Approved view - Receiving > History already owns posted receives.
new My Drafts view on the Receiving page, ?view= in the URL.
new Shipment Decisions page in the PO module, banner on the PO list.

NotificationBell gains a type-to-route map for the two person-targeted types. unmapped types keep mark-read-only.

Warehouse sidebar item and the User Management role list gain 'Warehouse Manager'.

migrations

078
receive_drafts
receive_draft_line_items
receive_decisions
purchase_orders.created_by_user_id
no backfill

079 adds to notification_type
RECEIVE_DRAFT_SUBMITTED
RECEIVE_DRAFT_REJECTED
RECEIVE_DECISION_REQUIRED

known trade-offs

- a DISPATCHED relay failure mid-approve parks the draft in APPROVING, resumable only with the same key. same residual window createReceive already had, now visible.
- a terminal outbox failure after a queued approve leaves the draft APPROVED with no receive record. the existing outbox failure UI is the backstop, no auto-revert - re-approving after an ambiguous failure risks double-posting GP.
- person-targeted notifications are visible in everyone's bell, so their messages name the person.
- GET_OPEN_POS pending math ignores draft-held quantities; the "Draft pending" chip on the PO row mitigates.
- notifications.project_id is NOT NULL, so stock-PO drafts skip notifications; the approvals queue is the backstop.


backend/app/schemas/warehouse.py
```diff
@@ -701,50 +860,140 @@ def warehouse(self, info: strawberry.Info, id: strawberry.ID) -> Warehouse | Non
...
+        What the split adds:
+
+        - `received_by` is the DRAFT AUTHOR's name, read off the draft. The approver's identity goes
+          on the draft's review fields. Neither comes from the client (#427).
+        - The claim (`_claim_draft_for_approval`) is committed before the relay call, because a
+          database lock cannot span a network round trip. It is what makes a second approver bounce
+          off instead of posting a duplicate receipt, and it is why a failure BEFORE the relay call
+          releases it - a draft parked in APPROVING for a validation error nobody can see would be a
+          worse outcome than the error itself.
+        - A DISPATCHED relay failure deliberately does NOT release: GP may hold the receipt, so the
```
Xilous: relay failure that is not `RelayUnavailableError` leaves the draft stuck in `APPROVING` forever. `relay_call` also raises `RelayCallError` (relay answered ok=false, eConnect rejected the receipt), `RelayOpUnsupportedError`, `RelayTimeoutError` - none subclass `RelayUnavailableError`. only the `except RelayUnavailableError` arm is handled, so those three propagate past the claim without calling `_release_draft_claim`.

no way out of `APPROVING`:
- `reject_receive_draft` and `_assert_can_edit` require `PENDING_APPROVAL`
- `delete_receive_draft` refuses `_IN_FLIGHT_STATUSES`
- `resubmit_receive_draft` requires `REJECTED`
- `claim_for_approval` conflicts unless the caller carries the identical idempotency key

that key only lives in a `useRef` in the review modal, and `ReceiveApprovalsPage` makes the row unclickable once status is not `PENDING_APPROVAL` - closing the modal or reloading the tab bricks the draft permanently. it also holds the PO line's pending quantity hostage through `_assert_no_conflicting_claim`, which then refuses every later draft on that line.

an eConnect rejection is the #425 failure this codebase already sees, and GP definitively did not commit in that case.

release the claim for `RelayCallError` and `RelayOpUnsupportedError`. keep retain-on-ambiguous for a dispatched failure and a timeout.

backend/app/schemas/warehouse.py
```diff
@@ -701,50 +860,140 @@ def warehouse(self, info: strawberry.Info, id: strawberry.ID) -> Warehouse | Non
...
+            gp_company, payload = await asyncio.to_thread(
+                _prepare_create_receive,
+                po_id=ctx.po_id,
+                received_by=ctx.author_name,
+                line_items_data=ctx.line_items_data,
+            )
+        except Exception:
+            # Nothing reached GP, so the draft belongs back in the queue with the reason surfaced to
+            # the reviewer rather than parked in a state only a retry can clear.
+            await asyncio.to_thread(_release_draft_claim, draft_id, key)
```
Xilous: `_prepare_create_receive` runs before the `state.relay_result` check, so on a resumed approval it can release a claim for a receipt GP already holds.

attempt 1 claims -> posts GP receipt -> `record_relay_result` succeeds -> `_persist_create_receive` fails (transient DB error) -> attempt 2 same key finds `result_id` still `None` -> falls through to the claim path -> re-runs `_prepare_create_receive` -> `validate_receive_eligibility` now refuses (another receive landed against the line, PO closed, `require_gp_setup_ok` flipped) -> this `except` releases the claim -> draft returns to `PENDING_APPROVAL` while GP holds an unbooked receipt -> manager approves again with a fresh idempotency key -> `gp_idempotency.load` returns `None` -> second GP receipt posted.

two fixes stack:
- skip `_prepare_create_receive` entirely when `state.relay_result is not None`
- do not release when the ledger already holds a relay result

frontend/src/modules/warehouse/ReceiveDraftReviewModal.tsx
```diff
@@ -0,0 +1,477 @@
...
+    setConfirmOpen(false);
+    setMutationError(null);
+    setGpError(null);
+    setSubmitting(true);
+
+    try {
+      // Save the edits FIRST, so what gets approved is what is on screen. A failure here stops
+      // everything: nothing has reached GP, and approving the un-edited draft would silently post
+      // numbers the reviewer had just changed.
+      if (isDirty) {
```
Xilous, line 218: `initialSignature` is never refreshed after a successful `updateDraft`, so a retry after a failed approve re-sends the update against a draft that is now `APPROVING`.

manager corrects a quantity -> `isDirty` true -> `updateDraft` commits -> `approveDraft` claims the draft, fails with a dispatched relay error or a GP rejection -> draft left `APPROVING` -> manager presses Approve again -> `isDirty` still true, nothing reset the signature -> `updateDraft` fires first -> backend refuses it (`_assert_can_edit` only allows `PENDING_APPROVAL` or `REJECTED`) -> exception caught and reported as "Approving this receive failed" -> `approveDraft` with the same idempotency key never reached, which is the one recovery path the design provides.

reset `initialSignature` after the update succeeds, or send the approve first and the update only while the draft is still pending.

backend/app/repositories/warehouse/receive_drafts.py
```diff
@@ -0,0 +1,527 @@
...
+        elif draft.status == ReceiveDraftStatus.APPROVING:
+            raise ConflictError(f"{draft.reviewed_by_name or 'Another reviewer'} is approving this draft right now")
+        elif draft.status == ReceiveDraftStatus.APPROVED:
+            raise ConflictError("This draft has already been approved")
+        else:
+            raise InvalidStateTransitionError(
+                f"Only a draft awaiting approval can be approved, got {draft.status.value}"
+            )
+    else:
+        _assert_no_conflicting_claim(session, draft)
```
Xilous, line 470: `_assert_no_conflicting_claim` reads the other drafts' statuses inside the same uncommitted transaction that writes this draft's own claim, so under READ COMMITTED two concurrent approvals on the same PO line both see no conflict. both `UPDATE`s go to different rows, so neither blocks the other, and `_claim_draft_for_approval` does not commit until after the assertion has run. both claims commit, both approvals post a GP receipt, and the second `create_receive` raises "Receive quantity exceeds pending quantity" after GP has committed - the split-brain the check exists to prevent.

take the PO line rows with `lock_rows` before the aggregate, or commit the claim and re-check after.

backend/app/schemas/converters.py
```diff
@@ -128,6 +133,77 @@ def receive_record_to_type(rr) -> ReceiveRecord:
...
+    return ReceiveDecision(
+        id=strawberry.ID(str(decision.id)),
+        status=decision.status,
+        decision=decision.decision,
+        po_id=strawberry.ID(str(decision.po_id)),
+        po_number=po.po_number if po is not None else None,
+        project_id=strawberry.ID(str(decision.project_id)),
+        receive_record_id=strawberry.ID(str(decision.receive_record_id)),
+        receipt_number=receive_record.receipt_number if receive_record is not None else None,
+        received_at=receive_record.received_at,
```
Xilous, line 197: `receive_record` is guarded for `receipt_number` on the line above, then dereferenced unguarded three times:
`received_at`
`received_by`
`.line_items`

either the `None` case is real and this raises `AttributeError` instead of returning the decision, or it is not and the guard on line 191 is misleading. `_load_decision` returns whatever `session.scalars(...).first()` gave it, so a receive record deleted or not yet visible produces a 500 on `decideReceiveDecision` and on `myReceiveDecisions`.

frontend/src/modules/import/ImportWizard.tsx
```diff
@@ -781,6 +794,50 @@ export default function ImportWizard({ open, project, onClose }: ImportWizardPro
...
+    if (!canStartFromLatest || parser.state !== 'idle') return;
+    autoStartedRef.current = true;
+    // eslint-disable-next-line react-hooks/set-state-in-effect -- pressing "use last schedule" for the user; there is no external event to hang it off
+    handleLoadFromLatest();
+  }, [open, autoStartFromLatest, canStartFromLatest, parser.state, handleLoadFromLatest]);
+
+  useEffect(() => {
+    // Once hydrated with a purpose already chosen, both of the upload step's jobs are done - drop
+    // the user straight on the openings they came to pick.
+    if (!open || !autoStartFromLatest || !hydratedFromPersisted || !purpose) return;
```
Xilous, line 842: this effect has no one-shot guard, so it re-fires every time the user reaches the upload step and bounces them back to `openings`. the other two deep-link effects use `seededPurposeRef` / `autoStartedRef`; this one only checks `activeStepId !== 'upload'`.

after a deep-linked entry the user is on `openings` with `purpose` set and `hydratedFromPersisted` true. `handleBack` walks `openings -> purpose -> upload`, and the moment `upload` is reached this effect sets the step forward again. the upload step is unreachable for the rest of the session, so a user who deep-linked to the wrong schedule cannot change source.

add a consumed-once ref like the sibling effects.

frontend/src/components/NotificationBell.tsx
```diff
@@ -89,6 +106,11 @@ export default function NotificationBell() {
         ],
       });
     }
+    const to = NOTIFICATION_LINKS[notification.type];
```
Xilous, line 111: `recipient_role` is an audience tag and the bell queries with no recipient filter, so every signed-in user sees `RECEIVE_DRAFT_REJECTED` and `RECEIVE_DECISION_REQUIRED` in their own bell. adding navigation makes those rows look actionable to everyone: `handleNotificationClick` marks the notification read first and then navigates, so a bystander who clicks lands on an empty My Drafts or Shipment Decisions page and has silently cleared the unread signal for the person who owed the action.

consider skipping the mark-read and the navigation when the notification's recipient is a Clerk user id that is not the caller's.

`RECEIVE_DRAFT_SUBMITTED` has no mapping, and it is the one of the three actionable by its whole audience. `/app/warehouse/receive-approvals` is the obvious target.

backend/app/schemas/warehouse.py
```diff
@@ -766,39 +1015,73 @@ async def create_receive(self, info: strawberry.Info, input: CreateReceiveInput)
...
+        Recording the answer is all this does. SHIP_OUT does not build a shipping-out request: only
+        the hardware schedule knows which opening and leaf a fungible quantity is owed to, so the
+        frontend takes the recorded choice and deep-links into Start a Task, where that identity is
+        re-attached (docs/HARDWARE_IDENTITY_LIFECYCLE.md).
+        """
+        from app.models.enums import ReceiveDecisionChoice as ReceiveDecisionChoiceDB
+
+        user = current_user(info)
+        actor_name = resolve_display_name(user["user_id"])
+        gp_buyer_id = user_repository.get_user_gp_buyer_id(user["user_id"])
```
Xilous: `get_user_gp_buyer_id` is called unconditionally, but `decide_receive_decision` only uses it on the fallback arm where `decision.target_user_id` is `NULL`. every PO raised since this PR stamps `created_by_user_id`, so the common path pays a Clerk `/users/{id}` HTTP round trip it never reads, on top of the `resolve_display_name` call directly above it. `get_user_gp_buyer_id` also raises `AppError("CLERK_UNAVAILABLE")` on any `httpx.HTTPError`, so a Clerk hiccup refuses a decision that needed nothing from Clerk.

resolve it lazily inside the repository, only when `target_user_id` is `NULL`.

frontend/src/modules/po/index.tsx
```diff
@@ -750,6 +752,14 @@ function POListPage() {
     return ids;
   }, [outboxData]);
 
+  // Shipments this user has to say where to put. Scoped to the caller server-side, so the count is
+  // either zero or theirs.
+  const { data: decisionsData } = useQuery<{ myReceiveDecisions: { id: string }[] }>(
+    GET_MY_RECEIVE_DECISIONS,
+    { fetchPolicy: 'cache-and-network' },
+  );
+  const pendingDecisionCount = decisionsData?.myReceiveDecisions?.length ?? 0;
```
Xilous, line 761: `GET_MY_RECEIVE_DECISIONS` runs with `cache-and-network` on every PO list mount for every user, and the resolver behind it makes a synchronous Clerk `/users/{id}` HTTP call per request (`user_repository.get_user_gp_buyer_id`). that is a Clerk round trip on the render path of the most-visited page in the module, for a banner that is usually zero. `get_user_gp_buyer_id` raises `CLERK_UNAVAILABLE` on any HTTP error, so a Clerk blip turns the PO list into an error state.

a scalar `myReceiveDecisionCount` field, or resolving the buyer id only when the stamped-target arm returns nothing, keeps the banner without the round trip.

backend/app/repositories/warehouse/receive_drafts.py
```diff
@@ -0,0 +1,527 @@
...
+    _assert_can_edit(draft, actor_user_id, actor_is_manager)
+
+    validate_receive_eligibility(session, draft.po_id, draft.created_by_name, line_items_input)
+    po = (
+        session.scalars(select(POModel).options(selectinload(POModel.line_items)).where(POModel.id == draft.po_id))
+        .unique()
+        .first()
+    )
+
+    draft.warehouse_id = warehouse_id
```
Xilous: `update_receive_draft` assigns `draft.warehouse_id = warehouse_id` unconditionally, and `UpdateReceiveDraftInput.warehouse_id` defaults to `null`. any caller that sends only corrected line items silently clears the draft's destination warehouse, and the approval then falls back to `get_primary_warehouse_id`, so the hardware is booked into the wrong building with no error. both current callers pass it back, so this is latent rather than live, but the input cannot distinguish "leave it alone" from "clear it".

make the assignment conditional on the argument being present, or document that omitting it clears.

frontend/src/modules/warehouse/ReceivingPage.tsx
```diff
@@ -228,6 +253,25 @@ export default function ReceivingPage() {
     skip: !showReceive,
   });
 
+  // Everybody's pending drafts, for the "already counted" chip on the PO rows. Scoped to PENDING
+  // rather than mine, because the point of the chip is to stop a SECOND person re-counting a
+  // delivery that is already in the queue.
+  const { data: pendingDraftsData } = useQuery<{ receiveDrafts: ReceiveDraft[] }>(GET_RECEIVE_DRAFTS, {
```
Xilous: this query pulls every pending draft in the system with all `lineItems` and their `locations` (the shared `RECEIVE_DRAFT_FIELDS` selection); the page only counts drafts per PO for a chip and a button badge. every row also makes the backend build the full `ReceiveDraftLineItem` / `ReceiveDraftLocation` types.

repo CLAUDE.md, GraphQL / SQLAlchemy Performance Rules, rule 3: "Don't request fields you don't use... Prefer scalar summary fields (openingCount) for list views."

a `pendingDraftCountsByPo` scalar field, or a slim selection of `{ id poId totalQuantity }`, does the same job.

backend/app/schemas/warehouse.py
```diff
@@ -174,11 +196,101 @@ def _persist_create_receive(*, key, po_id, received_by, line_items_data, warehou
...
+
+    Committing here rather than at the end is the whole point: the claim has to be visible to another
+    request while this one is waiting on the relay, which is exactly the window a second approver
+    would otherwise post a duplicate GP receipt in. Returns the context the GP pipeline needs, plus
+    the JSON copy of it for a queued receipt's outbox row - both read inside this session, so nothing
+    detached is touched afterwards.
+    """
+    with SessionLocal() as session:
+        ctx = warehouse_repository.claim_for_approval(session, draft_id, reviewer_user_id, reviewer_name, key)
+        json_lines = warehouse_repository.json_line_items_for_outbox(session, draft_id)
```
Xilous: `json_line_items_for_outbox` re-runs `_get_draft` (a second SELECT plus a `selectinload` of the line items) for data `ctx.line_items_data` already holds, on the hot approve path. `ApprovalContext.line_items_data` is the same list with the UUIDs unstringified, so the JSON copy can be derived from `ctx` without touching the database again.

frontend/src/modules/warehouse/ReceiveOutcomePanels.tsx
```diff
@@ -0,0 +1,97 @@
...
+            </Box>
+          )}
+        </Typography>
+      ))}
+    </Box>
+  );
+}
+
+/** The relay was down: the receipt is on the durable outbox and will post itself (#353 PR E). */
+export function ReceiveQueuedAlert({
```
Xilous: `ReceiveQueuedAlert` and `ReceivePostedAlert` are exported and never imported anywhere; only `PostedReceiptLines` is used, by `ReceiveDraftReviewModal`. that is about two thirds of the new file. delete them, or wire them into the review modal's queued and posted alerts, which currently duplicate the same copy inline.

Issue #451, opened 2026-07-31
shipping out interface


Xilous:
there needs to be an interface for the use case below

at the time of a shipping out request being created, in theory there are possibly these entities in the project's inventory:

assembled door leafs, loose site hardware, loose shop hardware (let me know if i missed anything here)

during shipping out request creation, the user should be somehow building out which items from the above types of inventory to be included in the ship out request.

a really useful feature as well is since the creation of a shipping out request is hardware schedule driven, then technically nexus should be able to at least intuitively auto-suggest/add all of the site hardware that is available in the current project inventory quantities that belongs to the openings/door leafs that were selected from the hardware schedule - showing the user "this is what you have in your cart so far, but these door leafs you selected have these as the site hardware and these as the shop hardware but arent already assembled onto the opening (shop assembly skipped this part, maybe because it wasnt available at the time), and are still not in your cart, for each line of site or shop hardware, these are the ones that are on the way (ordered but not here yet), these are the ones that are in inventory and you can add to your shipping request right now.

flesh it out and come up with more intuitive ideas on how this should work overall

Xilous:
A shipping request is just a communication of what needs to be present on site 

Shipping module users/managers still need to determine what goes and when

So almost like a SOR just moved stuff from inventory to Shipping Out's staging. From there it sits until it's organized into shipments


Shipments also has containers in it:
Skid
Door Cart
Box
Envelope
Bundles

If a door leaf is on a skid, there needs to be a stacking list that indicates in which order the door leafs are stacked. Up to a maximum 30 door leafs

^ SO users drag and drop experience. The step will be referred to as "organizing a shipment". SO user creates "containers", and can name it either "Skid 1" or "Door Cart 3" etc



Shipment method management - user crud


SO user can also request to trigger to add items from a projects inventory to their staged for shipping pool. Also the ability to just add items from inventory to a ship out request before even approving the ship out request itself

Xilous:
all four phases on master. closing.

- #463 phases 1-2, coverage-aware request builder, raise or edit a request from Shipping off project inventory
- #464 phase 4, shipment method management
- #465 phase 3, staging workspace and shipment containers
- #469 rest of phase 3, session cart retired, containers a shipment went out in printed on the Delivery Request and shown on the Shipments list

answering the "did I miss anything" in the body - three entity types, complete.
assembled door leaves -> OpeningItem
loose hardware -> InventoryLocation, fungible
[...]

Issue #311, opened 2026-07-22
Epic: door-leaf awareness across shop assembly, inventory, and shipping


Xilous:
opening -> leaf (1 or 2) -> hardware items
Nexus has no leaf concept today.
thread leaf through shop assembly + inventory + shipping. leave PO + receiving alone.

TITAN import (Start a Task) -> PO request -> warehouse receive -> shop assembly -> inventory -> shipping out

opening = 1 or 2 door leaves; every door + hardware item belongs to exactly one leaf.
two leaves of a pair assembled, stocked, shipped separately.

validated against `testing/fixtures/contracterp-74.xml` (11.5MB) + current code.

current state
- parser (`frontend/src/workers/parserLogic.ts`, `extractHardwareItems`) reads `Material_List` + `Material_List_Fields`, iterates `Assignments/Assignment` for `@_Code` / `Material_ID` / `Qty_Per` / `Phase_Code`; never touches `<Attributes>` -> leaf dropped
- `HardwareItem` (`backend/app/models/hardware.py`) - `opening_id` only, no leaf column
- `ShopAssemblyOpening` / `ShopAssemblyOpeningItem` (`backend/app/models/shop_assembly.py`) per-opening + flat
- `complete_opening` (`backend/app/repositories/shop_assembly_repository.py`) emits one `OpeningItem` (`quantity=1`) per opening, one location, one state, flat `OpeningItemHardware`
- shipping keys on `opening_item_id` as atomic ship unit
- `complete_pull_request` (`warehouse/pull_requests.py`) sets `SHIP_READY`
- `confirm_shipment` (`shipping_repository.py`) writes `PackingSlipItem` per `opening_item_id` + sets `SHIPPED_OUT`
- `OpeningItemState` `IN_INVENTORY -> SHIP_READY -> SHIPPED_OUT` per opening

source data facts, from `testing/fixtures/contracterp-74.xml`
- openings in `Assignments/Assignment_Level_3` (Code, Single_Pair, Hand); hardware in `Detail/Material_List`
- 2,919 Material_Lists, each with exactly one `<Attributes>` block (sibling of `<Assignments>`, alongside door attrs Height / Hand / Door Thickness); 29,126 assignment nodes
- leaf is a Material_List-level field, read once per ML, not per-assignment. `<Attribute Code="Leaf">` = Leaf 1 / Leaf 2 lives in the per-ML `<Attributes>` block. 0 of 29,126 assignment nodes carry it inline.
[...]

Issue #341, opened 2026-07-26
Shop assembly slice 3: replacement loop closure + shipping deficiency guard


Xilous:
Part 3 of 6. Depends on #339 (sa_opening_item_id stamp) and slice 2 (persisted per-item state). Today the deficiency flow dead-ends: the PR-REPL pull draws replacement hardware from inventory but nothing ever attaches it to the leaf, and `complete_pull_request` finds no openings hanging off a PR-REPL pull so its completion is a silent no-op.

## Scope
1. **Restore expectation on replacement arrival.** When a PR-REPL pull completes, for each line carrying `sa_opening_item_id`: reduce that item's `deficient_quantity` by the line quantity (the unit re-enters the leaf's expectation). If the leaf is still PENDING/IN_PROGRESS it simply reappears as remaining work in My Work — no new UI needed.
2. **Replacement install after leaf completion.** If the leaf is already COMPLETED, surface a "replacement install" work item (assignee = the leaf's last assembler, reassignable). Completing it appends an `OpeningItemHardware` row to the existing `OpeningItem` — the one legitimate post-completion write. Design the state so a future un-install/correction flow could reuse it, but do not build corrections.
3. **Shipping guard (business decision: warn + confirm override, never silent, never hard-block).** An `OpeningItem` with outstanding deficiencies (derivable from its source `ShopAssemblyOpeningItem` rows) currently ships as if whole. The shipping wizard must show an "incomplete — awaiting replacement" flag on such leaves and require an explicit confirm to ship anyway. Deliberate short-shipping stays possible (reallocation exists for it).
4. **Replacement arrives after the leaf shipped.** If the `OpeningItem` is SHIPPED_OUT when the replacement pull completes, do not strand the hardware silently: detect, flag, and surface it (notification + visible state) for routing into the reallocation/site-shipment world. Full routing is out of scope; visibility is in scope.

## Constraints
- CLAUDE.md architecture rules; typed errors; migration (if any) with working downgrade; DB-backed tests for all four paths; frontend tests for the wizard guard; update SDL docs, `docs/HARDWARE_IDENTITY_LIFECYCLE.md`, `testing/CLAUDE.md`.

PR #463, opened 2026-08-03
feat(shipping): coverage-aware request builder, and raising/editing requests from the Shipping module (#451)

Xilous:
phases 1 and 2 of #451. remaining phases (persistent staging workspace with containers and stacking lists, shipment method management) land as their own PRs.

phase 1 - coverage-aware request builder

the two things the shipper had to remember and the step never said
- site hardware never goes through shop assembly at all, so it is always still owed loose, however complete the assembled leaf looks
- shop hardware assembly skipped (not available at the time) never got fitted, so it also still has to travel loose

new query `shippingCoverage(projectId, openingNumbers)` in `backend/app/repositories/shipping_coverage.py`, one row per door leaf of the selected openings. each row carries:
leaf status (NOT_ASSEMBLED / IN_INVENTORY / SHIP_READY / SHIPPED_OUT)
assembled unit
whether a live shipping request already claimed it
one line per product

per product line:
owedQuantity = the schedule
installedQuantity = the assembled leaf
suggestedQuantity = owed - installed, floored at 0
onOrderQuantity = placed but not received across the project's live POs
classification

suggested = owed - installed is one formula for all three classifications. site hardware is never fitted at the bench so installed is 0 and the whole quantity is owed, shop hardware nets off what was fitted leaving exactly what was skipped, unclassified degrades to the same thing.

availability is deliberately not in the response. projectInventoryAvailability stays the single number the creation gate is applied against (#342); a second figure computed here at a different instant is what would let the coverage panel say 3 while the shortfall alert says 2. the builder joins the two by (hardwareCategory, productCode).

six queries regardless of how many openings are selected
openings
schedule items (grouped)
assembled units + their hardware (selectinload)
live shipping claims
on-order aggregate

leaf enumeration = union of three sources, so a leaf cannot go missing because one of them is silent
the opening's leaf_count
the leaves the schedule names
the leaves that were actually assembled

leaf-less schedule lines fold onto the lowest leaf, same rule the wizard applies when building shop-assembly work units.

ShippingPRsStep now shows, per draft, what is on the request so far -> assembled leaves (unchanged) -> loose lines grouped by
Site hardware
Shop hardware not assembled onto the leaf
Unclassified

each group carries a hint line. each loose row is chipped with what it is owed, what is in inventory (and how much is spoken for), what is on the way, or "none ordered" when there is neither.

quantities are per-line and editable, capped at what is free once every other line in the session has claimed its share, so a partly-covered line can be sent short instead of tripping the shortfall gate with no way forward. "Add all available" fills a whole group, tracking what earlier rows took.

phase 2 - raising and editing a request from the Shipping module

two mutations, both landing in a new `backend/app/repositories/shipping_requests.py` that the wizard's finalize now calls too:
- `createShippingOutRequest` raises one straight off project inventory
- `editShippingOutRequest` rewrites a PENDING one's lines

extracting the guards was the point rather than a side effect - every one of them is a property of the request, not of the screen that composed it
- one physical leaf ships once
- a leaf on the bench cannot leave the building
- an empty request is refused
- loose lines have to fit what is genuinely free

the edit is a full replace, and it releases the request's own reservation before re-gating. without that ordering, trimming a line from 4 to 3 would be measured as asking for 3 more against stock this very request is holding.

migration 081 makes opening_number nullable on shipping_out_request_items and pull_request_items. inventory is keyed by project, warehouse, category and product and carries no opening; a line raised off a shelf has nothing to attribute, and writing one in anyway would be a claim the schedule never made. the downgrade backfills "(unattributed)" before restoring NOT NULL rather than deleting rows.

everything downstream that grouped by opening_number now reads those lines as unattributed rather than blank
pick sheet
ship-ready pool (sorted last, as its own group)
cart
leafIdentity

the Shipping module's Requests page gains a "New request" button and an Edit action on pending requests, both opening the same builder dialog - "what should this request contain" is the same question either way.

`gate_on_available_inventory` moved from import_repository into the warehouse package, beside the arithmetic it applies, since both request types and both composing screens gate identically.

all new fields are SIGNED_IN. one migration (081), additive, downgrade tested by CI's upgrade -> downgrade -> upgrade.

the issue's three inventory types are complete: assembled door leaves (OpeningItem), loose hardware (InventoryLocation, fungible). site vs shop is the schedule's HardwareItem.classification, not a property of inventory, which is why coverage needed a new query rather than a filter. adjacent but not new types: frames (out of scope per #311), stock-pool items (become project inventory before shipping), returned items (restock as inventory again).

`backend/tests/test_shipping_coverage.py` covers 17 cases:
site vs shop suggestion
over-supplied leaf
hardware on the leaf the schedule dropped
pair enumeration
leaf_count-only leaves
leaf-less folding
legacy null-leaf matching
live-beats-shipped
on-order with draft POs excluded and over-receives not cancelling backorders
project scoping

`backend/tests/test_shipping_request_composition.py` covers:
a loose line with no opening
reserving exactly as the wizard does
over-claim refused
another request's claim not free
empty request refused
leaf already on a live request
leaf already shipped
opening taken from the leaf not the client
duplicate request number
adding a line
reducing a line not gated against its own stock
a refused edit leaving the original claim
keeping a leaf across an edit
emptying refused
accepted request not editable
edited lines carried onto the pull

frontend tests
- availability gating suite gains a coverage-builder describe block
- `RequestBuilderDialog.test.tsx` covers the composition dialog
- ImportWizard walk asserts the coverage-derived loose list

verified end to end on the pr-463 Railway environment against TUBC, real UI.

seeded
schedule import over opening 0501-EX (a pair)
two single-line POs registered in GP (PO0000084, PO0000085) and received
one shop-assembly request through pick -> stage -> assemble -> complete, leaf 2 assembled carrying its Housing, missing the Concealed Vert

coverage query read live off that leaf
- leaf 2, IN_INVENTORY: `NB SG 56 MD8610 J ETJ US32D LHR` owed 1, installed 0, suggested 1 - the shop hardware assembly skipped, still owed loose
- leaf 2, same leaf: `6042 112 US32D` owed 1, installed 1, suggested 0 - fitted at the bench, nothing loose needed
- leaf 1, NOT_ASSEMBLED: everything owed in full
- pair enumerated as two leaves from leaf_count, per-leaf line attribution (24 lines vs 17)

flipping exactly one of two POs to a placed status gave onOrderQuantity to only that PO's five products; the still-DRAFT one contributed nothing, which is the DRAFT exclusion working.

builder step renders all three groups with their hint lines
- line with stock shows "OWED 2 · 2 IN INVENTORY · leaf 1: 1, leaf 2: 1" and an enabled "Add 2"; adding moves the summary to "1 loose line(s), 2 unit(s)"
- empty shelf -> all 28 Add buttons disabled
- quantity 9 against 2 free fires the shortfall alert "Keyswitch MKA: need 9, 2 available - short 7" and disables Next; back to 2 clears it
- Finalize created the request, reservation held

phase 2 ran the whole chain
- new request off inventory -> line with `openingNumber: null`, `createdBy` from the signed-in caller, reserving 2 of 2 on hand
- edit asking 5 against 2 refused with the per-combo shortfall, original 2-unit claim intact
- edit trimming 2 to 1 succeeded and freed a unit (reserved 1, available 1) - the release-before-re-gate ordering, which would have read as asking for 1 more if it ran the other way round
- accept carried the null opening onto the pull line; pick sheet renders "Unattributed", the schedule-driven request beside it renders "0501-EX · L2"
- confirming the pick deducted inventory, Mark as Pulled -> ship-ready pool, Ship tab's Loose Hardware grid shows it under "Unattributed"

two things turned up that are not this PR's
- GP register dialog's button is inert rather than disabled while validation fails, and a fresh PR environment has no buyer assignments, so the first registration anywhere silently does nothing until a buyer is assigned
- PO with more than one line fails GP registration with eConnect 9191 on taPoLine, single-line succeeds; five registrations isolated it to the line count, not the item codes. worth a bug of its own, workaround is the register dialog's Remove-line-item buttons

both written up in testing/CLAUDE.md.


frontend/src/modules/shipping/RequestBuilderDialog.tsx
```diff
@@ -0,0 +1,455 @@
...
+      itemType: line.itemType,
+      openingNumber: line.openingNumber ?? null,
+      openingItemId: line.openingItemId ?? null,
+      leaf: line.leaf ?? null,
+      hardwareCategory: line.hardwareCategory ?? null,
+      productCode: line.productCode ?? null,
+      requestedQuantity: line.requestedQuantity,
+    }));
+    if (editing) {
+      editRequest({ variables: { input: { id: request.id, items, acknowledgeIncompleteLeaves: true } } });
```
Xilous: both mutations send acknowledgeIncompleteLeaves: true unconditionally. the #341 guard can never fire from this dialog - a leaf awaiting a replacement, or carrying units that were never pulled, goes onto a request with no warning.

send false -> catch the ValidationError -> show its message -> re-send with true on confirm, as the wizard does.

frontend/src/modules/shipping/RequestBuilderDialog.tsx
```diff
@@ -0,0 +1,455 @@
...
+  leaf?: number | null;
+  hardwareCategory?: string;
+  productCode?: string;
+  requestedQuantity: number;
+}
+
+function lineKey(line: BuilderLine): string {
+  return line.itemType === 'OPENING_ITEM'
+    ? `OI|${line.openingItemId}`
+    : `LOOSE|${line.openingNumber ?? ''}|${line.hardwareCategory}|${line.productCode}`;
```
Xilous, line 47: lineKey puts openingNumber in a LOOSE line's key. every line the inventory picker builds has no openingNumber, so its key is `LOOSE||HINGE|HG-100`. editing a request raised from Start a Task, whose loose lines carry openings, seeds `LOOSE|A01|HINGE|HG-100`.

the two never match -> inventory row reads current 0 and offers Add -> clicking it appends a second line for the same product with a null opening -> save replaces the request with both lines.

key a loose line on (category, product) only, or match the seeded line regardless of opening.

frontend/src/modules/shipping/RequestBuilderDialog.tsx
```diff
@@ -0,0 +1,455 @@
...
+  /**
+   * How much of a product this request may still take. Project availability is already net of every
+   * OTHER request's claim (#342); an edit also has to add back what this request itself is holding,
+   * or trimming a line would read as asking for more.
+   */
+  const headroomFor = useCallback(
+    (row: AvailabilityRow) => {
+      if (!editing) return row.availableQuantity;
+      const held =
+        request?.items.find(
```
Xilous: `.find` returns the first matching item. a request holding two loose lines of the same product, one per opening, which is what Start a Task produces, adds back only the first quantity. availableQuantity is already net of both - headroom under-reports by the rest and the Qty box caps below what the request already holds.

sum every matching item instead of taking the first.

backend/app/repositories/shipping_coverage.py
```diff
@@ -0,0 +1,364 @@
...
+                "hardware_category": key[0],
+                "product_code": key[1],
+                "classification": _dominant_classification(owed_line),
+                "owed_quantity": owed_quantity,
+                "installed_quantity": installed_quantity,
+                # Site hardware is never fitted at the bench, so `installed` is 0 for it and this is
+                # its whole owed quantity. Shop hardware nets off what assembly actually fitted,
+                # leaving exactly what was skipped. Floored at 0: a leaf carrying more than the
+                # current schedule asks for is over-supplied, not owed a negative quantity.
+                "suggested_quantity": max(0, owed_quantity - installed_quantity),
```
Xilous: suggested is owed minus what is bolted onto the leaf, nothing else. it never subtracts what earlier shipping requests already pulled or shipped for this opening.

the list it replaces was clamped to receivedQtyByKey, and reconcile_schedule moves SHIPPING_OUT and SHIPPED_OUT quantities out of the RECEIVED bucket per (opening, product), so the old path stopped offering hardware that had already gone.

opening 0101 owes 2 site locks -> pulled and shipped -> wizard opened on 0101 again -> still says owed 2.

the only remaining bound is project-wide availability - any free stock lets a second set go out and starves the opening that owns it.

net the shipped and in-flight quantities off owed.

backend/app/repositories/shipping_requests.py
```diff
@@ -0,0 +1,408 @@
...
+    Already claimed: one physical leaf sits on exactly one live request. The UI hides claimed
+    leaves, but a stale tab or a non-UI caller must not be able to send the same leaf twice.
+
+    Still in assembly: the same leaf cannot be on its way out the door and inside a live
+    shop-assembly work unit.
+    """
+    if not opening_items_by_id:
+        return
+
+    referenced = list(opening_items_by_id)
```
Xilous, line 321: _guard_leaves only looks at claims that already exist in the database. nothing rejects the same opening_item_id appearing twice inside this call.
- two OPENING_ITEM lines naming one leaf on one request mint two lines that both move it
- two drafts in one finalize each naming leaf X create two live requests holding it, which the already-on-a-live-request guard below exists to prevent

collect the ids per call and refuse a repeat before the database check.

frontend/src/modules/shipping/RequestBuilderDialog.tsx
Xilous: fixed. dialog sends false -> catches the refusal -> confirm dialog -> re-sends true once the user picks "Send them short".

needed a signal to tell that refusal from the two beside it - already-claimed and still-in-assembly both blame opening_item_id, and neither becomes true because someone clicked a button. incomplete-leaves ValidationError now carries field="acknowledge_incomplete_leaves". code stays VALIDATION_ERROR.

2 tests
- confirm-then-resend path
- already-claimed refusal stays a plain error, no "send them short" button

frontend/src/modules/shipping/RequestBuilderDialog.tsx
Xilous: fixed. lineKey keeps the opening - it is what the "on this request" list renders and removes on, and two seeded lines for one product would collide on a React key. inventory picker matches on (category, product) through isLooseFor.

frontend/src/modules/shipping/RequestBuilderDialog.tsx
Xilous: fixed. headroomFor sums every held LOOSE line for the product.

picker Qty box is the same aggregate - current sums across openings, setLooseTotal spreads a change over the lines that carry it
- growth onto the last line
- shrinking off the newest first, so the schedule-attributed lines keep their openings longest

backend/app/repositories/shipping_coverage.py
Xilous: fixed. spoken-for term per (opening, category, product), three sources arranged so nothing is counted twice
- shipped, from PackingSlipItem
- staged, what completed SHIPPING_OUT pulls fulfilled with no slip consuming it. max(shipped, fulfilled) folds these two, since a shipped unit was fulfilled first
- in flight, lines on pulls still being picked plus lines on PENDING requests. a request stops counting at accept, because the accept copies its lines onto a pull the previous term counts

cancelled pulls and rejected requests excluded, so those units come back.

a loose line carries an opening, never a leaf, so this is per opening only. the opening's leaves consume one budget in order rather than each being charged the whole history - a pair owed one lock per leaf that has sent one reads as covered on leaf 1, still owing on leaf 2.

exposed as spokenForQuantity, shown as a chip.

11 tests. query budget 6 -> 9, still fixed regardless of how many openings are selected.

backend/app/repositories/shipping_requests.py
Xilous: fixed. _guard_unique_leaves runs before the database checks on both entry points, refuses a repeated opening_item_id across everything the call is about to write.

4 tests
- twice on one request
- twice across two drafts in one finalize
- twice in an edit
- two different leaves still fine

Issue #453, opened 2026-07-31
Dedup the Delivery Request header field enumeration


Xilous:
deferred from PR #450 review (backend/app/schemas/inputs.py).

20-field Delivery Request header enumerated longhand roughly 17 times across the stack:
migration 077
PackingSlip model
DELIVERY_REQUEST_FIELDS in shipping_repository
ConfirmShipmentInput
UpdateShipmentDetailsInput
PackingSlip GraphQL type
packing_slip_to_type
both SDL reference files
PACKING_SLIP_FIELDS in frontend/src/graphql/shipping.ts
five shapes/helpers in frontend/src/modules/shipping/deliveryRequest.ts

a 21st field drifts silently in two directions, and neither breaks the build or any test:
- missing from DELIVERY_REQUEST_FIELDS, GraphQL accepts it but never persists or clears it
- missing from PACKING_SLIP_FIELDS, saves but reads back undefined, prints blank on the form

collapse the enumeration:
- shared Strawberry input base for the two inputs
- converter iterates DELIVERY_REQUEST_FIELDS instead of naming fields
- one derived frontend field list, PACKING_SLIP_FIELDS and form/DTO shapes generated from it
- parity test asserts model, tuple, inputs, frontend list agree

PR #464, opened 2026-08-04
feat(shipping): shipment method management (#451)

Xilous:
phase 4 of #451. stacked on #463 (phases 1 and 2), review that first. phase 3, staging workspace with containers, still open.

how a load travelled was nowhere on the Delivery Request, and the answer comes back from the same handful of options every time - our truck, a named courier, customer pickup - so free text would spell each of them several ways across a year of shipments.

new `ShipmentMethod` table + CRUD, maintained from the Shipping module, not Admin: the people who pick from the list are the people who maintain it, and a list that needs an admin to extend is a list that gets worked around with free text.

retiring is `isActive = false`, not a delete. a carrier that comes back keeps its spelling, the form offers only active methods, the management screen shows all so retirement is not a one-way door. delete still offered for a row that was simply a mistake.

`packing_slips.shipment_method` is a plain string like `pickup_location` beside it, not a foreign key. a reprint is the copy pulled up in a site dispute, so a method renamed or deleted since must not be able to rewrite or erase what the driver was told - which is also what makes deleting safe at all.

field reaches confirm input -> edit input -> columns -> GraphQL selection -> form state -> printed document by being added to `DELIVERY_REQUEST_FIELDS` on each side, the pair of lists from #453, with `test_delivery_request_field_parity.py` holding them together.

where it shows up:
- Delivery Request form, create and edit, select over active methods; free text when no methods exist yet, so a company that has not set any up can still book
- slip booked under a since-retired method keeps showing it in the edit form rather than silently blanking
- SHIPMENT METHOD on the printed Delivery Request
- Method column on the Shipments page
- truck icon in the Shipping module header opens the management dialog

all four fields SIGNED_IN. one migration, 082, additive.

`backend/tests/test_shipment_methods.py` covers
dropdown ordering
case-insensitive uniqueness
blank and trimmed names
retire and reactivate
renaming onto another name
renaming to itself
partial updates
rename not touching a shipment already sent under the old name
delete leaving that shipment readable

frontend `deliveryDetailsInput` test counts the header off the shared field list instead of the literal 20, so adding a field reads as the addition it is rather than as a regression.

simulated user testing not yet run against a PR environment. once one is up:

1. Shipping -> truck icon in header -> add "Flatbed" and "Our truck"
2. retire "Our truck" -> keeps a Retired chip on the management screen
3. ship something -> Delivery Request's Shipment method select offers Flatbed, not Our truck
4. confirm -> Shipments page Method column and printed Delivery Request both read Flatbed
5. delete Flatbed from the management screen -> reprint that shipment's Delivery Request -> still says Flatbed

### DIFF
--- a/frontend/src/modules/shipping/ShipmentMethodsDialog.tsx
+++ b/frontend/src/modules/shipping/ShipmentMethodsDialog.tsx
@@ -1,0 +1,198 @@
+import { useState } from 'react';
+import {
+  Alert,
+  Box,
+  Button,
+  Chip,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogTitle,
+  IconButton,
+  Stack,
+  TextField,
+  Typography,
+} from '@mui/material';
+import { Plus, Trash2 } from 'lucide-react';
+import { useMutation, useQuery } from '@apollo/client/react';
+import {
+  CREATE_SHIPMENT_METHOD,
+  DELETE_SHIPMENT_METHOD,
+  GET_SHIPMENT_METHODS,
+  UPDATE_SHIPMENT_METHOD,
+} from '../../graphql/shipping';
+import ConfirmDialog from '../../components/ConfirmDialog';
+import { useToast } from '../../components/Toast';
+import { microLabelSx, monoSx } from '../../theme';
+import type { ShipmentMethod } from './useShipmentMethods';
+
+interface Props {
+  open: boolean;
+  onClose: () => void;
+}
+
+/**
+ * The shipping department's own list of how a load can travel (#451).
+ *
+ * Kept here rather than in Admin because the people who maintain it are the people who pick from it
+ * on the Delivery Request, and a list that needs an admin to extend is a list that gets worked
+ * around with free text.
+ *
+ * Retiring is the ordinary action and deleting is the exception: a carrier that comes back should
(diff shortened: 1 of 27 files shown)


frontend/src/modules/shipping/ShipmentMethodsDialog.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+            <Stack direction="row" spacing={1}>
+              <TextField
+                size="small"
+                label="New method"
+                value={newName}
+                onChange={(e) => setNewName(e.target.value)}
+                onKeyDown={(e) => {
+                  if (e.key === 'Enter' && newName.trim()) {
+                    createMethod({ variables: { name: newName.trim(), sortOrder: methods.length } });
```
Xilous: sortOrder is taken from methods.length, which counts retired methods too and reuses an existing value after any delete. dropdown order is not stable.

the Enter handler does not check `creating`, unlike the Add button beside it - a fast double-Enter fires two createShipmentMethod calls and the second comes back as a name conflict.

Xilous: both fixed. sortOrder comes off the highest sortOrder in use, retired rows included since they hold their place for reactivation. Enter handler and Add button go through one function carrying the in-flight guard. 4 tests.

PR #465, opened 2026-08-04
feat(shipping): staging workspace and shipment containers (#451)

Xilous:
phase 3 of #451, last one. stacked on #464 (phase 4), which is stacked on #463 (phases 1 and 2) - review those first.

shipping out went straight from "these units are ship-ready" to a packing slip. the work in between - stacking a skid in an order the site can unload, building a shipment up over days out of several containers - happened on the floor and was recorded nowhere, so nobody could see what was already loaded without walking to it.

new `ShipmentContainer` and `ShipmentContainerItem`, named by whoever loads it because the label goes on the physical thing in marker. a container is one of:
skid
door cart
box
envelope
bundle

`packingSlipId` carries the whole lifecycle, no status column - the only question ever asked is "has this left":
- null while being built
- stamped when its shipment is confirmed
- history from then on

placement enforces three ceilings, all about physical objects rather than policy:
- 30 leaves to a skid, because a taller stack cannot be strapped
- one leaf in one open container, because there is one of it
- loose placements bounded by what is staged and unplaced, counting what other open containers hold

re-saving a container is measured with its own units added back, so saving unchanged or trimming is never refused for what it already holds - the same release-before-measure trap the request edit had in phase 2.

`setContainerItems` is one batch call, not place + remove + reorder as three: a drag-and-drop surface already holds the whole list, and saving it in pieces leaves a moment where the stack is in an order nobody chose. list index IS the position, position 0 loaded first -> bottom of a skid.

`confirmShipmentFromContainers` is a thin wrapper over `confirmShipment` rather than a second confirm path, because that function still owns:
quarantine gate
slip-number uniqueness check
SHIP_READY transition
loose arithmetic

wrapper adds only where items come from, plus stamping the slip onto the containers afterwards.

staging pool is one query, not two. left column reads `unplaced*`, container cards read `containers`; two queries could disagree about whether a leaf has been loaded, which is the exact mistake containers exist to prevent.

new Staging tab in the Shipping module. drag to place and to restack, with an equal control beside every item - a "Place in" menu on each pool row, up/down buttons in each stack. the buttons are not a fallback bolted on for compliance: this screen is worked on a tablet in a warehouse, and a drag that misses puts a door leaf on the wrong skid, where the undo is walking to the rack.

a skid shows n/30 and refuses the 31st leaf client-side with a toast as well as server-side. confirm dialog prints each container's stack in order, so the driver and the site read the same thing.

cart untouched on the Ship tab, retired in the follow-up that moves its last consumer over - shipping has to keep working while both exist.

`stagingPool` and the five container mutations are all SIGNED_IN. one migration, 083, additive, reuses the `pull_request_item_type` enum rather than creating a second one.

`backend/tests/test_shipment_containers.py` covers:
open-name uniqueness
blank name
breaking down returns contents to the pool
unstaged leaf refused
one leaf in two containers refused
same leaf twice in one container refused
skid 30 cap
door cart has no cap
loose over-placement refused
another container's placement counting against what is free
re-saving not gated against its own units
position from list order
reordering
emptying allowed
shipping stamps the slip, closes the containers, flips the leaf to SHIPPED_OUT
empty container refused
shipping nothing refused
shipped container refuses edit/delete/rename
container from another project refused

simulated user testing not yet run against a PR environment. once one is up, with a project that has staged hardware:

1. Shipping -> Staging -> add "Skid 1" (Skid) and "Box 1" (Box)
2. drag a door leaf onto Skid 1, use the Place in menu for a second - both leave the pool
3. reorder the skid by dragging, then with the up/down buttons - numbers renumber
4. put loose hardware in Box 1 - pool row's unplaced count drops
5. tick both containers, Ship - dialog lists both, skid in stack order
6. confirm, check Shipments shows the slip and Staging no longer lists either container
7. place a leaf already on Skid 1 into Box 1 - refused, naming the container holding it

### DIFF
--- a/backend/app/auth_policy.py
+++ b/backend/app/auth_policy.py
@@ -169,6 +169,14 @@ ROOT_FIELD_POLICY: dict[str, str | frozenset[str]] = {
     "returnableLines": SIGNED_IN,
     # The shipping department's own list of how a load travels (#451). SIGNED_IN like the rest of
     # shipping: it is documentation the same people maintain and consume, and it gates nothing.
+    # The staging workspace and its containers (#451). SIGNED_IN like the rest of shipping - the
+    # warehouse and the shipping department both load a truck, and neither owns the screen.
+    "stagingPool": SIGNED_IN,
+    "createShipmentContainer": SIGNED_IN,
+    "renameShipmentContainer": SIGNED_IN,
+    "deleteShipmentContainer": SIGNED_IN,
+    "setContainerItems": SIGNED_IN,
+    "confirmShipmentFromContainers": SIGNED_IN,
     "shipmentMethods": SIGNED_IN,
     "createShipmentMethod": SIGNED_IN,
     "updateShipmentMethod": SIGNED_IN,
--- a/backend/app/auth_policy.py
+++ b/backend/app/auth_policy.py
@@ -167,8 +167,6 @@ ROOT_FIELD_POLICY: dict[str, str | frozenset[str]] = {
     # --- shipping.py ----------------------------------------------------------------------
     "packingSlips": SIGNED_IN,
     "returnableLines": SIGNED_IN,
-    # The shipping department's own list of how a load travels (#451). SIGNED_IN like the rest of
-    # shipping: it is documentation the same people maintain and consume, and it gates nothing.
     # The staging workspace and its containers (#451). SIGNED_IN like the rest of shipping - the
     # warehouse and the shipping department both load a truck, and neither owns the screen.
     "stagingPool": SIGNED_IN,
@@ -177,6 +175,8 @@ ROOT_FIELD_POLICY: dict[str, str | frozenset[str]] = {
     "deleteShipmentContainer": SIGNED_IN,
     "setContainerItems": SIGNED_IN,
     "confirmShipmentFromContainers": SIGNED_IN,
+    # The shipping department's own list of how a load travels (#451). SIGNED_IN like the rest of
+    # shipping: it is documentation the same people maintain and consume, and it gates nothing.
     "shipmentMethods": SIGNED_IN,
     "createShipmentMethod": SIGNED_IN,
     "updateShipmentMethod": SIGNED_IN,
--- a/backend/app/repositories/shipment_containers.py
+++ b/backend/app/repositories/shipment_containers.py
@@ -32,6 +32,21 @@ from app.models.shipment_container import (
 STACKED_TYPES = (ShipmentContainerType.SKID, ShipmentContainerType.DOOR_CART)
 
 
+def loose_key(opening_number: str | None, hardware_category: str, product_code: str) -> tuple[str | None, str, str]:
+    """How a staged loose quantity is identified, everywhere on this path.
+
+    The opening is part of it because `get_ship_ready_items` groups the staged pool that way and
+    `confirm_shipment` checks availability that way. Keying containers on the product alone reads as
+    harmless right up until two openings stage the same product: the pool would report one of the two
+    quantities as if it were the total, and a placement the staging rules accepted would be refused
+    at confirm - or booked against an opening it was never pulled for.
+
+    Null is a real value here, not a missing one. A pull raised straight off inventory has no opening
+    to attribute (#451), and those units are their own bucket rather than everyone's.
+    """
+    return (opening_number, hardware_category, product_code)
+
+
 def get_containers(session: Session, project_id: uuid.UUID, *, open_only: bool = True) -> list[ShipmentContainer]:
     """Containers for one project, items eagerly loaded (the type builder walks them).
 
@@ -115,7 +130,7 @@ def set_container_items(
     staged_pool = build_staged_pool(session, container.project_id)
 
     leaf_ids: set[uuid.UUID] = set()
-    loose_wanted: dict[tuple[str, str], int] = {}
+    loose_wanted: dict[tuple[str | None, str, str], int] = {}
     for item in items:
         item_type = PullRequestItemType(item["item_type"])
         if item_type == PullRequestItemType.OPENING_ITEM:
@@ -129,7 +144,7 @@ def set_container_items(
                 raise ValidationError("The same door leaf cannot be placed twice in one container.", field="items")
             leaf_ids.add(oi_id)
         else:
-            key = (item["hardware_category"], item["product_code"])
+            key = loose_key(item.get("opening_number"), item["hardware_category"], item["product_code"])
             loose_wanted[key] = loose_wanted.get(key, 0) + int(item.get("quantity", 1))
 
     if container.container_type is ShipmentContainerType.SKID and len(leaf_ids) > MAX_LEAVES_PER_SKID:
@@ -195,7 +210,10 @@ def confirm_shipment_from_containers(
     if not container_ids:
         raise ValidationError("Pick at least one container to ship.", field="containerIds")
 
-    containers = [_open_container(session, cid) for cid in container_ids]
+    # Deduplicated before loading. The same id twice builds the item list twice, which doubles every
+    # loose quantity on the slip and makes `confirm_shipment`'s leaf count disagree with the number
+    # of distinct leaves it found - reported as the leaf not existing.
+    containers = [_open_container(session, cid) for cid in dict.fromkeys(container_ids)]
     for container in containers:
         if container.project_id != project_id:
             raise ValidationError(f"{container.name} belongs to another project.", field="containerIds")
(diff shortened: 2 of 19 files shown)


backend/app/repositories/shipment_containers.py
```diff
@@ -0,0 +1,353 @@
...
+    The staged side comes from `get_ship_ready_items`, which is the existing definition of what is
+    out of inventory and not yet shipped; this only adds where it has been put since.
+    """
+    from app.repositories import shipping_repository
+
+    ready = shipping_repository.get_ship_ready_items(session, project_id)
+    pool = {
+        "leaves": {oi.id: None for oi in ready["opening_items"]},
+        "loose": {
+            (li["hardware_category"], li["product_code"]): {"staged": li["available_quantity"], "placed": 0}
```
Xilous: pool collapses loose by (hardware_category, product_code), get_ship_ready_items groups loose_items by (opening_number, hardware_category, product_code). two openings staging the same product produce two rows, the last one survives, `staged` becomes that row's quantity instead of the sum.

opening 101 staged 4 HG-100 -> opening 102 staged 3 -> pool reports staged 3 -> placing 5 in a box refused with "only 3 staged and unplaced", 7 on the floor.

sum instead of overwrite, or key the pool the way get_ship_ready_items keys its rows.

backend/app/schemas/shipping.py
```diff
@@ -142,6 +194,55 @@ def shipping_coverage(
...
+                        floor=by_id[oi_id].floor,
+                        location=by_id[oi_id].location,
+                        placed_in_container_id=strawberry.ID(str(holder)) if holder else None,
+                    )
+                    for oi_id, holder in pool["leaves"].items()
+                    if oi_id in by_id
+                ],
+                loose_items=[
+                    StagedLooseItem(
+                        opening_number=next(
```
Xilous: StagedLooseItem takes its opening_number from the first ready loose row matching (category, product). two openings staging the same product -> one workspace row wearing an arbitrary opening number -> placeLoose writes that value onto the container item.

confirm_shipment checks loose availability on (opening_number, hardware_category, product_code). a container the staging rules accepted is refused at confirm with "Insufficient loose hardware ... for opening 101", or units book against an opening they were never pulled for.

give the container path the same three-part key the confirm uses.

frontend/src/modules/shipping/StagingWorkspace.tsx
```diff
@@ -0,0 +1,727 @@
...
+      if (!over) return;
+      const activeId = String(active.id);
+      const overId = String(over.id);
+
+      // Reorder inside one container: both ends are items of the same stack.
+      const owner = containers.find((c) => c.items.some((i) => i.id === activeId));
+      if (owner) {
+        const from = owner.items.findIndex((i) => i.id === activeId);
+        const to = owner.items.findIndex((i) => i.id === overId);
+        if (to >= 0 && from !== to) save(owner, arrayMove(owner.items, from, to));
```
Xilous: the handler always returns once `owner` is found. a drag that starts on a container item can only reorder inside that same container.

drag a leaf from Skid 1 onto Skid 2, or onto Skid 2's card, and `to` stays -1, nothing is saved, the row snaps back with no message. moving a leaf between skids is the main thing this screen exists for, and remove-then-place is the only way left to do it.

when `overId` resolves to a different container, remove from `owner` and append to that container.

frontend/src/modules/shipping/StagingWorkspace.tsx
```diff
@@ -0,0 +1,727 @@
...
+            ...container.items,
+            {
+              id: 'new',
+              itemType: 'LOOSE' as const,
+              openingItemId: null,
+              openingNumber: item.openingNumber,
+              leaf: null,
+              hardwareCategory: item.hardwareCategory,
+              productCode: item.productCode,
+              quantity: item.unplacedQuantity,
```
Xilous: placeLoose always takes the whole `unplacedQuantity`, and the pool row is filtered out once unplaced hits 0. a product can never be split across two containers - 4 hinges cannot go 2 in Box 1 and 2 in Box 2.

no quantity control on a container row either, the only correction is removing the line and starting over.

backend already takes a per-placement quantity. add a number beside Place in.

backend/app/schemas/shipping.py
```diff
@@ -142,6 +194,55 @@ def shipping_coverage(
...
+
+        One query for both halves of the workspace - what is still loose on the floor, and what is
+        already in a container - because two would be able to disagree about whether a leaf has been
+        loaded, and that disagreement is exactly the mistake containers exist to prevent.
+        """
+        pid = uuid.UUID(str(project_id))
+        with SessionLocal() as session:
+            pool = shipment_containers.build_staged_pool(session, pid)
+            containers = shipment_containers.get_containers(session, pid, open_only=True)
+            ready = shipping_repository.get_ship_ready_items(session, pid)
```
Xilous, line 211: build_staged_pool already runs get_ship_ready_items and get_containers internally, and both are run again here. the workspace's main read issues each of them twice, and get_ship_ready_items is three queries plus a selectinload.

have build_staged_pool return the ready set and the containers it already loaded, or take them as arguments.

backend/app/repositories/shipment_containers.py
```diff
@@ -0,0 +1,353 @@
...
+
+    All this adds is where the items come from and, afterwards, stamping the slip onto the
+    containers so they read as shipped instead of staying open and re-shippable.
+    """
+    from app.repositories import shipping_repository
+
+    if not container_ids:
+        raise ValidationError("Pick at least one container to ship.", field="containerIds")
+
+    containers = [_open_container(session, cid) for cid in container_ids]
```
Xilous: container_ids is not deduplicated. the same id twice builds the item list twice - loose quantities double on the slip, and for leaves confirm_shipment's `len(locked_ois) != len(oi_ids)` check fires and reports the leaf as not found.

run dict.fromkeys over container_ids before loading.

backend/app/auth_policy.py
```diff
@@ -169,6 +169,14 @@
     "returnableLines": SIGNED_IN,
     # The shipping department's own list of how a load travels (#451). SIGNED_IN like the rest of
     # shipping: it is documentation the same people maintain and consume, and it gates nothing.
+    # The staging workspace and its containers (#451). SIGNED_IN like the rest of shipping - the
```
Xilous, line 170: the comment block above belongs to the shipmentMethods entries, and these staging entries were inserted between it and them. it now reads as the header for stagingPool.

move the staging comment above it, or the staging entries below the method entries.

backend/app/schemas/shipping.py
```diff
@@ -142,6 +194,55 @@ def shipping_coverage(
                 for leaf in leaves
             ]
 
+    @strawberry.field
+    def staging_pool(self, info: strawberry.Info, project_id: strawberry.ID) -> StagingPool:
```
Xilous, line 199: backend/app/graphql/shipping.graphql was updated for the coverage builder in #463 and for shipment methods in #464, nothing in this PR reaches it. the file is hand-maintained and calls itself the shipping module's SDL reference, so nothing catches the drift. missing:
stagingPool
the five container mutations
ShipmentContainer
ShipmentContainerItem
StagingPool
StagedLeaf
StagedLooseItem
ShipmentContainerType
SetContainerItemsInput
ContainerItemInput
ConfirmShipmentFromContainersInput

backend/app/repositories/shipment_containers.py
Xilous: fixed, with the StagedLooseItem half below - same bug.

one loose_key helper, (opening, category, product), used by
- the pool
- the placement check
- the container rows

matches what get_ship_ready_items groups by and what confirm_shipment checks. staged quantities sum rather than overwrite. null opening is a real bucket for stock pulled off inventory, not a catch-all.

backend/app/schemas/shipping.py
Xilous: fixed by the same change. opening comes off the pool key - a product staged for two openings is two workspace rows. frontend keeps them apart through sameLooseStock rather than merging on product. 6 backend tests.

frontend/src/modules/shipping/StagingWorkspace.tsx
Xilous: fixed. handler resolves the drop target first
- same container -> restack
- different container -> remove from source -> append to target

source save is awaited before the target write, or the add is refused for a leaf still recorded in the skid it is leaving. a skid at 30 leaves refuses with a message rather than silently.

frontend/src/modules/shipping/StagingWorkspace.tsx
Xilous: fixed. quantity sits beside Place in, defaults to everything unplaced and capped there. drag and menu both read it, held in the parent so the two cannot disagree. container lines carry an editable quantity, zero takes the line out.

backend/app/schemas/shipping.py
Xilous: fixed. takes ready and containers as optional arguments, resolver passes what it already read.

backend/app/repositories/shipment_containers.py
Xilous: fixed with dict.fromkeys, plus a test that the same id twice ships once.

backend/app/auth_policy.py
Xilous: fixed, moved above them.

backend/app/schemas/shipping.py
Xilous: fixed. added
stagingPool
the five container mutations
ShipmentContainerType
ShipmentContainer
ShipmentContainerItem
StagingPool
StagedLeaf
StagedLooseItem
SetContainerItemsInput
ContainerItemInput
ConfirmShipmentFromContainersInput

checked field by field against the resolvers, not from the plan - nullability and defaults differed in three places.

PR #466, opened 2026-08-04
ci: verification run for the #451 stack (do not merge)

Xilous:
Throwaway draft so CI runs over the whole #451 stack - the workflow only triggers on PRs targeting master, so #464 and #465 get no backend run on their own. Closing once it is green; review happens on #463 -> #464 -> #465.


Xilous:
Closing - it did its job. Green over the whole stack (1712 passed), after catching migration 083's duplicate CREATE TYPE and a delivery-request fixture that was one field short. Both fixed on the branches that introduced them. Review continues on #463 -> #464 -> #465.

PR #473, opened 2026-08-04
feat(inventory): entity types and custom attributes for non-schedule stock (#454)

### LINKED ISSUE #454: inventory entity types
openings <- door <- hardware

frames

specialties

consumables

---

for openings > door leaves > hardware (shop or site) these have their own attributes

for frames, specialties, and consumables, technically these are just known as "other" inventory types.

but nonetheless, these custom items must go through the Nexus/business pipeline. they must be somehow purchased, received, stored in inventory, and shipped out

these "other" items must also have what we will refer to as "custom attributes" that warehouse users can create. so frames will have their own custom attributes, apart from specialties' custom attributes, apart from consumables'.

and the warehouse user can write down whatever value they want for each of these custom attributes for each item that belongs to that category of inventory item.

Xilous:
closes #454.

frames, specialties, consumables are inventory the TITAN schedule never describes. each type carries its own custom attributes, defined by warehouse staff, free-text values.

the type rides in `hardware_category`.

every pipeline table keys an item on the free-text pair `(hardware_category, product_code)`:
po_line_items
receive_draft_line_items
inventory_locations
stock_items
inventory_reservations
pull lines
shipping lines

an inventory item type owns an immutable uppercase code (FRAME, SPECIALTY, CONSUMABLE). a PO line writes that code into `hardware_category` -> receiving -> put-away -> picking -> shipping, untouched. zero changes to existing pipeline tables, mutations or fungibility keys. a frame stays as fungible as a hinge. the code is immutable, the name is renameable.

backend
- 4 new tables, migration 084
`inventory_item_types`
`inventory_item_attributes`
`custom_inventory_items`
`custom_inventory_item_values`
- migration seeds FRAME/Frames, SPECIALTY/Specialties, CONSUMABLE/Consumables. attributes are not seeded
- a code already in use as a hardware category on existing stock or POs is refused. hardware categories come off the TITAN schedule and are registered nowhere, so the only evidence HINGE is taken is that something was bought under it
- nothing is deleted. types, attributes and items retire to `is_active = False`
- attribute values keyed by attribute id, not name. renames carry every value, retirement keeps them
- an attribute belonging to another type is refused on an item
- 8 root fields, all SIGNED_IN in `auth_policy.py`

frontend
- Warehouse -> Custom Items
type list
per-type attribute editor
item catalog grid, one column per active attribute
- Create PO gains an "Add Custom Item" picker beside "Add Item". a picked row locks its hardware category and product code: they are the pair the warehouse receives the stock under, and a hand-edit produces inventory whose catalog entry cannot be found. quantity and unit cost stay the buyer's, `orderAs` seeds from the description
- Inventory and Stock Pool show a type's display name with its code as a chip, and a catalogued item's description and attribute values under its product code
- receiving, put-away, pull, pick and shipping screens unchanged

tests
`backend/tests/test_custom_items.py`, 24 tests. `test_a_frame_travels_the_hardware_pipeline_untouched` orders a frame, receives it, claims it on a shipping-out request, through the hinge code paths with no branch that knows what a frame is.

not in this PR: attribute value typing (numbers, dates, dropdowns), per-unit values, TITAN import of other types, shop assembly involvement, attribute values on pick sheets and packing slips.


backend/app/schemas/custom_items.py
```diff
@@ -0,0 +1,291 @@
...
+            return _item_type_to_type(custom_items_repository.get_item_type(session, type_id))
+
+    @strawberry.mutation
+    def create_custom_inventory_item(
+        self,
+        info: strawberry.Info,
+        type_id: strawberry.ID,
+        product_code: str,
+        description: str | None = None,
+        values: list[CustomInventoryItemValueInput] = strawberry.field(default_factory=list),
```
Xilous: `values` argument default is a `StrawberryField`, not an empty list.

`strawberry.field(default_factory=list)` is for input-type fields, not resolver argument defaults. every other use of it in this repo is on an `@strawberry.input` attribute. used as a parameter default it does two things.

- SDL renders `createCustomInventoryItem(..., values: [CustomInventoryItemValueInput!]!)`, required and non-null with no default
- omit `values` and graphql-core passes the raw `StrawberryField` object through as the argument value, `_values_input(values)` raises `TypeError: 'StrawberryField' object is not iterable` -> 500

verified against the schema built from this branch. current frontend always sends `values`, so it is latent.

`values: list[CustomInventoryItemValueInput] | None = None`, then `values=_values_input(values or [])`, matching `update_custom_inventory_item` directly below.

backend/app/repositories/custom_items_repository.py
```diff
@@ -0,0 +1,420 @@
...
+) -> list[CustomInventoryItem]:
+    """The catalog, filtered the way the two callers need it.
+
+    The PO dialog's picker passes a type and `active_only`; the management screen passes a type and
+    nothing else. Both walk `.values` and `.item_type` per row, so both are eagerly loaded.
+    """
+    stmt = (
+        select(CustomInventoryItem)
+        .options(
+            selectinload(CustomInventoryItem.values),
```
Xilous: N+1, `_item_to_type` walks `v.attribute.name` and nothing eager-loads it.

`selectinload(CustomInventoryItem.values)` loads the value rows, not `CustomInventoryItemValue.attribute`. `_item_to_type` reads `v.attribute.name` for every value, so `customInventoryItems` issues one extra SELECT per value row. HardwareItemsTab calls this unfiltered: 200 items with 5 attributes each is 1000 extra round trips over Railway's hop to managed Postgres.

CLAUDE.md: "Every relationship a `_xxx_to_type` helper iterates must be `selectinload`-ed by the calling resolver."

`get_item` at line 235 needs the same fix.

```suggestion
            selectinload(CustomInventoryItem.values).selectinload(CustomInventoryItemValue.attribute),
```

backend/alembic/versions/084_inventory_item_types.py
```diff
@@ -0,0 +1,122 @@
...
+        sa.Column("created_at", sa.DateTime(), nullable=False),
+        sa.Column("updated_at", sa.DateTime(), nullable=False),
+        sa.ForeignKeyConstraint(["item_id"], ["custom_inventory_items.id"], ondelete="CASCADE"),
+        sa.ForeignKeyConstraint(["attribute_id"], ["inventory_item_attributes.id"], ondelete="CASCADE"),
+        sa.UniqueConstraint("item_id", "attribute_id", name="uq_custom_inventory_item_values_item_attr"),
+    )
+    op.create_index("ix_custom_inventory_item_values_item_id", "custom_inventory_item_values", ["item_id"])
+
+    now = datetime.utcnow()
+    op.bulk_insert(
```
Xilous: seed writes FRAME/SPECIALTY/CONSUMABLE past the code-in-use check.

`_check_code_free` refuses a type whose code already appears as a `hardware_category` on a PO line, inventory location or stock item, because such a type "would show every hinge in the building as an uncatalogued item of that type". this bulk_insert bypasses that guard.

on a database already carrying `hardware_category = 'FRAME'` anywhere in the pipeline, the seeded type adopts it.

- HardwareItemsTab relabels the group "Frames" with a FRAME chip
- StockPoolView does the same in the Category column
- code is permanently claimed with no way to change it

either query the three tables in the migration and skip a code already in use, or confirm against production that none of the three codes appears in `po_line_items` / `inventory_locations` / `stock_items` / `hardware_items` before merging.

backend/app/repositories/custom_items_repository.py
```diff
@@ -0,0 +1,420 @@
...
+    never registered anywhere, so the only way to know HINGE is taken is to look at what has been
+    bought and received. A type created on a code already carrying hardware would show every hinge in
+    the building as an uncatalogued item of that type, which reads as data corruption to the person
+    looking at the screen.
+    """
+    existing = session.scalars(select(InventoryItemType).where(InventoryItemType.code == code)).first()
+    if existing is not None:
+        raise ConflictError(f"The code {code} is already used by the type {existing.name}", field="code")
+
+    in_use = session.execute(
```
Xilous, line 436: `_check_code_free` does not look at `hardware_items`.

`hardware_items` is where a category first appears, before anything is bought. a schedule imported with a `GASKETS` category and no PO raised yet leaves `GASKETS` looking free, a type claims it, and every gasket purchased afterwards lands in that type's group.

`HardwareItem.hardware_category` is the column (backend/app/models/hardware.py:29). a fourth EXISTS closes it.

frontend/src/modules/warehouse/HardwareItemsTab.tsx
```diff
@@ -781,30 +825,40 @@ export default function HardwareItemsTab({ projectId }: HardwareItemsTabProps) {
...
-                totalValue={pc.totalValue}
+      {filteredGroups.map((group) => {
+        // Only in category mode: a vendor group's label is a vendor name, which will never be a
+        // type code, and matching it against one would be a coincidence rather than a fact.
+        const itemType = groupBy === 'category' ? typesByCode.get(group.label) : undefined;
+        return (
+          <Accordion key={group.label} disableGutters elevation={0} sx={ACCORDION_SX}>
+            <AccordionSummary expandIcon={<ChevronDown size={18} strokeWidth={1.75} />}>
+              <GroupSummary
+                name={itemType ? itemType.name : group.label}
```
Xilous, line 847: search matches the raw code while the group displays the type name.

`GroupSummary` now renders `itemType.name` ("Frames"), `filteredGroups` (line 667) still tests `group.label.toLowerCase().includes(search)`, and `group.label` is the raw category ("FRAME"). typing "Frames" into the search box drops the group visibly labelled "Frames", it survives only if one of its product codes happens to match.

the category dropdown has the same split but is correct, its `value` stays the code.

match the resolved display name too, e.g. `(typesByCode.get(group.label)?.name ?? group.label).toLowerCase().includes(search)`.

frontend/src/modules/warehouse/CustomItemsPage.tsx
```diff
@@ -0,0 +1,675 @@
...
+    CREATE_INVENTORY_ITEM_TYPE,
+    {
+      onCompleted: (data) => onCreated(data.createInventoryItemType),
+      onError: (e) => setError(e.message),
+    },
+  );
+
+  // Mirrors the backend's `_normalize_code`. Shown, not sent: the server derives it again from the
+  // name, so the preview can never disagree with what is stored.
+  const previewCode = useMemo(
```
Xilous, line 482: code preview and backend normalizer disagree on non-ASCII.

the comment claims the preview "can never disagree with what is stored". `_normalize_code` keeps any character where `str.isalnum()` is true, which includes accented letters and non-Latin scripts, the JS `[^A-Z0-9]` replaces them with `_`.

name "Größe" -> dialog previews `GR_SSE`, server stores `GRÖSSE`. the code is immutable, so the user is left holding a code they were shown a different value for.

make the backend ASCII-only, `c if c.isascii() and c.isalnum() else "_"`.

frontend/src/modules/warehouse/CustomItemsPage.tsx
```diff
@@ -0,0 +1,675 @@
...
+              key={attribute.id}
+              label={attribute.name}
+              variant={attribute.isActive ? 'filled' : 'outlined'}
+              size="small"
+              onDelete={
+                attribute.isActive
+                  ? () => updateAttribute({ variables: { id: attribute.id, isActive: false } })
+                  : undefined
+              }
+              deleteIcon={
```
Xilous, line 374: the retire action wears the reactivate icon.

on an active attribute chip `onDelete` retires it and `deleteIcon` is `<RotateCcw />`.
on a retired chip the `icon` slot is also `<RotateCcw />`, and that one reactivates.

the same glyph means two opposite things one chip apart, and the one that reads most like undo is the destructive one. use `X` or `EyeOff` on the active chip.

frontend/src/modules/warehouse/HardwareItemsTab.tsx
```diff
@@ -550,6 +587,13 @@ export default function HardwareItemsTab({ projectId }: HardwareItemsTabProps) {
   const [debouncedSearch, setDebouncedSearch] = useState('');
   const [groupBy, setGroupBy] = useState<GroupBy>('category');
 
+  // Non-schedule stock - frames, specialties, consumables (#454) - carries its type's code as its
+  // hardware category, so a category group whose label is a known type code IS that type, and the
+  // catalog can describe every product code under it. Both queries degrade to empty: without them
+  // this page renders exactly what it rendered before the catalog existed.
+  const { byCode: typesByCode } = useInventoryItemTypes();
+  const { byKey: catalogByKey } = useCustomInventoryItems();
```
Xilous, line 595: whole catalog is fetched on every mount of the inventory tab.

`useCustomInventoryItems()` with no arguments sends `typeId: null, activeOnly: false`, pulling every catalogued item of every type, retired ones included, with all their attribute values, only to build a lookup map. `fetchPolicy: 'cache-and-network'` re-requests it on every mount. with the missing `selectinload` on `values.attribute` that is one query plus one per value row, on a page that used to issue a single hierarchy query.

- pass `activeOnly: true` at minimum
- better, pass `typeId` for the categories actually on screen, or move the join server-side into `inventoryHierarchy`

backend/app/auth_policy.py
```diff
@@ -84,6 +84,19 @@
...
     "deleteBuyerAssignment": ADMIN_ROLE,
     "saveBuyerAssignment": ADMIN_ROLE,
+    # --- custom_items.py ------------------------------------------------------------------
+    # The catalog of non-schedule inventory - frames, specialties, consumables (#454). SIGNED_IN
+    # throughout: the issue puts maintenance in warehouse users' hands and there is no Clerk role for
+    # a plain warehouse user, the reads are consumed by the PO dialog and the inventory screens as
+    # well as by the management page, and nothing here moves stock or reaches GP. It is description.
+    "inventoryItemTypes": SIGNED_IN,
+    "customInventoryItems": SIGNED_IN,
+    "createInventoryItemType": SIGNED_IN,
```
Xilous, line 97: every signed-in caller can create and retire types.

the six catalog mutations are all `SIGNED_IN`. a type's code is stamped into `hardware_category` on every row its items produce, is immutable, and has no delete path, so `createInventoryItemType` is the most permanent write on this PR. any authenticated account can make it, and can retire the three seeded types out from under the pickers.

reads have to stay `SIGNED_IN`, the PO dialog consumes them. the mutations' only frontend caller is `CustomItemsPage` in the warehouse module, and `approveReceiveDraft` / `rejectReceiveDraft` already gate comparable warehouse writes on `frozenset({ADMIN_ROLE, WAREHOUSE_MANAGER_ROLE})`.

backend/app/repositories/custom_items_repository.py
```diff
@@ -0,0 +1,420 @@
...
+    item = get_item(session, item_id)
+
+    if description is not None:
+        item.description = description.strip() or None
+    if is_active is not None:
+        item.is_active = is_active
+    if values is not None:
+        _apply_values(session, item, values)
+    session.flush()
+    session.refresh(item)
```
Xilous: `session.refresh(item)` throws away a SELECT.

`_apply_values` already flushed, and every caller of `update_item` re-reads through `get_item` after the commit (custom_items.py:291), so this emits a query whose result is discarded a moment later. drop it, behaviour is unchanged.

frontend/src/modules/warehouse/CustomItemsPage.tsx
```diff
@@ -0,0 +1,675 @@
...
+  const submitAttribute = () => {
+    const name = newAttribute.trim();
+    if (!name || creatingAttribute) return;
+    const sortOrder = type.attributes.reduce((highest, a) => Math.max(highest, a.sortOrder + 1), 0);
+    createAttribute({ variables: { typeId: type.id, name, sortOrder } });
+  };
+
+  const activeAttributes = type.attributes.filter((a) => a.isActive);
+
+  const columns = useMemo<GridColDef[]>(() => {
```
Xilous, line 264: the columns memo never hits.

`activeAttributes` is `type.attributes.filter(...)` computed inline on line 260, so it is a new array every render, `useMemo(..., [activeAttributes])` rebuilds `columns` every render, and DataGrid gets a fresh `columns` identity each time.

wrap `activeAttributes` in `useMemo` (ItemForm already does exactly that at line 566) or depend on `type.attributes` directly.

frontend/src/hooks/useCustomItems.ts
```diff
@@ -0,0 +1,100 @@
...
+    GET_CUSTOM_INVENTORY_ITEMS,
+    {
+      variables: { typeId: options?.typeId ?? null, activeOnly: options?.activeOnly ?? false },
+      skip: options?.skip,
+      fetchPolicy: 'cache-and-network',
+    },
+  );
+  const items = useMemo(() => data?.customInventoryItems ?? [], [data]);
+  const byKey = useMemo(
+    () => new Map(items.map((i) => [`${i.hardwareCategory}::${i.productCode}`, i])),
```
Xilous: `byKey` re-declares the key format that `catalogKey` exists to own. same template literal, ten lines apart.

```suggestion
    () => new Map(items.map((i) => [catalogKey(i.hardwareCategory, i.productCode), i])),
```

`catalogKey` has to be declared above this hook for that to work.

backend/app/repositories/custom_items_repository.py
```diff
@@ -0,0 +1,420 @@
...
+            selectinload(CustomInventoryItem.item_type),
+        )
+        .order_by(CustomInventoryItem.product_code.asc())
+    )
+    if type_id is not None:
+        stmt = stmt.where(CustomInventoryItem.type_id == type_id)
+    if active_only:
+        stmt = stmt.where(CustomInventoryItem.is_active.is_(True))
+    if product_code_contains:
+        stmt = stmt.where(CustomInventoryItem.product_code.ilike(f"%{product_code_contains.strip()}%"))
```
Xilous: `product_code_contains` does not escape LIKE metacharacters.

`%` and `_` are wildcards. searching `50%` returns everything, `FR_1` matches `FR-1`. no frontend caller passes it yet, it is an argument on a `SIGNED_IN` root field.

escape the backslash, `%` and `_`, and pass `escape="\\"` to `ilike`.

backend/app/repositories/custom_items_repository.py
```diff
@@ -0,0 +1,420 @@
...
+    product_code: str,
+    description: str | None = None,
+    values: list[dict] | None = None,
+) -> CustomInventoryItem:
+    """Catalog a product. `values` is [{attribute_id, value}] and may cover any subset of the type's
+    attributes - an item whose fire rating is not known yet is still worth having a code for."""
+    product_code = (product_code or "").strip()
+    if not product_code:
+        raise ValidationError("An item needs a product code.", field="product_code")
+    get_item_type(session, type_id)
```
Xilous: nothing stops cataloguing new products under a retired type.

`create_item` and `create_attribute` call `get_item_type` without checking `is_active`. the UI hides retired types from the pickers, the mutations accept them, so a stale tab or a direct call keeps extending a type the warehouse deliberately switched off.

frontend/src/modules/warehouse/CustomItemsPage.tsx
```diff
@@ -0,0 +1,675 @@
...
+            ))}
+          </Stack>
+        </Box>
+      )}
+
+      {item && (
+        <Button
+          size="small"
+          sx={{ alignSelf: 'flex-start' }}
+          onClick={() => updateItem({ variables: { id: item.id, isActive: !item.isActive } })}
```
Xilous: "Retire item" silently discards whatever is in the form.

the button fires `updateItem` with `isActive` only, and `onCompleted: onSaved` closes the dialog and refetches. anything typed into Description or the attribute fields first is gone with no indication.

either include the current form values in the same mutation, or move the toggle out of the form.

Xilous:
fixed in ede5622. 13 of 15 findings applied, 2 answered.

the `values` argument. building the schema gave `createCustomInventoryItem(..., values: [CustomInventoryItemValueInput!]!)`, and omitting it raised `TypeError: 'StrawberryField' object is not iterable`. `strawberry.field(default_factory=list)` is for input-type attributes, not resolver arguments. now `list[...] | None = None`, SDL reads `values: [CustomInventoryItemValueInput!] = null`, omitting it reaches the resolver body. the frontend always sends the argument, which is why e2e never hit it.

- selectinload chained to `CustomInventoryItemValue.attribute` in `get_items` and `get_item`. `_item_to_type` reads `v.attribute.name` per value, so this was one SELECT per value
- `_check_code_free` sweeps `hardware_items` first. that is where a category enters the system: a schedule can be imported long before anything is ordered against it
- migration 084 skips seeding a code already present as a hardware category on
  hardware_items
  po_line_items
  inventory_locations
  stock_items
- fresh database still seeds all three
- `create_item` and `create_attribute` go through a new `_require_active_type`, retired type takes nothing new
- `product_code_contains` escapes LIKE metacharacters, passes `escape="\\"`
- dropped the discarded `session.refresh(item)`
- inventory search matches the type's display name as well as the raw code, via a `groupSearchText` helper. searching "Frames" was filtering out the FRAME group
- retire is an X icon, reactivate stays the undo arrow
- code preview uses `\p{L}\p{N}` with the u flag to match Python `str.isalnum()`
- "Retire item" sends description and values alongside isActive, so it no longer discards unsaved edits
- `activeAttributes` memoised in TypePanel so DataGrid columns keep a stable identity
- `byKey` uses the `catalogKey` helper

four new tests cover
hardware_items collision
retired-type guard
LIKE escaping
item created with no values

- auth policy stays SIGNED_IN. the issue puts creation and maintenance in warehouse users' hands and no Clerk role exists for a plain warehouse user. can gate the type mutations on Warehouse Manager if you want the irreversible half locked down, your call.
- full catalog fetch on the inventory tab stays, two queries with the selectinload fixed. the catalog is a curated list, not a hardware master, and the join it powers has no foreign key to push into the resolver. past a few hundred rows -> a resolver filtering by the categories present in the hierarchy.
- two design-hook findings on HardwareItemsTab (1.125rem font sizes at the Inventory Summary totals) are pre-existing on master, lines 760 and 771. diff adds none, line numbers shifted.

PR #478, opened 2026-08-04
test(shipping): pin the staging drag decision, and a fixture that can fill a skid (#470)

### LINKED ISSUE #470: staging containers: the leaf paths are untested - cross-container leaf move and the 30-leaf skid cap
two behaviours from #451 phase 3 never exercised against a running system, one with no unit test either. both need an assembled door leaf; every e2e run on the #451 stack seeded loose hardware only.

moving an assembled leaf between containers

`handleDragEnd` (`StagingWorkspace.tsx:298`) branches when the drag starts on a container item and the drop target is a different container: remove from source -> await save -> append to target. the await is required, add-first is refused for a leaf still recorded in the skid it is leaving. `OPENING_ITEM` also checks `canTakeAnotherLeaf(target)` first and warns instead of moving (`:303`).

no frontend coverage. dnd-kit needs a pointer with real geometry, jsdom has none, so `StagingWorkspace.test.tsx` drives the click-equivalent controls.

the `LOOSE` half was verified live on pr-465 through dnd-kit's keyboard path (space lift -> arrow move -> space drop) plus the aria-live region. a leaf never was, and the leaf path is the one carrying the extra guard.

the 30-leaf skid cap

`MAX_LEAVES_PER_SKID` is enforced server-side, covered by `test_a_skid_stops_at_thirty_leaves`, negative case `test_a_door_cart_has_no_leaf_ceiling`.

untested on the frontend
- `canTakeAnotherLeaf` gating the Place-in menu
- the `n/30` chip on the skid card (`:723`)
- the warning toast at `:193` and `:303`

never seen live, no e2e run has had 30 assembled leaves, hand-seeding 30 is not practical.

## getting both to full e2e

one assembled leaf is reachable through the UI today, on a PR env with received inventory seeded. ~20 minutes on top of an existing seed
- Start a Task, purpose "Pull Request for Shop Assembly" over the opening -> mints the SAR + its pull
- Warehouse -> Pull Requests -> Shop Assembly -> start pick -> confirm pick -> Mark as Pulled
- Shop Assembly -> assign the work unit -> open assembly modal -> mark lines installed -> complete leaf -> IN_INVENTORY
- shipping-out request carrying the OPENING_ITEM line -> accept -> pick -> Mark as Pulled -> SHIP_READY, shows in staging pool under Door leaves

that one leaf reaches the whole cross-container branch
- create two skids
- place the leaf in the first
- move it to the second through dnd-kit's keyboard path (space lift, arrows move, space drop, watch the aria-live region for the droppable id)
- assert it left the source, arrived in the target with opening and leaf intact, pool still counts it placed exactly once

thirty leaves is not reachable by hand. needs a fixture endpoint - a TESTING_ENABLED-gated route that mints N SHIP_READY OpeningItems for a project without the pull cycle, gated exactly like `/testing/clerk-sign-in`, TESTING_ENABLED plus an admin bearer or the X-Testing-Secret whose SHA-256 sits in TESTING_SIGN_IN_SECRET_HASH. same double gate #422 and #424 settled on, and the same reason - a fixture has to be unreachable anywhere that is not a test target.

cap test on top of that
- mint 31 leaves
- fill a skid to 30
- assert the chip reads 30/30
- Place-in menu offers the skid as full
- 31st refused in the UI
- `setContainerItems` refuses it too when driven straight at the API, which is the gate that actually matters

alongside, not instead: pull the cross-container decision out of `handleDragEnd` into a pure function in `staging.ts` beside `sameLooseStock` and `toItemsInput`. takes the containers, the active id and the over id, returns what to save. that stops the e2e run being the only thing between this branch and a regression.

Xilous:
closes #470

two behaviours from #451 phase 3 never ran against a live system, one with no unit test either. both need an assembled door leaf; every e2e run on the #451 stack seeded loose hardware only.

planContainerDrag(containers, activeId, overId) in staging.ts takes the cross-container branch out of handleDragEnd, sits beside sameLooseStock and toItemsInput, and returns:
- reorder - same container, new stacking order
- move - source list and target list, both written, source first and awaited
- refuse - leaf onto a skid already holding 30
- none - dropped where it started
- null - drag came off the pool, needs the pool rows, stays in the component

source first and awaited because one leaf, one container is enforced server-side against the pool as it stands at save time. refused rather than moved because the alternative empties a skid the leaf was safely on and then has the second save rejected.

containerDropId and dropTargetContainer moved with it, so the pool branch resolves a drop target through the same lookup.

19 unit tests in frontend/src/modules/shipping/__tests__/staging.test.ts:
cross-container leaf move
leaf counted exactly once across the two writes
append to top of target stack
30 ceiling, 30th allowed
door carts have no ceiling
loose stock tops up a matching line
another opening's units stay a separate line

POST /testing/seed-ship-ready-leaves mints N SHIP_READY OpeningItems for a project with no pull cycle behind them. reaching 30 by hand is 30 openings through purchase, receipt, a shop assembly pull, a pick, assembly, a shipping pull and a second pick - so the n/30 chip, the greyed-out Place-in entry and the refusal toast shipped having been read and never run.
each gets its own opening number and its own opening_id, satisfying uq_opening_items_live_leaf by construction, so a second call adds to the pool.
one OpeningItemHardware row each, inert to the pool arithmetic, which counts pull requests and packing slips.
count arrives on a query string, capped at 200.

require_testing_request in app/auth.py checks TESTING_ENABLED -> admin bearer or X-Testing-Secret preimage of TESTING_SIGN_IN_SECRET_HASH. TESTING_ENABLED first so a production deployment refuses outright rather than leaking whether the credential would have been good enough.
/testing/clerk-sign-in calls it too, identical behaviour and response bodies.
added to _ROUTE_GATES in test_resolver_gate_completeness.py, so a /testing/* route added without it fails test_route_calls_an_auth_gate.

backend tests in tests/test_seed_ship_ready_leaves.py:
four gate refusals
the secret opening it
seeded leaves landing unplaced in the staging pool
second call not colliding
31 seeded, skid filled to 30, 31st refused
unknown project
count bounds

e2e on pr-478, project 99000 BC-TESTING, relay connected and pinned to TUBC.

assembled-leaf chain driven through the UI
buyer assignment for mira -> import opening 0501-EX (pair, RHRA/LHR) with purpose Create Purchase Orders -> single-line draft PO-REQ-001 -> registered as PO0000091 (single line on purpose, the 9191 multi-line relay bug) -> received 3 units into WRD A-1-1 -> Warehouse Manager approved -> SA-470-E2E accepted -> picked -> both carts staged -> both leaves assembled and completed -> SHIP-470-E2E accepted -> fetched -> marked as pulled.
both leaves reached SHIP_READY, appeared in the staging pool as Door leaves.

cross-container leaf move, dnd-kit keyboard path
leaf 1 -> Skid 1, leaf 2 -> Skid 2, focus leaf 1 Reorder handle -> Space to lift -> one ArrowDown (aria-live region named droppable `container:4c55e169-...`, which is Skid 2) -> Space to drop.

- Skid 1 went to 0 items, chip 0/30
- Skid 2 holds both, leaf 1 appended at position 1 behind leaf 2, opening number and leaf intact, same openingItemId
- the container row got a NEW id, so the source really was emptied and the target rewritten rather than a row relabelled
- stagingPool reports leaf 1 placedInContainerId = Skid 2, appears exactly once across every container

30-leaf cap, off 31 leaves minted by the fixture route in one call

- chip reads 30/30, flips to colorWarning
- Place-in menu renders the option as `Skid 3 (full)`, disabled
- dragging a leaf out of Skid 2 onto Skid 3 refused with `Skid 3 already holds 30 leaves - start another skid.`, neither container changes
- setContainerItems driven straight at the API with 31 leaves fails VALIDATION_ERROR / field items, Skid 3 still holds exactly 30, so nothing lands partially

one assertion not exercised live. `placeLeaf`'s full-skid guard fires only on a pool-to-container drag, and there is no keyboard drag out of the pool - a pool row is plain useDraggable, sortableKeyboardCoordinates has no sortable context for it, arrow keys do nothing. the Place-in menu disables a full skid rather than letting the click through, so nothing reaches the guard from there either. same canTakeAnotherLeaf predicate and same message string proven on the cross-container path. pre-existing, untouched by this PR, worth its own issue.


Xilous:
superseded by #528. that branch was cut from the pre-rewrite master, so it read as 64 commits diverged and conflicting; only 2 of them were its own work. those 2 are replayed onto current master, same files, same insertions.

Issue #506, opened 2026-08-05
warehouse inventory summary should be a flat item-level table, not an accordion


Xilous:
warehouse inventory - inventory summary table should not be an accordion.

it should be a raw excel table at a hardware or other inventory item level, not at an opening level.

Xilous:
the Hardware Items tab in frontend/src/modules/warehouse/InventoryView.tsx renders the category -> product -> locations accordion served by get_inventory_hierarchy (backend/app/repositories/warehouse/inventory.py lines 50-100). the accordion goes, replaced by a flat table at inventory-item level ("a raw excel table"). the Opening Items tab stays as is.

1. flat MUI DataGrid, one row per InventoryLocation line - the warehouse reads it as "what is on which shelf", and a product-level rollup is one sort/filter away. each row shows:
   Hardware Category
   Product Code
   Warehouse
   Location (aisle-row-bay, dash when unlocated)
   Qty
   Deficient
   Unit Cost
   Line Value
   Received date
   Project (All Projects view only)
   sortable everywhere, quick filter (search box), column-level filtering via the DataGrid.
2. enable the DataGrid toolbar with CSV export.
3. footer strip with total units and total value for the filtered set, computed client-side from loaded rows.
4. new flat query inventoryItems(projectId, warehouseId) - the repository already selects flat rows before grouping (inventory.py lines 51-70: InventoryLocation outerjoin POLineItem for unit_cost, quantity > 0 filter). single query, no N+1. keep get_inventory_hierarchy if anything else consumes it (grep first), delete it if the tab was its only consumer.
5. tests cover resolver flat rows with unit cost and location fields, frontend render + export, totals math.

simulated user testing runs click-through on a PR preview environment with the test relay attached:
1. create job -> register PO (two products) -> receive into two different locations.
2. open Warehouse -> Inventory -> Hardware Items, verify a flat grid (no expanding rows) with one row per product/location, correct qty, unit cost, value, location text.
3. sort by Product Code, filter by category, search a product, verify each narrows the grid.
4. export CSV, verify the file downloads and contains the visible rows.
5. switch to All Projects, verify the Project column appears and rows from both projects list.

Xilous:
built on PR #534. two deviations from the plan.

tab is `HardwareItemsTab.tsx`, not `InventoryView.tsx` - InventoryView is the shell hosting the Hardware Items and Opening Items tabs.

[...]

PR #510, opened 2026-08-05
feat(admin): migrate the legacy SharePoint inventory list into Nexus

Xilous:
one-time admin wizard at Admin -> SharePoint Migration, migrates the pre-Nexus warehouse inventory list (SharePoint, UCH-Operations site, 3,506 rows) into Nexus stock and project inventory. also adds a general flag for inventory that does not line up with a hardware schedule.

reading
app/services/sharepoint_inventory.py -> Microsoft Graph, app-only client credentials, same Entra app registration the Gone Fishing repo uses (application-type Sites.ReadWrite.All, tenant-wide, admin-consented). read-only, paged, module-level token cache with a 5 minute refresh buffer.
AZURE_TENANT_ID
AZURE_CLIENT_ID
AZURE_CLIENT_SECRET
unset -> wizard refuses with a configuration error, app still boots.

writing
sharepoint_migration_repository.migrate_inventory writes each resolved entry through receive_into_stock -> allocate_stock_to_project. both destinations go through the stock pool because migrated hardware has no PO in Nexus, and ck_inventory_locations_has_origin requires a PO origin, a stock origin or a return origin.
whole batch validated before anything is written, all-or-nothing. no resume state, so a half-failure re-run would double whatever landed.
both root fields ADMIN_ROLE.

wizard prompts
- locations. A-62R parses to aisle A / row 62 / bay R, hyphen-less G8R parses too, ~1,954 rows. remaining ~1,212 collapse to a few dozen distinct strings (NS-Q, Coast, Warehouse Overflow, SHIPPED), given a warehouse and shelf by hand or excluded. 340 rows have no location at all and migrate unlocated. a partially-read value like "G-22L - Blue Stand Up Rack" is NOT auto-parsed.
- projects. auto-matched on project number only, including a number recovered from the front of a name like "21968 - VPO". never on name alone. unmatched projects are picked or excluded. the wizard never creates projects.
- categories. 189 rows have no Part Category 1; one prompt assigns a label or excludes them.

decisions
- product_code is the Scheduled Part Number (TITAN-schedule identity), falling back to the SharePoint Part Number. Part Number is what Nexus calls "order as", not inventory identity. 904 of 953 company-stock rows have no Scheduled PN - shelf stock was never on a schedule.
- hardware_category is Part Category 1 verbatim, no normalization.
- Staged Qty excluded: those units were already deducted by a pull request into shop assembly or shipping out, so migrating them would put back hardware that has left. Ordered / Received / Shipped are pipeline and history.
- a multi-location row goes entirely on its first coordinate. the source records that a line spans several shelves but never how much sits in each.

schedule mismatch flag
ProductCodeNode.matches_schedule is false when a (hardware_category, product_code) pair appears nowhere in the project's imported schedule. flagged units can never be claimed, because both pull request builders start from the schedule and match the exact pair. surfaces as a "Not on schedule" chip in the warehouse inventory view. one grouped query per request, no N+1. defaults true where the answer is unknowable - the by-vendor hierarchy, and the unscoped inventoryHierarchy.
general mechanism, catches stock allocations and reclassifies too.

drive-by
inventoryByVendor omitted the required total_available_quantity when constructing ProductCodeNode, so every call raised TypeError. now passes 0.

accepted limitations
- unit cost not carried. StockItem and InventoryLocation have no cost column, by-value views show 0 for migrated rows.
- received_at is migration time, not the original receipt date.
- one-shot tool. no idempotency marker beyond a warning banner when Nexus already holds inventory.

deployment
AZURE_TENANT_ID / AZURE_CLIENT_ID / AZURE_CLIENT_SECRET set by hand on the PR environment and on production at cutover; PR environments only clone production variables at creation.
no DB migration.


frontend/src/graphql/warehouse.ts
```diff
@@ -34,6 +34,7 @@ export const GET_INVENTORY_HIERARCHY = gql`
         totalQuantity
         totalAvailableQuantity
         totalValue
+        matchesSchedule
```
Xilous: GET_INVENTORY_BY_VENDOR never selects matchesSchedule. ProductCodeGroup (HardwareItemsTab.tsx:67) declares it non-optional, and both the category and vendor hierarchies render through the same ProductCodeDetail.

vendor mode -> pc.matchesSchedule undefined -> !matchesSchedule at HardwareItemsTab.tsx:510 true -> "Not on schedule" chip on every single product code in the view.

defeats the `matches_schedule: bool = True` default on the by-vendor node - the default is never asked for.

add matchesSchedule to the productCodes selection set in GET_INVENTORY_BY_VENDOR (warehouse.ts:308-317).

frontend/src/modules/warehouse/HardwareItemsTab.tsx
```diff
@@ -500,6 +507,18 @@ function ProductCodeDetail({
                 : undefined
             }
           />
+          {!matchesSchedule && (
```
Xilous: chip fires on all legitimately-catalogued non-schedule stock. #454/#473 made non-schedule inventory (frames, specialty, consumables) a first-class concept with its own catalog, and none of it appears in hardware_items -> get_scheduled_pairs never contains those pairs -> every catalogued row branded "Not on schedule ... no shop assembly or shipping out request can claim it".

- suppress the chip when catalogItem is set - ProductCodeDetail already receives it for exactly these rows (HardwareItemsTab.tsx:884)
- or union the catalog's (category, code) pairs into get_scheduled_pairs on the backend, so the flag means what it says

backend/app/services/sharepoint_inventory.py
```diff
@@ -0,0 +1,160 @@
...
+            for entry in payload.get("value", []):
+                fields = entry.get("fields", {})
+                items.append({"sp_item_id": str(entry.get("id", "")), **fields})
+
+            next_link = payload.get("@odata.nextLink")
+            if not next_link:
+                break
+            url = next_link
+        else:
+            logger.warning("fetch_inventory_items hit the %d-page cap at %d rows", _MAX_PAGES, len(items))
```
Xilous: silent truncation. the for/else logs a warning, then fetch_inventory_items returns the partial list as if it were complete.

Graph's effective page size for /lists/{id}/items?$expand=fields is routinely far below the requested $top=2000 (200, sometimes 100), so 3,506 rows can need 18-36 pages, not the two the comment assumes.

at the cap the wizard reports a smaller "Rows in SharePoint" count with no way for the operator to tell, a subset migrates, and there is no idempotency marker, so the missing rows can never be safely added afterwards.

raise SharePointUnavailableError in the else branch instead of returning short.

backend/app/schemas/sharepoint_migration.py
```diff
@@ -0,0 +1,91 @@
...
+
+def _to_str(value) -> str:
+    return str(value).strip() if value is not None else ""
+
+
+@strawberry.type
+class SharepointMigrationQueries:
+    @strawberry.field
+    def sharepoint_inventory_snapshot(self, info: strawberry.Info) -> SharepointInventorySnapshot:
+        rows = sharepoint_inventory.fetch_inventory_items()
```
Xilous: sharepoint_inventory_snapshot is a sync resolver making a blocking HTTP call. graphql-core invokes sync resolvers inline on the event loop, so fetch_inventory_items() (30s token POST + up to 120s list read, per page) stalls every other request to the backend for its whole duration, health checks included.

_token_lock is a threading.Lock held across the token POST - a second concurrent caller blocks the loop on lock acquisition.

every other resolver in this codebase that talks to an external system is async def:
relay.py gp_* queries
po.register_po_in_gp
project.sync_gp_jobs

make the resolver async def and use httpx.AsyncClient, or offload with asyncio.to_thread.

backend/app/repositories/sharepoint_migration_repository.py
```diff
@@ -0,0 +1,184 @@
...
+    # Validate the whole batch before writing anything, so a bad row 2000 does not leave the caller
+    # guessing which of the first 1999 went in.
+    _validate_entries(session, entries)
+
+    now = datetime.utcnow()
+    stock_items = 0
+    project_locations = 0
+    total_units = 0
+
+    for index, entry in enumerate(entries):
```
Xilous, line 147: the whole migration is one mutation, one transaction, ~2,000+ entries, roughly five statements per entry
_find_or_create_stock_row SELECT
INSERT + flush
get_stock_item
IL insert + flush
2-4 audit inserts

on the order of 10k round trips to Railway's managed Postgres - the trap CLAUDE.md's GraphQL/SQLAlchemy performance rules describe.

edge proxy timeout -> client sees a failure while the backend may well have committed -> the wizard's only recourse is to run it again, which doubles everything.

chunk into batches with a server-side idempotency key.

backend/app/repositories/sharepoint_migration_repository.py
```diff
@@ -0,0 +1,184 @@
...
+        destination = entry.get("destination")
+        if destination not in (DESTINATION_PROJECT, DESTINATION_STOCK):
+            raise ValidationError(f"{label}: destination must be PROJECT or STOCK", field="destination")
+
+        if not (entry.get("hardware_category") or "").strip():
+            raise ValidationError(f"{label}: hardware_category is required", field="hardware_category")
+        if not (entry.get("product_code") or "").strip():
+            raise ValidationError(f"{label}: product_code is required", field="product_code")
+
+        quantity = entry.get("quantity")
```
Xilous, line 241: location fields are never length-validated on the STOCK path. _validate_entries checks nothing about aisle/row/bay, and receive_into_stock never calls _validate_location_fields - allocate_stock_to_project and destock_inventory do.

StockItem.aisle/row/bay are String(20). the wizard's aisle/row/bay TextFields (SharePointMigrationPage.tsx:352-364) carry no maxLength.

"Warehouse Overflow Rack" typed into Aisle for a hand-mapped location string -> _find_or_create_stock_row's session.flush() raises StringDataRightTruncation -> except (ValidationError, NotFoundError) does not catch it -> batch dies with a raw 500 naming no entry, which is what the entry-numbering exists to prevent.

add _validate_location_fields to _validate_entries.

backend/app/services/sharepoint_inventory.py
```diff
@@ -0,0 +1,160 @@
...
+
+    items: list[dict] = []
+    with httpx.Client(headers={"Authorization": f"Bearer {token}"}, timeout=120) as client:
+        for _ in range(_MAX_PAGES):
+            try:
+                resp = client.get(url)
+                resp.raise_for_status()
+            except httpx.HTTPStatusError as e:
+                logger.warning("Graph list read failed: %s %s", e.response.status_code, e.response.text[:400])
+                raise SharePointUnavailableError(
```
Xilous, line 173: a 401 during the list read leaves the rejected token cached. _get_token only refreshes on expiry, and the 401/403 handler here raises without clearing _token / _token_expires_at.

client secret expires mid-session -> the wizard's Retry button re-sends the same rejected token for up to an hour -> failure looks permanent.

reset _token = None / _token_expires_at = 0 before raising on a 401 or 403.

backend/app/repositories/sharepoint_migration_repository.py
```diff
@@ -0,0 +1,184 @@
...
+        found = set(session.scalars(select(ProjectModel.id).where(ProjectModel.id.in_(project_ids))).all())
+        missing = project_ids - found
+        if missing:
+            raise ValidationError(
+                f"Unknown project(s): {', '.join(str(m) for m in sorted(missing, key=str))}",
+                field="project_id",
+            )
+
+
+def has_any_inventory(session: Session) -> bool:
```
Xilous, line 288: has_any_inventory cannot answer the question it is asked. any Nexus environment that has ever received a PO has inventory_locations rows -> already_has_inventory unconditionally true in production -> the "Nexus already holds inventory" banner fires for everyone on the very first run, and a warning that is always on gets clicked past.

it also cannot distinguish "the migration already ran" from "the warehouse has been operating normally".

a persisted marker written by migrate_inventory - an audit-log row or a settings key - answers it.

backend/app/repositories/sharepoint_migration_repository.py
```diff
@@ -0,0 +1,184 @@
...
+                    target_product_code=code,
+                    quantity=quantity,
+                    target_aisle=aisle,
+                    target_row=row,
+                    target_bay=bay,
+                    performed_by=performed_by,
+                )
+                project_locations += 1
+            else:
+                stock_items += 1
```
Xilous: stock_items counts entries, not rows created. receive_into_stock goes through _find_or_create_stock_row, which merges into an existing row matching (warehouse, category, code, aisle, row, bay).

two SharePoint rows for the same part on the same shelf -> one StockItem, stock_items incremented twice -> the completion screen's "N stock rows" overstates and cannot be reconciled against the warehouse view.

collect the distinct stock_row.id values and report len().

backend/app/schemas/warehouse.py
```diff
@@ -721,6 +732,10 @@ def inventory_by_vendor(
                             product_code=pc["product_code"],
                             items=[inventory_location_to_type(il) for il in pc["items"]],
                             total_quantity=pc["total_quantity"],
+                            # This hierarchy groups by vendor across projects and has never computed
+                            # the available split; 0 is the documented placeholder. It was omitted
+                            # entirely until now, which made every call raise TypeError.
+                            total_available_quantity=0,
```
Xilous: this drive-by ships with no test. by the PR's own account inventoryByVendor raised TypeError on every call - the by-vendor toggle was completely broken on master and nothing caught it.

test_warehouse_repository_inventory_hierarchy.py covers only the repository function, never the resolver, so a missing required constructor arg went unnoticed.

add a resolver-level test that executes inventoryByVendor through the schema. it would have caught the missing matchesSchedule in the frontend query too.

backend/tests/test_inventory_schedule_match.py
```diff
@@ -0,0 +1,119 @@
...
+            product_code=code,
+            quantity=quantity,
+            deficient_quantity=0,
+            received_at=datetime.utcnow(),
+        )
+    )
+    session.flush()
+
+
+def test_scheduled_pairs_returns_the_exact_pairs(db_session):
```
Xilous, line 79: matches_schedule is never tested through the resolver. this file exercises get_scheduled_pairs in isolation; the behaviour under review is the wiring in warehouse.py:422-426. none of these three is asserted anywhere:
- matches_schedule=False for a pair absent from the schedule
- True for every node when project_id is omitted
- True on the by-vendor hierarchy

backend/app/repositories/sharepoint_migration_repository.py
```diff
@@ -0,0 +1,184 @@
...
+            )
+
+
+def has_any_inventory(session: Session) -> bool:
+    """Whether Nexus already holds inventory, for the wizard's re-run warning.
+
+    The migration has no idempotency marker, so running it twice doubles everything it wrote. This
+    is what the first step warns on.
+    """
+    from app.models.inventory import InventoryLocation
```
Xilous: InventoryLocation and StockItem are imported function-locally while Project and Warehouse are imported at module level in the same file. no circular-import reason - app.repositories.stock already imports both at module scope.

move them to the top.

backend/app/repositories/sharepoint_migration_repository.py
```diff
@@ -0,0 +1,184 @@
...
+    project_ids: set[uuid.UUID] = set()
+
+    for index, entry in enumerate(entries):
+        label = f"Entry {index + 1}"
+        destination = entry.get("destination")
+        if destination not in (DESTINATION_PROJECT, DESTINATION_STOCK):
+            raise ValidationError(f"{label}: destination must be PROJECT or STOCK", field="destination")
+
+        if not (entry.get("hardware_category") or "").strip():
+            raise ValidationError(f"{label}: hardware_category is required", field="hardware_category")
```
Xilous, line 237: _validate_entries tests the stripped form, migrate_inventory writes the raw value. "Hinge " passes validation and lands in stock_items / inventory_locations as a distinct identity that can never match a schedule pair.

the wizard trims, but this is an admin mutation on the public schema and should not depend on one caller's behaviour.

strip in _validate_entries and write the stripped value back onto the entry.

backend/tests/test_sharepoint_migration.py
```diff
@@ -0,0 +1,188 @@
...
+    with pytest.raises(ValidationError) as e:
+        migration_repo.migrate_inventory(db_session, [_entry(wh), _entry(wh, quantity=0, product_code="BAD")], ACTOR)
+    assert "Entry 2" in e.value.message
+
+
+def test_has_any_inventory_tracks_the_rerun_warning(db_session):
+    wh = warehouse_admin_repository.get_primary_warehouse_id(db_session)
+    migration_repo.migrate_inventory(db_session, [_entry(wh)], ACTOR)
+    db_session.flush()
+    assert migration_repo.has_any_inventory(db_session) is True
```
Xilous, line 181: only the True side is pinned. nothing asserts has_any_inventory(db_session) is False on an empty database - a helper that unconditionally returned True would pass.

frontend/src/modules/admin/SharePointMigrationPage.tsx
```diff
@@ -0,0 +1,585 @@
...
+                              </Select>
+                            </TableCell>
+                            {(['aisle', 'row', 'bay'] as const).map((field) => (
+                              <TableCell key={field}>
+                                <TextField
+                                  size="small"
+                                  value={r?.[field] ?? ''}
+                                  onChange={(e) =>
+                                    setLocation(loc.raw, {
+                                      [field]: e.target.value || null,
```
Xilous, line 423: editing any location cell silently re-includes an excluded row. every aisle/row/bay keystroke forces excluded: false, and clearing a field does the same.

exclude SHIPPED -> tidy its aisle field -> all of its rows back in the batch, no signal beyond the button label flipping.

leave excluded alone in the field handler and let the Include/Exclude button own it.

Xilous:
write path is compatible with #473 - receive_into_stock -> allocate_stock_to_project is the same free-text (hardware_category, product_code) model, and the views degrade gracefully for unknown categories. the migration writes Part Category 1 verbatim and never consults item types, and that produces four problems:

1. 904 of 953 company-stock rows have no Scheduled PN, and verbatim mixed-case categories match no type code (typesByCode is an exact-case Map keyed on the normalized uppercase code). none of the migrated stock gets a type badge, catalog description, or attributes.
2. create_item_type refuses a code already in use as a hardware category (case-insensitive sweep of inventory_locations and stock_items), codes are immutable, and there is no adopt-existing path. every distinct Part Category 1 spelling permanently blocks a type of that name. migrating "Frames" stock means a Frames type (code FRAMES) is refused forever.
3. the reverse direction is unguarded - nothing stops a migrated category colliding with an existing type code. a category spelled exactly FRAME is silently adopted into the seeded FRAME type's grouping; spelled "Frame", it becomes a parallel category group non-fungible with the type's stock.
4. the empty-category prompt is free text with no type awareness - typing "Consumables" creates a near-miss against the seeded CONSUMABLE type. the "Not on schedule" chip renders regardless of catalogItem, so a migrated frame allocated to a project shows the warning chip next to the catalog chip.

these would reconcile it:
- a category-mapping step symmetrical to the location/project steps - map each distinct Part Category 1 value onto a schedule category spelling or an item type (writes the type's code), with create-on-the-fly
- or the cheaper backstop, migrate_inventory refusing categories that case-insensitively collide with a type code without exactly matching it
- suppress the "Not on schedule" chip when the category is a known type code or the pair has a catalog entry

settle 1 and 2 before this runs on production - the migration is one-shot and all-or-nothing, so the category spellings are effectively permanent and they constrain which entity types can ever exist.

Xilous:
parked until the warehouse remediation batch lands, then one reconcile against master:

- receive_into_stock / allocate_stock_to_project moved to repositories/stock/movements.py
- the Not on schedule chip re-homes onto HardwareItemsFlatTable (#506 replaced the accordion tab)
- the inventoryByVendor drive-by fix is probably dead code by then - drop it if the accordion queries are deleted
- category prompts get an explicit exclusion path for door/frame and shop-assembly stock, out of scope since #554
- squash-merge only: a merge commit message on this branch fails the no-ai-attribution scan
- then re-run #512's full-scale timing on a fresh throwaway env before this touches production

Issue #438, opened 2026-07-30
drop the 27 deprecated client-supplied actor arguments once deployed frontends stop sending them


Xilous:
follow-up to #427 (PR #437).

#427 stopped every audit-writing resolver from reading a client-supplied actor. the arguments stayed in the schema, nullable + @deprecated, accepted and discarded server-side - dropping them outright would break a tab loaded before that deploy, which LazyBoundary cannot catch (it only handles chunk errors). same pattern as saveBuyerAssignment(costCodes:) from #430.

the frontend already sends none of them. drop when deployed frontends no longer reference them.

arguments
startPullRequestPick.startedBy
savePickDraft.enteredBy
confirmPick.pickedBy
setPullItemFetched.fetchedBy
completePullRequest.completedBy
acceptShopAssemblyRequest.acceptedBy
rejectShopAssemblyRequest.rejectedBy
acceptShippingOutRequest.acceptedBy
rejectShippingOutRequest.rejectedBy

input fields
TransferInventoryInput.performedBy
ConfirmShipmentInput.shippedBy
CreateShipmentReturnInput.returnedBy
RecordAssemblyProgressInput.performedBy
InstallReplacementInput.performedBy
StagePullOpeningsInput.stagedBy
CancelPullRequestInput.cancelledBy
CompleteOpeningInput.completedBy
DestockInventoryInput.performedBy
AllocateStockToProjectInput.performedBy
AdjustStockQuantityInput.performedBy
MoveStockLocationInput.performedBy
ReclassifyStockItemInput.performedBy
ReportInventoryDeficiencyInput.performedBy
OverrideInventoryQuantityInput.performedBy
ReportStockDeficiencyInput.performedBy
ReportDeficiencyAtAssemblyInput.performedBy
ResolveDeficiencyInput.reviewedBy

also drop saveBuyerAssignment(costCodes:) and BuyerAssignment.costCodes from #430 in the same pass.

separately, ShippingOutPRDraftInput.requestedBy is dead: the import wizard collects it and imports.py forwards it, but import_repository.finalize_import_session never reads it - it hardcodes created_by="Hardware Schedule Import". not a security issue, nothing client-supplied is written, but it is a misleading field in the wizard UI. remove the field and the UI input, or wire it through if the intent was to record the requester.

PR #511, opened 2026-08-05
refactor(vendors): remove the Nexus-local vendors table, GP owns vendors

### LINKED ISSUE #509: remove the Vendors admin page and the local vendors table - GP owns vendors
delete the `vendors` table and every consumer. GP (PM00200) becomes the only vendor authority. a draft PO shows no vendor until it is registered into GP.

#200 already moved GP vendor identity onto the PO - `gp_vendor_id` + `vendor_name_snapshot`, picked live from `gpVendors` through the relay at register time, never cached. `vendors` survived as an optional contact record on `purchase_orders.vendor_id` with no GP linkage.

`createVendor` is SIGNED_IN, not admin - `VendorSelect` offers "+ Create new vendor" to every PO user, so anyone can invent a vendor with no PM00200 counterpart. that is the drift #200 set out to end.

the contact columns are dead. the PO document takes the vendor mailing address from `PODocumentData.vendor_address`, typed into the generate dialog. nothing reads `Vendor.contact_name` / `email` / `phone`.

these go:

`vendors` table
`purchase_orders.vendor_id`
admin Vendors page (Create / Edit / Delete)
`createVendor` / `updateVendor` / `deleteVendor` / `vendors` / `vendor`
`VendorSelect`, including "+ Create new vendor"
`GpPurchaseOrderDialog` draft-mode `VendorSelect`
`adminStats.vendorCount`
Vendors card on the admin landing

three deliberate behaviour changes fall out:

- `markPoAsOrdered` drops its vendor requirement, keeps the PO-number one
- `priorOrderAsValues` re-keys from vendor id to product code alone. today the suggestions are skipped unless someone linked a local vendor; keyed by product code they always fire. what a part was last ordered as is a property of the part, not of who sold it
- `inventoryByVendor` groups by `vendor_name_snapshot` instead of joining `PO.vendor_id` -> `Vendor.name`. fixes existing degradation: the modern Create PO flow never sets `vendor_id`, so the By Vendor view was already collapsing everything into "No Vendor"

these stay:

`gpVendors`
`suggestVendorForManufacturer`
`manufacturer_vendor_map`
`PurchaseOrder.gp_vendor_id`
`vendor_name_snapshot`
`vendor_quote_number`
`PODocumentData.vendor_address`
the relay
`HardwareItem.vendor_no`

if vendor contact info is ever wanted in Nexus, read it live from PM00200 through the relay.

Xilous:
closes #509.

deletes the Nexus-local `vendors` table and every consumer. GP (PM00200) becomes the only vendor authority.

`createVendor` was SIGNED_IN, not admin, so any PO user could invent a vendor with no PM00200 counterpart.

this PR removes:
`vendors` table
`purchase_orders.vendor_id` (migration 085)
admin Vendors page
`VendorEditDialog`
`VendorSelect`
`vendors` resolver
`vendor` resolver
`createVendor` resolver
`updateVendor` resolver
`deleteVendor` resolver
`markPoAsOrdered` resolver
their auth_policy entries
`adminStats.vendorCount`
Vendors card on the admin landing
draft-mode vendor field in `GpPurchaseOrderDialog`
edit-mode vendor field in `PODetailModal`
dead `VendorRef` / `PODetails.vendor` / `OpenPO.vendor` types
test mocks returning vendor objects Apollo discards

migration 085 backfills `vendor_name_snapshot` from the local vendor name, scoped to POs that reached GP. downgrade recreates the table and the FK.

three behaviours change:
- `markPoAsOrdered` deleted. it flipped a DRAFT to GP_REGISTERED without ever contacting GP. `createPo` and `registerPoInGp` are the only ways in. the relay-less seeding shortcut no longer works.
- `priorOrderAsValues` re-keys vendor id -> project id. `order_as` becomes the GP line's item number, so keying on product code alone would offer one supplier's catalogue number while raising a PO for another. remembered per job.
- `inventoryByVendor` groups on `gp_vendor_id` instead of joining `PO.vendor_id` -> `Vendor.name`. a pre-column PO with a snapshot and no id falls back to grouping on the name. fixes By Vendor collapsing everything into "No Vendor".

these stay:
`gpVendors`
`suggestVendorForManufacturer`
`manufacturer_vendor_map`
`PurchaseOrder.gp_vendor_id`
`PurchaseOrder.vendor_name_snapshot`
`PurchaseOrder.vendor_quote_number`
`PODocumentData.vendor_address`
the relay
`HardwareItem.vendor_no`

a draft PO shows no vendor until it is registered into GP.

out of scope: the relay's `GET /vendors` docstring still references the long-removed `syncGpVendors` mutation.

`testing/CLAUDE.md` updated for the removed page, the deleted Mark-as-Ordered path, and the wizard's PO cards.


frontend/src/modules/import/PurchaseOrdersStep.tsx
```diff
@@ -123,9 +122,11 @@ function VendorGroupCard({
     [aggregated],
   );
 
+  // Keyed on product code alone (#509). It used to need a Nexus vendor picked first, so the
+  // suggestions were usually absent at the one moment they are useful.
   const { data: priorData } = useQuery<PriorOrderAsQueryData>(GET_PRIOR_ORDER_AS_VALUES, {
-    variables: { vendorId: info.vendorId ?? '', productCodes },
-    skip: !info.vendorId || productCodes.length === 0,
+    variables: { productCodes },
+    skip: productCodes.length === 0,
```
Xilous, line 134: the skip guard used to be `!info.vendorId`, and nothing pre-fills a vendor, so in practice this query never fired from the wizard. now every `VendorGroupCard` fires it on mount, selected or not. testing/CLAUDE.md pins 41 manufacturer groups for contracterp-74.xml -> 41 concurrent `priorOrderAsValues` round trips the moment step 6 renders.

each one is now unindexed on the backend too: `get_prior_order_as_values` lost `PurchaseOrder.vendor_id == vendor_id` (which rode `ix_purchase_orders_vendor_id`, and migration 085 drops that index), leaving `po_line_items.product_code IN (...)` as the only filter. 008_indexes.py indexes `po_line_items.po_id` only - there is no index on `product_code`, so this is a seq scan of every PO line ever written, x41, against the Railway pool.

at minimum gate on `isSelected` so an untouched card costs nothing; better, hoist one query for the union of product codes across selected groups.

frontend/src/modules/po/poVendorName.ts
```diff
@@ -1,6 +1,7 @@
-// The GP vendor snapshot (issue #200) is the source of truth for display; the local vendor FK is only
-// a fallback for rows created before the snapshot existed. Kept in its own file so PO-adjacent modules
-// can reuse it without importing from index.tsx (which would create a component/non-component export mix).
-export function poVendorName(po: { vendorNameSnapshot: string | null; vendor: { name: string } | null }): string {
-  return po.vendorNameSnapshot ?? po.vendor?.name ?? '';
+// The GP vendor snapshot is the only vendor a PO has (#200, #509) - GP owns vendors and Nexus keeps
+// no records of its own. Empty for a DRAFT that has not been registered into GP yet, which is honest:
+// no vendor has been chosen. Kept in its own file so PO-adjacent modules can reuse it without
+// importing from index.tsx (which would create a component/non-component export mix).
+export function poVendorName(po: { vendorNameSnapshot: string | null }): string {
```
Xilous, line 5: the signature narrowed here but the type it reads from did not: `frontend/src/modules/po/index.tsx:146` still declares `vendor: VendorRef | null` as a required field of `PurchaseOrder`, and `VendorRef` is still exported at line 108.

no query selects `vendor` anymore, so at runtime the property is `undefined` on every `PurchaseOrder` the app holds while TypeScript reports it as `VendorRef | null`. `po.vendor?.name` still compiles and silently yields undefined.

it also means the fixtures this PR edited no longer satisfy their own annotation - `const stockDraft: PurchaseOrder = {...}` in GpPurchaseOrderDialog.test.tsx and `draftPo` in PODetailModal.test.tsx are missing a required property. `tsc -b` does not catch it only because tsconfig.app.json excludes `src/**/__tests__/**`.

delete `VendorRef` and the `vendor` field from the interface.

frontend/src/graphql/warehouse.ts
```diff
@@ -216,11 +212,6 @@ export const GET_PO_RECEIVING_DETAILS = gql`
       gpCompany
       gpVendorId
       vendorNameSnapshot
```
Xilous, line 214: the sub-selection is gone from the query but the type that consumes it is not. `frontend/src/modules/warehouse/receiveLines.ts:37` still has `vendor: { id: string; name: string; contactName: string | null } | null` on `PODetails`, and `frontend/src/modules/warehouse/ReceivingPage.tsx:54` still has `vendor: { id: string; name: string } | null` on `OpenPO` (GET_OPEN_POS, line 117 here).

both are now permanently undefined at runtime while typed as present-or-null. the warehouse test mocks still hand back a `vendor` object (ReceiveModal.test.tsx:71,150 and ReceiveDraftReviewModal.test.tsx:126) which Apollo now discards, so those mocks quietly over-specify and would not catch the field coming back.

frontend/src/modules/po/GpPurchaseOrderDialog.tsx
```diff
@@ -157,8 +156,8 @@ export default function GpPurchaseOrderDialog({
   const [taxDetailId, setTaxDetailId] = useState('');
   const [miscellaneous, setMiscellaneous] = useState('');
   const [tradeDiscount, setTradeDiscount] = useState('');
-  // Issue #256: create mode drafts carry an optional Nexus vendor link + the PM's preferred date.
-  const [vendorId, setVendorId] = useState<string | null>(null);
+  // Issue #256: create mode drafts carry the PM's preferred date. No vendor - GP owns those, and the
+  // GP vendor is picked at register time (#509).
```
Xilous, line 160: a fourth behaviour change the PR body does not list: the register dialog loses its GP-vendor pre-fill for every new draft.

line 424 still calls `bestGuessGpVendor(vendors, poVendorName(registerPo))`. `registerPo` is always a DRAFT, a DRAFT never has `vendorNameSnapshot` (only `register_po_in_gp` writes it, and that same call advances the status), so `poVendorName` returns `""` and `bestGuessGpVendor` returns `{vendorId: null, vendorName: null, confident: false}` on every open. the old fill came from the local vendor picked in the wizard, which this PR removes.

everything downstream of it is now unreachable:
- `importedVendorName` (738) is always null -> the `Imported as: X - confirm the GP vendor` helper (742) never renders
- `setVendorConfirmed(guess.confident)` always sets false, but `!gpVendorId` short-circuits first at 534, so the `Confirm the suggested GP vendor before registering` branch at 535 can never fire
- the `gpVendorId && !vendorConfirmed` checkbox at 873 can never show

#232 manufacturer suggestions now carry the pre-fill alone. either delete `bestGuessGpVendor` / `vendorConfirmed` / `importedVendorName` with the rest, or say in the PR body that the imported-name pre-fill is gone.

backend/alembic/versions/085_drop_local_vendors.py
```diff
@@ -0,0 +1,69 @@
...
+        SET vendor_name_snapshot = v.name
+        FROM vendors v
+        WHERE v.id = purchase_orders.vendor_id
+          AND purchase_orders.vendor_name_snapshot IS NULL
+        """
+    )
+
+    op.drop_index("ix_purchase_orders_vendor_id", table_name="purchase_orders")
+    op.drop_constraint("fk_purchase_orders_vendor_id", "purchase_orders", type_="foreignkey")
+    op.drop_column("purchase_orders", "vendor_id")
```
Xilous, line 48: the backfill is not scoped to POs that reached GP, so it writes a Nexus-local vendor name into a column this same PR redefines as "the GP vendor pick".

any DRAFT that was given a vendor in the import wizard has `vendor_id` set and `vendor_name_snapshot` NULL, so it gets backfilled. after that:
- purchase_order.py:60-63 "A DRAFT that has not been registered yet has neither, and shows no vendor at all" is false for those rows
- converters.py:406 "The GP vendor pick, the only vendor a PO has" is false for those rows
- the PO list, PO detail and `bestGuessGpVendor` all present the local name as the PM00200 vendor

`AND (status <> DRAFT OR gp_vendor_id IS NOT NULL)` keeps the display-name rescue for placed POs without minting fake GP snapshots on drafts.

backend/app/repositories/warehouse/inventory.py
```diff
@@ -184,18 +183,22 @@ def get_unlocated_inventory(session: Session, project_id: uuid.UUID | None = Non
...
-    """Group inventory by vendor name (via PO.vendor_id → Vendor), then product_code."""
+    """Group inventory by the GP vendor the hardware was bought from, then product_code.
+
+    The name is the PO's `vendor_name_snapshot` - the PM00200 vendor frozen on at GP-push time, which
+    since #509 is the only vendor a PO has. Stock-allocated rows have no PO behind them and a
+    never-registered draft has no snapshot yet; both land under "No Vendor".
+    """
     stmt = (
-        select(InventoryLocationModel, POLineItemModel.unit_cost, VendorModel.name)
+        select(InventoryLocationModel, POLineItemModel.unit_cost, POModel.vendor_name_snapshot)
```
Xilous: the group key went from an identifier to free display text. `Vendor.id` was canonical - one row, one group, no matter how the name was typed. `vendor_name_snapshot` is frozen text captured per push, so the same PM00200 vendor split across two POs whose snapshots differ by case, trailing whitespace, or a name edit in GP between the two pushes now renders as two separate vendors in the By Vendor view, each with a slice of the stock.

`PurchaseOrder.gp_vendor_id` is on the row and is the stable key. group on it and take a name per group (`max(vendor_name_snapshot)` or the most recent PO's), keeping snapshot-only fallback for rows with no gp_vendor_id.

backend/app/repositories/warehouse/inventory.py
```diff
@@ -184,18 +183,22 @@ def get_unlocated_inventory(session: Session, project_id: uuid.UUID | None = Non
...
         .outerjoin(POLineItemModel, InventoryLocationModel.po_line_item_id == POLineItemModel.id)
         .outerjoin(POModel, POLineItemModel.po_id == POModel.id)
-        .outerjoin(VendorModel, POModel.vendor_id == VendorModel.id)
         # Hide rows fully emptied by destock/allocation (kept in DB for FK integrity).
         .where(InventoryLocationModel.quantity > 0)
     )
     if project_id is not None:
         stmt = stmt.where(InventoryLocationModel.project_id == project_id)
-    stmt = stmt.order_by(VendorModel.name, InventoryLocationModel.product_code)
+    stmt = stmt.order_by(POModel.vendor_name_snapshot, InventoryLocationModel.product_code)
```
Xilous: this ORDER BY buys nothing now and did not before. everything below is bucketed into `vendor_map` and re-emitted through `sorted(vendor_map.keys())` / `sorted(pc_map.keys())`, so the SQL sort over the full outer-joined result set is pure work thrown away. dropping it also removes the NULLs-ordering question the snapshot column introduces.

backend/app/repositories/po_repository.py
```diff
@@ -790,13 +753,16 @@ def update_line_item_unit_cost(
...
     """
-    For a given vendor + set of product codes, return distinct non-empty
-    `order_as` values from po_line_items, grouped by product_code,
-    ranked by most recent line item updated_at.
+    For a set of product codes, return distinct non-empty `order_as` values from po_line_items,
+    grouped by product_code, ranked by most recent line item updated_at.
+
+    Scoped by product code alone (#509). This used to be scoped by the local vendor link as well,
+    which meant the suggestions only appeared once someone had picked one; that record is gone, and
+    the GP vendor a PO ends up on is not known until register time - long after the import wizard
```
Xilous: widening to the product code alone contradicts what `order_as` is. it is the alias printed on that vendor's PO - their catalogue SKU for the part - which is why the field is per-line and not on `hardware_items`. two vendors selling the same product code have different SKUs, so the wizard will now offer vendor B's number to a buyer ordering from vendor A, and confidently, since it ranks by recency.

the argument that no scope is available does not hold: the wizard groups by TITAN manufacturer, `POLineItem.hardware_items` already derives that manufacturer (that is what #232 reads), and a registered PO carries `gp_vendor_id`. scoping on manufacturer keeps the suggestion grounded and still fires before anyone picks a GP vendor, which is the problem this was solving.

frontend/src/graphql/po.ts
```diff
@@ -213,21 +206,14 @@ export const GET_PROJECT_SHIP_TO = gql`
 `;
 
 export const UPDATE_PO = gql`
-  mutation UpdatePO($id: ID!, $vendorId: ID, $expectedDeliveryDate: Date, $preferredDeliveryDate: Date, $poNumber: String, $vendorQuoteNumber: String, $notes: String, $shippingCost: Float, $tariffAmount: Float) {
-    updatePo(id: $id, vendorId: $vendorId, expectedDeliveryDate: $expectedDeliveryDate, preferredDeliveryDate: $preferredDeliveryDate, poNumber: $poNumber, vendorQuoteNumber: $vendorQuoteNumber, notes: $notes, shippingCost: $shippingCost, tariffAmount: $tariffAmount) {
+  mutation UpdatePO($id: ID!, $expectedDeliveryDate: Date, $preferredDeliveryDate: Date, $poNumber: String, $vendorQuoteNumber: String, $notes: String, $shippingCost: Float, $tariffAmount: Float) {
```
Xilous, line 209: seven pieces of schema disappear in one deploy with no deprecation window: `PurchaseOrder.vendor`, `adminStats.vendorCount`, `priorOrderAsValues(vendorId:)`, `updatePo(vendorId:)`, `CreateDraftPOInput.vendorId`, `RegisterPOInput.vendorId`, `PODraftInput.vendorId`.

a tab still running the previous bundle keeps sending the old documents. `GetPurchaseOrders` asking for `vendor { id name ... }` fails GraphQL validation outright, so the PO list goes to an error state rather than losing a column; same for `GetAdminStats` on the admin landing and `GetPriorOrderAsValues` in the wizard. `LazyBoundary` only recovers ChunkLoadError, so nothing reloads the tab.

CLAUDE.md records the house pattern for exactly this - the client-supplied actor args from #427 stayed in the schema deprecated and ignored, with removal deferred to #438. the arguments here can do the same: keep `vendorId` accepted and ignored, keep `vendor`/`vendorCount` resolving null/0, drop them next release.

backend/app/repositories/po_repository.py
```diff
@@ -679,9 +645,12 @@ def mark_po_as_ordered(session: Session, po_id: uuid.UUID) -> PurchaseOrder:
     - Validate exists + not soft-deleted (NotFoundError)
     - Validate status == Draft (InvalidStateTransitionError)
     - Validate po_number is not None (ValidationError)
-    - Validate vendor_id is not None (ValidationError)
     - Set status=GP_Registered, ordered_at=datetime.utcnow()
     - Return updated PO
+
+    A GP PO number is the whole requirement. There used to be a second one - a local vendor link -
+    but GP owns vendors and Nexus no longer records them (#509); the GP vendor arrives with the
```
Xilous: dropping this guard leaves no way to name the vendor on a PO placed outside the relay. `mark_po_as_ordered` now advances DRAFT -> GP_REGISTERED with `gp_vendor_id` NULL and `vendor_name_snapshot` NULL, and `updatePo` lost its `vendorId` argument in the same PR, so that PO shows a blank vendor in the PO list, PO detail, Back-Ordered Items, POs Awaiting Receipt and Receiving History permanently.

the mutation has no frontend caller today, but it is `SIGNED_IN` in ROOT_FIELD_POLICY and testing/CLAUDE.md advertises it as the way to place a PO with GP refusing. worth either requiring `gp_vendor_id` here, or noting in the docstring that a PO placed this way carries no vendor anywhere.

backend/tests/test_resolver_auth_gates.py
```diff
@@ -360,7 +360,7 @@ def test_the_jwt_is_verified_once_for_a_query_hitting_several_root_fields(monkey
     _explodes(monkeypatch, shop_assembly_module, "SessionLocal")
 
     _execute(
-        "{ warehouses { id } pullRequests { id } vendors { id } assembleList { id } }",
+        "{ warehouses { id } pullRequests { id } backOrderedItems { poNumber } assembleList { id } }",
```
Xilous: `vendors` was the third module in this query - `app.schemas.vendor`, deliberately un-exploded so a real resolver ran alongside the patched ones. `backOrderedItems` lives in `app.schemas.warehouse`, which `warehouses` and `pullRequests` already cover, so the test now spans two modules instead of three and every field is patched.

`relayAdoptWindow` or `stockItems` would keep the cross-module reach the test is named for.

Xilous:
review disposition, all 11 findings. fixed in c543aa8 unless noted.

order_as -> GP line item number in `gp_po.build_create_po_payload`, so it is supplier-facing: widening to the product code alone would offer one vendor's catalogue number while raising a PO for another. it is not manufacturer-scoped either - order-as is remembered per JOB. `priorOrderAsValues` now takes a project_id, null meaning a stock PO and scoping to the other project-less POs. pinned by
test_does_not_leak_across_projects
test_a_stock_po_sees_only_other_stock_pos

perf finding settled by the same scope. `po_line_items` carries only `ix_po_line_items_po_id`, so the unscoped version was one full scan per manufacturer card, 41 concurrent on the contracterp-74.xml fixture.

`mark_po_as_ordered` now requires gp_vendor_id: GP_REGISTERED means the PO exists in GP, so it has to name the vendor it was placed with, which `create_po` and `register_po_in_gp` already enforce. this path gated on the local vendor link instead, which was never a GP vendor. consequence - the relay-less seeding shortcut stops working, because nothing outside GP can supply a GP vendor. testing/CLAUDE.md now says so.

`inventoryByVendor` groups on gp_vendor_id, snapshot as label only. a pre-column PO carrying a snapshot and no id falls back to grouping on the name rather than collapsing into No Vendor. key tagged ("id", ...) / ("name", ...) so an id can never collide with a name. dead ORDER BY dropped.

migration 085's backfill is scoped to POs that reached GP. writing a Nexus-local name into the GP snapshot column on a never-registered draft would contradict the invariant the rest of the change establishes.

these dead types are gone:
VendorRef (modules/po/index.tsx)
PurchaseOrder.vendor (modules/po/index.tsx)
PODetails.vendor
OpenPO.vendor
four test mocks returning vendor objects Apollo discards

`test_the_jwt_is_verified_once_for_a_query_hitting_several_root_fields` uses stockItems now, restoring the third un-patched schema module the test is named for.

no deprecation window for vendorId, deliberately. production carries no live data yet, so a stale tab is a refresh rather than an incident, and leaving vendorId accepted-and-ignored would keep the vendor concept half-alive in the schema - the thing this PR exists to end. reconsider if production gains live data before merge.

two beyond the review, found while testing. `GetOpenPOs` and the receive-modal query still selected `vendor { }`, which broke POs Awaiting Receipt on the live Receiving page. fixed in 0168241. GraphQL documents are template strings, checked against the schema by neither tsc nor vitest. all 174 documents in src/ then validated against the live schema to confirm nothing else was left.

backend 1063 passed
frontend 527 passed
ruff clean
eslint 0 errors

Issue #512, opened 2026-08-05
sharepoint migration is unverified at production scale - 553 entries tested, 2,433 rows to migrate


Xilous:
PR #510 ships the SharePoint inventory migration. largest run verified end to end is 553 entries / 26,672 units on pr-510. production scale is 2,433 on-hand rows, never run at that size.

migrate_inventory is one mutation in one transaction. each entry hits receive_into_stock (find-or-create + quantity update + audit row), PROJECT entries add allocate_stock_to_project with two more audit rows - about 5 round trips per entry, so ~2,400 entries is roughly 10,000 statements over Railway's network hop. that is the shape CLAUDE.md's GraphQL/SQLAlchemy rules warn about.

migration is non-idempotent. proxy timeout -> client sees failure -> backend may have committed. wizard has no resume state and no idempotency marker, re-run doubles every row that landed.

- run the full migration against a PR environment with nothing excluded, and time it
- if it does not finish comfortably inside the proxy timeout, chunk the write into batches or move it to a background job with a progress read
- has_any_inventory answers "does this database hold any inventory", not "has this migration already run", so the re-run warning fires on any environment that has ever received a PO. a persisted marker makes that warning mean something and makes a resumable run safe.

run also creates CustomInventoryItem catalog rows, full-scale timing covers both writes.

Xilous:
tracking moved to #510's merge plan - full-scale timing re-runs on a fresh throwaway env before production.

PR #514, opened 2026-08-05
chore(test): throwaway PR to run the SharePoint migration at production scale (#512)

Xilous:
throwaway PR, spawns a preview environment with a fresh empty database for #512. not for merge, nothing to review.

same tree as PR #510. branch test/512-sharepoint-migration-scale sits at 8001b6b, tip of feature/sharepoint-inventory-migration, no commits of its own.

pr-510's database already holds the 553-entry partial migration run. migrateSharepointInventory is non-idempotent, no resume state, so re-running there doubles every row that landed.

#512 runs the full 2,433 rows against an empty database.

close once the timing is recorded.


Xilous:
a fresh PR environment self-configures with an empty database now, so this environment is reproducible on demand. the timing run happens on the reworked branch (merge plan on #510), on a fresh throwaway PR, with AZURE_* re-set by hand.

PR #528, opened 2026-08-06
test(shipping): pin the staging drag decision, and a fixture that can fill a skid (#470)

Xilous:
replaces #478, which was cut from the pre-rewrite master and could not be rebased onto the current one.

the same two commits, replayed onto current master with author dates preserved:
- staging drag decision pinned, plus a fixture that can fill a skid
- the staging drag paths and the 30-leaf cap run recorded

the other 62 commits on the old branch were master commits it had absorbed along the way, and are already in master. same 10 files, same 865 insertions.

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -1,10 +1,12 @@
 # Manual reference for backend env vars. Copy to backend/.env and fill in real values.
 DATABASE_URL=postgresql://postgres:postgres@localhost:5432/uc_nexus
 
-# Enables the GET /testing/clerk-sign-in token endpoint used by simulated user testing.
+# Enables the /testing/* endpoints used by simulated user testing: the clerk-sign-in token, and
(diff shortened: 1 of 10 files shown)


Xilous:
the #554 staging rewrite dropped the leaf model this pins (StagedLeaf / leafCount / canTakeAnotherLeaf / MAX_LEAVES_PER_SKID / sameLooseStock). the best textual merge of the two leaves 27 tsc errors - planContainerDrag reads a ContainerItem.itemType that no longer exists. #470 stays real, the backend still caps skids at MAX_LEAVES_PER_SKID, so the leaf paths need fresh tests against the post-#554 workspace. the require_testing_request refactor + ship-ready fixture are worth salvaging on a new branch, ported to #558's testing_sign_in_secret_hash() so previews keep the inherited sign-in secret.

